PR #1, opened 2017-03-30
recursive readme was too recursive

davharris:
fixed typo; dbplyr is not its own backend.


naupaka:
I was wondering about that too. Seems like Travis prefers recursion, though.

davharris:
I think this is the reason behind the failure?

>  Warning: Installed R (R Under development (unstable) (2017-03-30 r72444)) different from R used to build dplyr (R Under development (unstable) (2017-03-30 r72439)).

Either that or maybe Travis really does prefer the recursive name that Travis prefers.

hadley:
We did some crazy history rewriting so was easier to fix by hand. Thanks!

davharris:
Yeah... merging after the histories diverged by 250+ commits would _probably_ be a bad idea.

PR #6, opened 2017-04-19
rlang update

lionel-:


lionel-:
@hadley R CMD check passes locally. Once this is merged the dplyr PR should pass the Travis build. I think it fails because it tries to install dbplyr, which depends on dplyr, which currently depends on the old rlang.

hadley:
Thanks!

PR #8, opened 2017-05-02
Adding translations for Hive and Impala

edgararuiz:


R/db-hive.R
```diff
@@ -0,0 +1,43 @@
+#' @export
+db_desc.Hive <- function(x) {
```
hadley: Can you just make this a generic for the ODBC connection class? There's no need copy and paste the same code for each driver.

R/db-hive.R
```diff
@@ -0,0 +1,43 @@
...
+
+}
+
+#' @export
+sql_translate_env.Hive <- function(con) {
+
+  sql_variant(
+    scalar = dbplyr::sql_translator(
+      .parent = dbplyr::base_scalar,
+      as.numeric = function(x) build_sql("CAST(", x, " AS DOUBLE)"),
```
hadley: I think it would be useful to build a helper here:

```R
sql_cast <- function(type) {
  function(x) {
   build_sql("CAST(", x, " AS ", type, ")")
  }
}
```

That should live in the same file as `sql_infix()`.

R/db-hive.R
```diff
@@ -0,0 +1,43 @@
...
+      as.double  = function(x) build_sql("CAST(", x, " AS DOUBLE)"),
+      as.integer  = function(x) build_sql("CAST(", x, " AS INT)"),
```
hadley: Can you please make sure all the equals line up? That makes it easier to scan for differences.

R/db-impala.R
```diff
@@ -0,0 +1,44 @@
...
+    "/", info$dbname , "]")
+
+}
+
+#' @export
+sql_translate_env.Impala <- function(con) {
+
+  sql_variant(
+    scalar = dbplyr::sql_translator(
+      .parent = dbplyr::base_scalar,
```
hadley: Since these are so similar, I think it's probably a good idea to define a `base_odbc_scalar`, `base_odbc_agg` etc so we can reduce the amount of duplicated code across these functions.

R/simulate.r
```diff
@@ -10,6 +10,28 @@ sql_escape_ident.DBITestConnection <- function(con, x) {
   sql_quote(x, "`")
 }
 
+sql_select.DBITestConnection <- function(con, select, from, where = NULL,
```
hadley: Not sure why you need this?

R/simulate.r
```diff
@@ -58,5 +80,24 @@ simulate_mssql <- function() {
 
 }
 
+#' @export
+#' @rdname tbl_lazy
+simulate_hive <- function() {
+  structure(
+    list(),
+    class = c("Hive", "DBITestConnection")
```
hadley: All these should have class "DBIConnection" not "DBITestConnection"

tests/testthat/test-hive.R
```diff
@@ -0,0 +1,172 @@
+context("Hive translation")
+
+# Initialize simulated DB connection -------------------------------------
+
+hive_connection <- dbplyr:::simulate_hive()
```
hadley: Can you follow the updated style in `test-translate-mssql()` (including the context and the file name)

tests/testthat/test-hive.R
```diff
@@ -0,0 +1,172 @@
...
+
+# Hive base_scalar conversions -----------------------------------------
+
+test_that("as.numeric field coercion translates to Hive's DOUBLE ", {
```
hadley: Given that the underlying code is 90% the same as the other odbc backends, I think you only need a couple of tests for the things that are different.

R/db-hive.R
hadley: (i.e. for the OdbcConnection class)

R/simulate.r
edgararuiz: If I use `DBIConnection` I get:  `unable to find an inherited method for function ‘dbQuoteIdentifier’ for signature ‘"Impala", "ident"...`, but if I use `DBITestConnection` and add `sql_select` to it, all tests work.  I figured I add the method where its least impactful

R/simulate.r
edgararuiz: The `tbl_lazy` tests fail without this command

tests/testthat/test-hive.R
edgararuiz: Yes! I hate that I saw that after submitting the PR, I would have the new tests conform before sending the PR your way

R/db-hive.R
edgararuiz: Sure thing, do you have a preference on where to place it?

R/simulate.r
hadley: Oh in that case the class should contain both "DBITestConnection" and "DBIConnection" and then you can delete the `sql_select()` method

R/db-hive.R
edgararuiz: I'm very glad that you added `ident_q`. The original code kept returning `CAST(field1 as 'INT')`, which I was able to fix by wrapping `type` with the new function

hadley: Using `SQL()` would be slightly more appropriate here because you're constructing literal SQL, not an unquoted identifier.

edgararuiz:
In addition to the requested changes, other major updates were: adding support for `paste` for Hive and Impala by using the `CONCAT_WS` command, and removing unnecessary quotes from the resulting SQL code: `SELECT am, COUNT(*) AS count, AVG(mpg) AS mean_mpg FROM impala_mtcars GROUP BY am`.  Removing quotes reduces the size of the SQL query and, to me, it makes it a bit easier to read (http://colorado.rstudio.com:3939/content/490/dbplyr-impala.nb.html) .   I also took the liberty of updating the error messages that the devel dplyr version returns, which was the last hurdle to pass the Travis CI checks.

R/db-impala.R
```diff
@@ -0,0 +1,13 @@
+#' @export
+sql_translate_env.Impala <- function(con) {
+  sql_variant(
+    scalar = sql_translator(.parent = base_odbc_scalar,
+      as.Date  = function(x) build_sql("CAST(", x, " AS VARCHAR(10))")
```
hadley: Can use `sql_cast()` here too

R/db-odbc.R
```diff
@@ -0,0 +1,24 @@
...
+  paste0(
+    info$dbms.name, " ", info$db.version,
+    "[", info$username, "@", host, port,
+    "/", info$dbname , "]")
+
+}
+
+#' @export
+sql_escape_ident.OdbcConnection <- function(con, x) {
+  x
```
hadley: It's really important to properly escape the identifiers. It does make the SQL longer than what you'd usually write by hand, but it protects the user against knowing all the special SQL keywords and ensure it works with unnamed `summarise()` which creates variable names like `mean(x)`

R/translate-sql-base.r
```diff
@@ -273,3 +273,49 @@ base_no_win <- sql_translator(
...
+#' @format NULL
+base_odbc_scalar <- sql_translator(
+  .parent = base_scalar,
+  as.numeric    = sql_cast("DOUBLE"),
+  as.double     = sql_cast("DOUBLE"),
+  as.integer    = sql_cast("INT"),
+  as.logical    = sql_cast("BOOLEAN"),
+  as.character  = sql_cast("STRING"),
+  as.Date       = sql_cast("DATE"),
+  round         = function(x, digits = 0L)
```
hadley: For multi-line functions, can you please always use `{``

R/translate-sql-base.r
```diff
@@ -273,3 +273,49 @@ base_no_win <- sql_translator(
   lag          = win_absent("lag"),
   order_by     = win_absent("order_by")
 )
+
+#' @export
+#' @rdname sql_variant
+#' @format NULL
+base_odbc_scalar <- sql_translator(
```
hadley: I think this should live in `db-odbc.R`, and then we should probably call the other files `db-odbc-msql.R`, `db-odbc-hive.R` etc

R/translate-sql-base.r
```diff
@@ -273,3 +273,49 @@ base_no_win <- sql_translator(
...
+                    )
+  )
+
+#' @export
+#' @rdname sql_variant
+#' @format NULL
+base_odbc_agg <- sql_translator(.parent = base_agg,
+  n             = function() sql("COUNT(*)"),
+  count         = function() sql("COUNT(*)")
+  )
```
hadley: `)` needs to be dedented

R/translate-sql-base.r
```diff
@@ -273,3 +273,49 @@ base_no_win <- sql_translator(
...
+
+
+
+#' @export
+#' @rdname sql_variant
+#' @format NULL
+base_odbc_win <- sql_translator(.parent = base_win,
+  n             = function() sql("COUNT(*)"),
+  count         = function() sql("COUNT(*)")
+  )
```
hadley: And here. Are `n` and `count` indented by two spaces?

R/translate-sql-helpers.r
```diff
@@ -153,3 +153,11 @@ sql_not_supported <- function(f) {
     stop(f, " is not available in this SQL variant", call. = FALSE)
   }
 }
+
+#' @rdname sql_variant
+#' @export
+sql_cast <- function(type) {
+  function(x) {
+    build_sql("CAST(", x, " AS ", ident_q(type), ")")
```
hadley: Slightly clearer to use `sql()` here

tests/testthat/test-translate-hive.R
```diff
@@ -0,0 +1,150 @@
+context("translate-Hive")
+
+# Hive base_scalar conversions -----------------------------------------
+
+test_that("as.numeric() translates to DOUBLE ", {
```
hadley: Since you now have one function that generates all the casts, you can just test that function and delete all these tests.

R/simulate.r
```diff
@@ -53,10 +53,27 @@ simulate_postgres <- function() {
 simulate_mssql <- function() {
```
hadley: Maybe we should now have:

```R
simulate_odbc <- function(type = NULL) {
  structure(
    list(),
    class = c(NULL, "OdbcConnection", "DBIConnection")
  )
}
simulate_hive <- function() simulate_odbc("hive")
simulate_impala <- function() simulate_odbc("impala")
simulate_mssql <- function() simulate_odbc("Microsoft SQL Server")
```

R/translate-sql-base.r
edgararuiz: It gives me an `unexpected symbol` error when I add the brackets, it also looks like the original `base_scalar` does not have them.  Maybe I'm missing something here?

hadley:
Thanks!

PR #10, opened 2017-05-03
Update by column not found error messages

ianmcook:
This resolves two test errors:
`Failure: join functions error on column not found for SQL sources #1928 (@test-joins.R#62)`
`Failure: join functions error on column not found for SQL sources #1928 (@test-joins.R#66)`


ianmcook:
Looks like Edgar also modified these same tests in #8.

PR #11, opened 2017-05-07
two stray dplyr->dbplyr changes

mdsumner:
the CRAN badge and the install from Github links still pointed to dplyr


hadley:
Thanks!

PR #12, opened 2017-05-10
adding ident class to table name in copy_to 

hannesmuehleisen:
so dbi_quote correctly calls dbQuoteIdentifier and not dbQuoteString


hannesmuehleisen:
Don't think the check failure is my fault given the previous revision is also failing with the same error.

hadley:
I think the correct fix is `name <- as.sql(name)`. Can you please double check that works? (or even better, figure out a test case, if that's at all possible)

hadley:
(I also fixed the build failure - sorry about that)

hannesmuehleisen:
closing in favor of https://github.com/tidyverse/dbplyr/pull/13

PR #15, opened 2017-06-10
Adding two new databases and increased support for the existing ones

edgararuiz:
Highlights:  

- Added support for ODBC connection based Oracle and PostgreSQL databases
- Created a separate db_copy_to() for ODBC connection types
- Added individual db_analyze() for each of the 4 databases (except PostgreSQL)
- Increased the number of working translations based on results from the new automated testing package: https://github.com/rstudio/dbtest (latest results attached)
- Added support for windowing functions
- Added support for certain functions by implementing their formula using SQL (i.e. COSH)
- Included links to comments where necessary
- Added tests for the new translations

oracle

other-dbs


edgararuiz:
Hi @hadley,  travis-ci checks keep failing even though tests pass when I run them locally in my computer.  Could this be an upstream dependency issue?

krlmlr:
I think this is the relevant line in the check output: https://travis-ci.org/tidyverse/dbplyr/jobs/241610300#L927. Why are you using plyr?

edgararuiz:
Hi @krlmlr , right, I saw that line yesterday and checked to see if I had any references to plyr but I couldn't find any.  That's why I think it may be an update to an upstream dependency, like maybe dplyr or rlang?

R/db-odbc-oracle.R
```diff
@@ -0,0 +1,235 @@
...
+
+  if (!is.null(limit) && !identical(limit, Inf)) {
+    assertthat::assert_that(is.numeric(limit), length(limit) == 1L, limit > 0)
+    out$limit <- build_sql(
+      "FETCH FIRST ", sql(format(trunc(limit), scientific = FALSE)), " ROWS ONLY ",
+      con = con
+    )
+  }
+
+  escape(unname(plyr::compact(out)), collapse = "\n", parens = FALSE, con = con)
```
krlmlr: @edgararuiz: I think we need to substitute at least this call.

edgararuiz: Oh man! How did I miss that?!  I'll fix it right now.  Thank you!

DESCRIPTION
```diff
@@ -81,7 +83,7 @@ Collate:
     'utils-format.r'
     'utils.r'
     'zzz.R'
-RoxygenNote: 6.0.1
+RoxygenNote: 6.0.1.9000
```
hadley: Can you please use the released version of roxygen2?

NEWS.md
```diff
@@ -175,6 +175,9 @@
 
 *  SQL translation for Apache Impala (@edgararuiz)
 
+*  SQL translation for Oracle  (@edgararuiz)
```
hadley: Can you please move up under `1.0.0.9000`. You should also mention the analyse support there too.

R/db-odbc-hive.R
```diff
@@ -1,8 +1,27 @@
 #' @export
 sql_translate_env.Hive <- function(con) {
   sql_variant(
-    base_odbc_scalar,
+    scalar = sql_translator(.parent = base_odbc_scalar,
+      var       = sql_prefix("VARIANCE"),
+      # cot calculation is based on this article
```
hadley: I think you can remove this comment since COT is 1/TAN by definition.

R/db-odbc-hive.R
```diff
@@ -1,8 +1,27 @@
...
     base_odbc_agg,
     base_odbc_win
   )
 }
+
+#' @export
+db_analyze.Hive <- function(con, table, ...) {
+  # Using ANALYZE TABLE instead of ANALYZE as recommended in this article
+  # https://cwiki.apache.org/confluence/display/Hive/StatsDev
+  sql <- dbplyr::build_sql(
```
hadley: I think this could all fit on one line?

hadley: You don't need `dbplyr::` since we're inside the package.

R/db-odbc-mssql.R
```diff
@@ -68,7 +68,28 @@
       as.double     = sql_cast("NUMERIC"),
       as.character  = sql_cast("VARCHAR(MAX)"),
       log           = sql_prefix("LOG"),
+      nchar         = sql_prefix("LEN"),
+      atan2         = sql_prefix("ATN2"),
+      substr        = sql_prefix("SUBSTRING"),
```
hadley: Can you confirm that the argument order is the same?

R/db-odbc-mssql.R
```diff
@@ -68,7 +68,28 @@
...
+      ceil          = sql_prefix("CEILING"),
       ceiling       = sql_prefix("CEILING"),
+      # Casting as BIT converts the Integer result into an actual logical
```
hadley: Can you provide an example where the default `x IS NULL` doesn't work?

R/db-odbc-hive.R
hadley: The other analyse methods need similar treatment.

R/db-odbc-oracle.R
```diff
@@ -0,0 +1,235 @@
+#' @export
+sql_select.Oracle<- function(con, select, from, where = NULL,
```
hadley: It feels like we're getting a lot of duplicated code in this selection methods. Maybe it's time to break out smaller methods `sql_select_select`, `sql_select_where`, `sql_select_group_by` etc?

R/db-odbc-oracle.R
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  escape(unname(compact(out)), collapse = "\n", parens = FALSE, con = con)
+}
+
+
+#' @export
+sql_translate_env.Oracle <- function(con) {
+  sql_variant(
+    scalar = sql_translator(.parent = base_odbc_scalar,
```
hadley: Needs indenting fix

R/db-odbc-oracle.R
```diff
@@ -0,0 +1,235 @@
...
+    base_odbc_agg,
+    base_odbc_win
+  )
+}
+
+
+#' @export
+sql_subquery.Oracle <- function(con, from, name = dbplyr:::unique_name(), ...) {
+  # Query fails if a quoted table is used.  Using sql() to remove quotes from an
+  # already 'idented' argument.
```
hadley: Can you provide an example?

R/db-odbc-oracle.R
```diff
@@ -0,0 +1,235 @@
...
+db_has_table.Oracle <- function(con, table){
+  # This function allows copy_to() to work.  This variance may
+  # be removed when the odbc package is able to handle a straight
+  # dbListTables
+  stopifnot(length(table) == 1)
+  toupper(table) %in% dbListTables(con)
+}
+
+#' @export
+db_copy_to.Oracle <- function(con, table, values,
```
hadley: I think this belongs in the `dbWriteTable()` method. 

@jimhester can we move the equivalent of this code into odbc?

R/db-odbc-postgres.R
```diff
@@ -0,0 +1,61 @@
...
+        build_sql(
+          "ROUND(", x, ", ", as.integer(digits),")"
+        )},
+      log = function(x, base = exp(1)) {
+        if (isTRUE(all.equal(base, exp(1)))) {
+          build_sql("ln(", x, ")")
+        } else {
+          # Use log change-of-base because postgres doesn't support the
+          # two-argument "log(base, x)" for floating point x.
+          build_sql("log(", x, ") / log(", base, ")")
```
hadley: Why are we using `log()` here but `ln()` above?

R/db-odbc-postgres.R
```diff
@@ -0,0 +1,61 @@
...
+        }
+      },
+      paste = function(..., sep = " "){
+        build_sql(
+          "CONCAT_WS(",sep, ", ",escape(c(...), parens = "", collapse = ","),")"
```
hadley: Missing some spaces here

R/translate-sql-odbc.R
```diff
@@ -9,15 +9,30 @@ base_odbc_scalar <- sql_translator(
...
-                    build_sql(
-                      "CONCAT_WS(",sep, ", ",escape(c(...), parens = "", collapse = ","),")"
-                      )
+  # cosh, sinh, coth and tanh calculations are based on this article
+  # https://en.wikipedia.org/wiki/Hyperbolic_function
+  cosh = function(x){
+    build_sql("(EXP(", x, ") + EXP(-", x,")) / 2")
+  },
+  sinh = function(x){
+    build_sql("(EXP(", x, ") - EXP(-", x,")) / 2")
```
hadley: I think you need more parens: i.e. if x is "a + b", you don't want `EXP(-a + b)`

R/translate-sql-odbc.R
```diff
@@ -35,7 +50,79 @@ base_odbc_agg <- sql_translator(
 base_odbc_win <- sql_translator(
   .parent = base_win,
   n             = function() sql("COUNT(*)"),
-  count         = function() sql("COUNT(*)")
+  count         = function() sql("COUNT(*)"),
+
+  # Window functions in this variant use field that its being
+  # 'windowed' as the secondary ORDER BY instead of the field
+  # from win_current_order().  In the tests, having an ORDER BY
```
hadley: That doesn't sound right to me. Can you provide an example?

tests/testthat/test-translate-oracle.R
```diff
@@ -0,0 +1,17 @@
+context("translate-Oracle")
+
+# Oracle base_scalar conversions -----------------------------------------
```
hadley: I think you need to include some more tests around the subquery stuff here.

R/db-odbc-oracle.R
jimhester: We can, although I have been trying to keep vender specific workarounds to the minimum. But Oracle's driver doesn't seem to support parameterized queries (at least the way that nanodbc / odbc does them), hence the below approach.

hadley: I'd also rather minimise vendor specific workarounds, but if they live anywhere, it seems like it should be in odbc.

R/db-odbc-mssql.R
edgararuiz: Yes sir, here it is:

``` 
Select mpg is null as test from mtcars
```

```
  nanodbc/nanodbc.cpp:1344: 42000: Incorrect syntax near the keyword 'is'., Statement(s) could not be prepared.,  
Failed to execute SQL chunk
```
I also looked at `isnull`, but needs a replacement value, so it's not helpful

```
Select isnull(mpg, 1)  as test from mtcars
```

R/db-odbc-oracle.R
edgararuiz: Yes, 

```
select * from `mtcars7`
```
```
  nanodbc/nanodbc.cpp:1344: HY000: ORA-00911: invalid character
 
Failed to execute SQL chunk
```
```
select * from 'mtcars7'
```

```
  nanodbc/nanodbc.cpp:1344: HY000: ORA-00903: invalid table name
 
Failed to execute SQL chunk
```
```
select * from "mtcars7"
```
```
  nanodbc/nanodbc.cpp:1344: HY000: ORA-00942: table or view does not exist
 
Failed to execute SQL chunk
```

This one was successful: 

```
select * from mtcars7
```

Another successful one was this one:

```
select * from "MTCARS7"
```
Because the double quotes seems to work like `_select()`, so the table needs to match the exact capitalization of the table name.

R/db-odbc-postgres.R
edgararuiz: I used the translation from `db-postgres.r` as the base, so I didn't really question it.  I can double check it.

R/db-odbc-mssql.R
hadley: What version of SQL server are you using? [This page](https://docs.microsoft.com/en-us/sql/t-sql/queries/is-null-transact-sql) suggests that `IS NULL` should work.

Does it only work in WHERE statements?

R/db-odbc-oracle.R
hadley: Why is the table getting created as "MTCARS7"? That suggests we are missing some quotes somewhere.

R/translate-sql-odbc.R
edgararuiz: Yes sir, 

I ran this code on two machines, one with the code from this PR and another with CRAN versions:

```
m <- tbl(con, "mtcars") %>%
  mutate(next_mpg = lead(mpg))

show_query(m)

m
```

CRAN version

```
<SQL>
SELECT "row_names", "mpg", "cyl", "disp", "hp", 
"drat", "wt", "qsec", "vs", "am", "gear", "carb", LEAD("mpg") AS "next_mpg"
FROM "mtcars"
 Show Traceback
Error in new_result(connection@ptr, statement) : nanodbc/nanodbc.cpp:1344: HY000: The function 'LEAD' must have an OVER clause., Statement(s) could not be prepared.,
```
From this PR, the query is successful:

```
<SQL>
SELECT "row_names", "mpg", "cyl", "disp", "hp", 
"drat", "wt", "qsec", "vs", "am", "gear", "carb", LEAD("mpg", 1) OVER (ORDER BY "mpg") AS "next_mpg"
FROM "mtcars"
```

Here is one using `group_by` to activate `PARTITION`: 

```
m <- tbl(con, "mtcars") %>%
  group_by(am) %>%
  mutate(next_mpg = lead(mpg))

show_query(m)

m
```
From CRAN:

```
<SQL>
SELECT "row_names", "mpg", "cyl", "disp", "hp",
 "drat", "wt", "qsec", "vs", "am", "gear", "carb", LEAD("mpg") AS "next_mpg"
FROM "mtcars"
 Show Traceback
Error in new_result(connection@ptr, statement) : nanodbc/nanodbc.cpp:1344: HY000: The function 'LEAD' must have an OVER clause., Statement(s) could not be prepared.,
```
From this PR, the query is successful:

```
<SQL>
SELECT "row_names", "mpg", "cyl", "disp", "hp", 
"drat", "wt", "qsec", "vs", "am", "gear", "carb", LEAD("mpg", 1) OVER (PARTITION BY "am" ORDER BY "mpg") AS "next_mpg"
FROM "mtcars"
```

R/db-odbc-mssql.R
edgararuiz: Yes, it does work in the WHERE statement, but it doesn't seem to work if used in the SELECT statement

```
Select * from mtcars where mpg is null
```

```
0 rows | 1-10 of 12 columns
```

R/db-odbc-oracle.R
edgararuiz: Sorry, I should have also mention that in the Oracle server I'm working on, any successfuly created table is capitalized,  if I create `mtcars7`, it's recorded as `MTCARS7`.  It does the same capitalization for field names, that's why the `count()` test fails, because the field name ends up being `N` and not `n`.

So when I call if I call `"mtcars7"`, it looks for a table that has that exact capitalization, but if I use `mtcars7` then it won't enforce the 'literal' search.  

The challenge I faced was to try to find a balance to when to use `toupper()`.  At one point I used it everywhere, but it made a bunch of test fail because of the constant change of names. 

I'll be glad to do some more work on this if you think we're not there yet.

R/db-odbc-mssql.R
edgararuiz: Good point,`SUBSTRING` uses x, start length, while `substr()` uses x, start, stop, I'll fix it.  Thank you

R/db-odbc-oracle.R
edgararuiz: I like that idea.  It looks like most of the clauses follow the same pattern, should I target a `sql_select_clause()` function that takes the clause as an argument?  So it may work like `out$group_by <- sql_select_clause("GROUP BY", group_by)`?

R/translate-sql-odbc.R
hadley: I'd say the CRAN version is correct because you haven't specified an ordering using `arrange()`. I think the user has to do this themselves, rather than us guessing.

R/db-odbc-oracle.R
hadley: If the table and field names are getting capitalised automatically, it usually indicates that something is not getting escaped correctly, and we should fix that upstream.

R/db-odbc-oracle.R
hadley: Yes, that sounds great.

R/db-odbc-mssql.R
hadley: What about `CAST((x IS NULL) AS BIT)`?

R/translate-sql-odbc.R
edgararuiz: Oh ok, I was attempting to match what we can do with dplyr locally, like: 
```
 mtcars %>%
  mutate(next_mpg = lead(mpg)) 
```
Also, I did try using `arrange()` but it fails for me on both PR and CRAN versions.  Mainly because `arrange()` creates a subquery with a`ORDER BY`, instead of adding the `ORDER BY` to the partition clause  .Here are the errors I get:

Code:
```
m <- tbl(con, "mtcars") %>%
  arrange(mpg) %>%
  mutate(next_mpg = lead(mpg))

m
```

CRAN:

```
<SQL>
SELECT "row_names", "mpg", "cyl", "disp", "hp", "drat", 
"wt", "qsec", "vs", "am", "gear", "carb", LEAD("mpg") AS "next_mpg"
FROM (SELECT *
FROM "mtcars"
ORDER BY "mpg") "jllnbfdoyt"
```
```
 Show Traceback
Error in new_result(connection@ptr, statement) : nanodbc/nanodbc.cpp:1344: HY000:
 The function 'LEAD' must have an OVER clause., Incorrect syntax near the keyword 
'ORDER'., Statement(s) could not be prepared.,
```

PR:

```
Error in new_result(connection@ptr, statement) : nanodbc/nanodbc.cpp:1344: HY000: 
The ORDER BY clause is invalid in views, inline functions, derived tables, subqueries,
 and common table expressions, unless TOP, OFFSET or FOR XML is also specified., 
Statement(s) could not be prepared.,
```

R/db-odbc-mssql.R
edgararuiz: It doesn't look like it would.  I tried:

```
Select CAST((mpg IS NULL) AS BIT)from mtcars 
```

```
  nanodbc/nanodbc.cpp:1344: 42000: Incorrect syntax near the keyword 'AS'., Statement(s) could not be prepared.,  
Failed to execute SQL chunk
```

This  query worked:

```
Select CAST(mpg AS BIT)from mtcars 
```
So I tried a different evaluation call inside a parenthesis, but it also fails: 

```
Select CAST((mpg > 23) AS BIT)from mtcars 
```

```
  nanodbc/nanodbc.cpp:1344: 42000: Incorrect syntax near the keyword 'AS'., Statement(s) could not be prepared.,  
Failed to execute SQL chunk
```

R/translate-sql-odbc.R
hadley: That works for me with a PostgreSQL source:

```R
<SQL>
SELECT "mpg", "cyl", "disp", "hp", "drat", "wt", "qsec", "vs", "am", "gear", "carb", LEAD("mpg", 1, NULL) OVER (ORDER BY "mpg") AS "next_mpg"
FROM (SELECT *
FROM "mtcars"
ORDER BY "mpg") "avqlegvybf"
```

hadley: The output that you're getting suggests that windows functions are not getting registered correctly.

edgararuiz: Oh ok, I'll go back and check, thank you

R/db-odbc-oracle.R
hadley: Could you please create a minimal reprex using only DBI code and file it on odbc?

edgararuiz: Ok, I will

R/translate-sql-odbc.R
edgararuiz: I ran the query from your previous comment against two databases.  The query failed when I run it against MSSQL, but it succeeds when running it against Impala.

MS SQL error:

```
  nanodbc/nanodbc.cpp:1344: HY000: The ORDER BY clause is invalid in views, 
inline functions, derived tables, subqueries, and common table expressions, 
unless TOP, OFFSET or FOR XML is also specified., Statement(s) could not be 
prepared.,  
Failed to execute SQL chunk
```
It looks like moving it to the MS SQL translation is the way to go, is that ok?

R/db-odbc-oracle.R
edgararuiz: @jimhester let me know that you met to review the code and figured that what we need is to use the double quote as the identifier.  Jim shared also this link with me: https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements008.htm.  Adding  `sql_escape.ident.Oracle()` function seemed to have done the trick.  I removed all but the `sql_select`, `sql_translate_env` and `db_analyze` from the script.  This means that `copy_to` will not work until a fix is available from Simba.  Thanks!

hadley: Why isn't the double quote getting picked automatically by odbc?

R/translate-sql-odbc.R
hadley: Can you file a separate issue? I think it will require more thought to make sure that we don't include an `ORDER BY` inside a subquery.

ianmcook: I had to solve this problem in implyr, and I did it by implementing a recursive function that's called in `sql_build()`. See https://github.com/ianmcook/implyr/commit/08a704a130104582c4b531817fd4016980f65114.

edgararuiz: Thanks @ianmcook 
@hadley  - I've opened issue #2874 in dplyr

NEWS.md
```diff
@@ -1,5 +1,14 @@
 # dbplyr 1.0.0.9000
 
+## Backends
```
hadley: I add the second level headings just before release (because it's too hard to know in advance exactly what will need a heading)

NEWS.md
```diff
@@ -1,5 +1,14 @@
...
+
+*  SQL translation for Oracle  (@edgararuiz)
+
+## Minor bug fixes and improvements
+
+* Added custom db_analyze_table() for MS SQL, Oracle, Hive and Impala (@edgararuiz)
+
+# dbplyr 0.5.0.9000
```
hadley: This line snuck in somehow.

hadley:
Thanks @edgararuiz!

PR #17, opened 2017-06-16
Fix for DESCRIPTION, NAMESPACE and man/sql_variant

edgararuiz:


hadley:
Thanks for the quick fix!

PR #18, opened 2017-06-19
Remove duplicate definition

krlmlr:
of `db_disconnector()`.

Other (active) definition at https://github.com/tidyverse/dbplyr/blob/2a4c041821282c3ecbbfa7f72fa67446911c69ef/R/src_dbi.R#L119.

How can we explicitly disconnect a source to avoid the message?


hadley:
Maybe we should just default to `quiet = TRUE`?

krlmlr:
Changed default, because disconnection is practically always needed (and still will be for new-style DBI backends).

R/src_dbi.R
```diff
@@ -90,9 +90,12 @@
 #'   summarise(n = n()) %>%
 #'   show_query()
 #' }
-src_dbi <- function(con, auto_disconnect = FALSE) {
+src_dbi <- function(con, auto_disconnect = TRUE) {
   # stopifnot(is(con, "DBIConnection"))
-  disco <- if (isTRUE(auto_disconnect)) db_disconnector(con)
+  if (is_false(auto_disconnect))
```
hadley: Can you use `{}` here just to match tidyverse style?

hadley:
Any idea why it's failing on R-devel?

krlmlr:
No idea. I purged the cache and re-triggered the build, need to wait for the result.

krlmlr:
All OK now.

hadley:
Thanks! Can you please add a bullet to NEWS?

R/src_dbi.R
```diff
@@ -90,9 +90,13 @@
 #'   summarise(n = n()) %>%
 #'   show_query()
 #' }
-src_dbi <- function(con, auto_disconnect = FALSE) {
+src_dbi <- function(con, auto_disconnect = TRUE) {
```
hadley: This can't be `TRUE` because we create a bunch of transient `src_dbi`s when working directly with DBI connections

hadley: Or at least we need to change to this in dplyr:

```R
#' @export
tbl.DBIConnection <- function(src, from, ...) {
  check_dbplyr()
  tbl(dbplyr::src_dbi(src), from = from, ..., auto_disconnect = FALSE)
}
```

krlmlr: Sent PR to dplyr. A coarse CRAN search suggests that the default for `src_dbi()` should be `auto_disconnect = FALSE`, but only from the next dplyr release on.

hadley: I rolled it back in dbplyr, I'm pretty sure defaulting to `FALSE` is correct - you should set to `TRUE` only when creating the connection in the `src_dbi()` call, or when you know the scope of the source and the connection are identical.

PR #20, opened 2017-06-20
Adds sd() support for 5 datasources

edgararuiz:
Adds sd() scalar, aggregate and window support for MSSQL, Oracle, Hive, Impala and Postgres (for both odbc and RPostgreSQL).  It also includes some further code tidying.  Addresses `dplyr` issue https://github.com/tidyverse/dplyr/issues/2887

```
> tbl(con, "iris") %>%
+   group_by(Species) %>%
+   mutate(A_sd = sd(Sepal.Length))
```

```
# Source:   lazy query [?? x 6]
# Database: Microsoft SQL Server 12.00.4422[rstudioadmin@EC2AMAZ-O5QKKM8/airontime]
# Groups:   Species
   Sepal.Length Sepal.Width Petal.Length Petal.Width Species      A_sd
          <dbl>       <dbl>        <dbl>       <dbl>   <chr>     <dbl>
 1          5.1         3.5          1.4         0.2  setosa 0.3524897
 2          4.9         3.0          1.4         0.2  setosa 0.3524897
 3          4.7         3.2          1.3         0.2  setosa 0.3524897
```

### DIFF
--- a/R/db-odbc-mssql.R
+++ b/R/db-odbc-mssql.R
@@ -47,6 +47,7 @@
       atan2         = sql_prefix("ATN2"),
       ceil          = sql_prefix("CEILING"),
       ceiling       = sql_prefix("CEILING"),
+      sd            = sql_prefix("STDEV"),
       substr        = function(x, start, stop){
                         len <- stop - start + 1
                         build_sql(
@@ -80,7 +81,10 @@
       cov           = sql_not_supported("cov()")
 
     ),
-    base_odbc_win
+    sql_variant(
+      scalar = sql_translator(.parent = base_odbc_win,
+        sd          = win_recycled("STDEV")
+      ))
   )}
 
 #' @export
--- a/R/db-odbc-mssql.R
+++ b/R/db-odbc-mssql.R
@@ -47,7 +47,6 @@
       atan2         = sql_prefix("ATN2"),
       ceil          = sql_prefix("CEILING"),
       ceiling       = sql_prefix("CEILING"),
-      sd            = sql_prefix("STDEV"),
       substr        = function(x, start, stop){
                         len <- stop - start + 1
                         build_sql(
--- a/R/db-odbc-mssql.R
+++ b/R/db-odbc-mssql.R
@@ -38,7 +38,7 @@
 #' @export
 `sql_translate_env.Microsoft SQL Server` <- function(con) {
   sql_variant(
-    scalar = sql_translator(.parent = base_odbc_scalar,
+    sql_translator(.parent = base_odbc_scalar,
       as.numeric    = sql_cast("NUMERIC"),
       as.double     = sql_cast("NUMERIC"),
       as.character  = sql_cast("VARCHAR(MAX)"),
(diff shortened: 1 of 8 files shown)


R/db-postgres.r
```diff
@@ -11,27 +11,30 @@ db_desc.PostgreSQLConnection <- function(x) {
...
-        if (isTRUE(all.equal(base, exp(1)))) {
-          build_sql("ln(", x, ")")
-        } else {
-          # Use log change-of-base because postgres doesn't support the
-          # two-argument "log(base, x)" for floating point x.
-          build_sql("log(", x, ") / log(", base, ")")
-        }
-      }
+      sd    =  sql_prefix("stddev_samp"),
+      log   = function(x, base = exp(1)) {
```
hadley: Indenting got messed up here

R/db-postgres.r
```diff
@@ -11,27 +11,30 @@ db_desc.PostgreSQLConnection <- function(x) {
...
-          build_sql("ln(", x, ")")
-        } else {
-          # Use log change-of-base because postgres doesn't support the
-          # two-argument "log(base, x)" for floating point x.
-          build_sql("log(", x, ") / log(", base, ")")
-        }
-      }
+      sd     = sql_prefix("stddev_samp"),
+      log    = function(x, base = exp(1)) {
+                  if (isTRUE(all.equal(base, exp(1)))) {
```
hadley: The previous indentation was correct

hadley:
Can you please also add a bullet point to NEWS?

R/db-postgres.r
edgararuiz: So sorry, I'm not able to discern what I need to revert.  Should I put the indentation in the translation back how it was before this PR?

hadley: Yes please

javierluraschi:
@edgararuiz should we consider adding also `cov`, `cor` and `var`? Not sure if those are also supported by the SQL engines you are modifying this for, is that's the case?

edgararuiz:
It would be awesome if we could get it working over the entire tbl_sql/tbl_spark instead of having to manually pass each variable pairs.  I've been thinking about this since I started using sparklyr, my first instinct was to try and find a way to iterate through each column against all other columns, essentially trying to match how `cor()` works in R base.  From the sql perspective, it sounds like we just need a mechanism that identifies all of the current fields in the tbl_sql/tbl_spark, and then creates an `cor(a,b) as cor_ab` argument for each unique pair in the Select clause.  Maybe I'm oversimplifying it.

R/db-odbc-mssql.R
```diff
@@ -47,6 +47,7 @@
       atan2         = sql_prefix("ATN2"),
       ceil          = sql_prefix("CEILING"),
       ceiling       = sql_prefix("CEILING"),
+      sd            = sql_prefix("STDEV"),
```
hadley: I just realised that `sd` shouldn't appear here - it's a summary or window function, not a scalar function.

edgararuiz: I was thinking scalar could be used for cases when you want to run an operation against the result of the entire dataset, like this: 

```
> tbl(con, "iris") %>%
+   mutate(A_sd = sd(Sepal.Length),
+          s = Sepal.Length / sd(Sepal.Length)) %>%
+   select(Sepal.Length,
+          A_sd,
+          s)
# Database: Microsoft SQL Server
#   12.00.4422[rstudioadmin@EC2AMAZ-O5QKKM8/airontime]
   Sepal.Length      A_sd        s
          <dbl>     <dbl>    <dbl>
 1          5.1 0.8280661 6.158928
 2          4.9 0.8280661 5.917402
 3          4.7 0.8280661 5.675875
```

hadley: In that context, it's a window function.

hadley:
Thanks!

PR #23, opened 2017-06-27
Adds support for Redshift

edgararuiz:
I was able to test the translation in the data warehouse.  I'm not able to run the full `dbtest` because it will create a new table and I don't want to change things in the DW.  Thanks @stevenolen, @jimhester and @javierluraschi for getting the drivers to work in that RSP instance!


hadley:
Thanks!

PR #24, opened 2017-06-28
Add support for `CROSS JOIN` through `full_join` over non-overlapping columns

javierluraschi:
Fix https://github.com/tidyverse/dplyr/issues/2924 by adding support for cross joins.

Currently,

```r
d1_mem <- memdb_frame(a = 1:5)
d2_mem <- memdb_frame(b = 1:5)
d1_mem %>% full_join(d2_mem)
```

Triggers: `Error: `by` required, because the data sources have no common variables`.

This PR implements `CROSS JOINS` support for `full_join` over non-overlapping columns.


R/tbl-lazy.R
```diff
@@ -201,7 +201,14 @@ distinct_.tbl_lazy <- function(.data, ..., .dots = list(), .keep_all = FALSE) {
 add_op_join <- function(x, y, type, by = NULL, copy = FALSE,
                         suffix = c(".x", ".y"),
                         auto_index = FALSE, ...) {
-  by <- common_by(by, x, y)
+  by_intersect <- intersect(tbl_vars(x), tbl_vars(y))
```
ianmcook: This doesn't handle the case where `by` is a named vector that maps the names of the join columns in the left table to the join columns in the right table, as in:
```r
full_join(tbl_left, tbl_right, by = c("xl" = "xr"))
```

ianmcook:
Let's please discuss at https://github.com/tidyverse/dplyr/issues/2924 whether this is the best approach for implementing cross joins.

R/tbl-lazy.R
ianmcook: Thanks; https://github.com/tidyverse/dbplyr/pull/24/commits/6e5daa9bb7293b074c540946df1810ef46e59839 fixes it

hadley:
I think it's ok to merge this as a short term work around for the absence of `full_join()`.

R/sql-generic.R
```diff
@@ -51,12 +51,13 @@ sql_join.DBIConnection <- function(con, x, y, vars, type = "inner", by = NULL, .
...
     right = sql("RIGHT JOIN"),
     full = sql("FULL JOIN"),
+    cross = sql("CROSS JOIN"),
     stop("Unknown join type:", type, call. = FALSE)
   )
 
   select <- sql_join_vars(con, vars)
 
-  on <- sql_vector(
+  on <- if (length(by$x) + length(by$y) <= 0) NULL else sql_vector(
```
hadley: Can you please pull this out into:

```R
if (...) {
  on <- ...
} else {
  on <- ...
}
```

Or maybe even consider pulling out into own function?

R/tbl-lazy.R
```diff
@@ -201,7 +201,14 @@ distinct_.tbl_lazy <- function(.data, ..., .dots = list(), .keep_all = FALSE) {
...
+  by <- if (length(by_intersect) > 0 || !identical(type, "full") || !is.null(by)) {
```
hadley: Similarly here - I really don't like this style of `if`-`else` because it's easy to miss that the entire block gets assigned to some variable.

tests/testthat/test-joins.R
```diff
@@ -92,3 +92,8 @@ test_that("sql generated correctly for all sources", {
 
   expect_equal_tbls(xy)
 })
+
+test_that("full join is promoted to cross join for no overlapping variables", {
+  result <- df1 %>% full_join(df2) %>% collect()
```
hadley: I don't like making this the default because if you accidentally supply the wrong data frames you might create a query that takes hours to run. Could we make it explicit, i.e. `by = character()`?

javierluraschi: Changed.

R/tbl-lazy.R
javierluraschi: Changed.

R/tbl-lazy.R
javierluraschi: Thanks for the reference, no change here.

R/sql-generic.R
javierluraschi: Changed.

hadley:
Thanks!

PR #25, opened 2017-06-30
Adds custom db_compute for MS SQL

edgararuiz:
Fixes tidyverse/dplyr#2931


R/db-odbc-mssql.R
```diff
@@ -101,3 +101,37 @@
...
+                sql,
+                temporary = TRUE,
+                unique_indexes = list(),
+                indexes = list(),
+                ...) {
+
+
+  # The user will have to pass the pound sign, or signs, to setup the
+  # temp table. Two pound signs indicate a global temporary table, a
+  # single pound sign indicates a local temporary table
```
hadley: Could we paste on the `#` ourselves?

If you do that I think you could just call `NextMethod(temporary = FALSE)`

edgararuiz: I looked into that some time back.  The issue I wasn't able to overcome was that the table name retains the double pound, so the R users will either have to know to use the double pound prefix when calling the table via `tbl()` or we will have to somehow continue to prefix the name on all functions that make table references.  I was about to submit an update that returns an error if the table name is not prefixed with a pound.

hadley: In most cases, you don't need to know the name, because you'll take the tbl directly. 

We could also display a message notifying what the user is doing.

edgararuiz:
Fixes https://github.com/tidyverse/dplyr/issues/2931

Hong-Revo:
Just wondering, when is the next update for dbplyr due? I've just run into tidyverse/dplyr#2931 myself.

R/db-odbc-mssql.R
```diff
@@ -100,3 +100,36 @@
...
+
+  if (substr(table, 1, 1) != "#") {
+    table <- paste0("##", table)
+    message("MS SQL temp tables names require a single or double pound sign prefix. Renaming table to: ", table)
+    }
+
+  current_query <- dbplyr::sql(sql)
+  current_query <- sql_select(con, sql("*"), sql_subquery(con, current_query))
+  res <- dbGetQuery(con, current_query)
+  values <- sqlData(con, res[, , drop = FALSE])
```
Hong-Revo: I'm probably misunderstanding the code, but is this reading the entire table into memory as a data frame? I thought the benefit of the SQL dplyr backend was that it avoided materialising the data like this?

edgararuiz: HI @Hong-Revo , you're right, I'm trying to remember why I made this look like `copy_to` instead of just passing the query in the FROM statement.  Let me retest and will either add another reply, or update the PR.  Thanks for reviewing it!

Hong-Revo: I think this should do the trick. Rather than making a new PR, I'll just paste it here.
```r
#' @export
`db_compute.Microsoft SQL Server` <- function(con, table, sql, temporary=TRUE, unique_indexes=list(), indexes=list(), ...)
{
    # check that name has prefixed '##' if temporary
    if(temporary && substr(table, 1, 1) != "#")
        table <- paste0("##", table)

    if(!is.list(indexes))
        indexes <- as.list(indexes)

    if(!is.list(unique_indexes))
        unique_indexes <- as.list(unique_indexes)

    db_save_query(con, sql, table, temporary=temporary)
    db_create_indexes(con, table, unique_indexes, unique=TRUE)
    db_create_indexes(con, table, indexes, unique=FALSE)
    table
}


#' @export
`db_save_query.Microsoft SQL Server` <- function(con, sql, name, temporary=TRUE, ...)
{
    # check that name has prefixed '##' if temporary
    if(temporary && substr(name, 1, 1) != "#")
        name <- paste0("##", name)

    tt_sql <- build_sql("select * into ", ident_q(name), " from (", sql, ") ", ident_q(name), con=con)

    dbExecute(con, tt_sql)
    name
}
```

Hong-Revo: Note: may not be Bobby Tables-resistant. Testing is advised.

edgararuiz: Yes, I did some testing before committing the changes

R/db-odbc-mssql.R
```diff
@@ -102,34 +102,46 @@
...
+                                              sql,
+                                              temporary = TRUE,
+                                              unique_indexes = list(),
+                                              indexes = list(),
+                                              ...)
+{
+  # check that name has prefixed '##' if temporary
+  if(temporary && substr(table, 1, 1) != "#"){
     table <- paste0("##", table)
     message("MS SQL temp tables names require a single or double pound sign prefix. Renaming table to: ", table)
```
Hong-Revo: By default, `compute(foo)` with no other arguments will trip this message. Perhaps it might be good to add a check if the user explicitly speciified `temporary=TRUE`, so they don't get messages about something dplyr did by itself.

R/db-odbc-mssql.R
Hong-Revo: No worries, I just spotted something else. Hope to see this on CRAN soon!

R/db-odbc-mssql.R
edgararuiz: By virtue of prefixing `##` to the table name we are really not giving a choice to the end-user to set the temporary flag, I wonder if we should just remove it as an argument.  Besides, `compute()` is meant to create temporary tables.

Hong-Revo: Prefixing ## is done only if `temporary` is TRUE. This can be changed to FALSE in the call to `compute`.

The issue of whether to have `temporary` as an arg to `compute` is probably too broad for this PR, but:
- There is a need to be able to create persistent tables, not just temporary ones
- `compute` seems as good a place to do it as any
- People are already using `compute` in this way, so removing the option would be a breaking change in the API
- You want the R-level interface to SQL Server to be as consistent as possible with other databases; implementation details like # to signify a temp table should be invisible to the user.

thomascwells:
Hi all, relatively new here, so forgive me if this is a silly or rude question - how do I track down the status of this PR? 

It looks like the branch passed checks and there are no conflicts with the base. However, @Hong-Revo [says over on SO](https://stackoverflow.com/a/46267726) that the PR is glitchy. I used the code shared in that SO response in a local workflow, and it seems to be working for me. (Thanks for posting to SO!) Maybe the "glitch" referred to on SO has already been resolved? 

Is there anything I can do to help move this bug-fix / feature along into base? I expect to need this functionality increasingly often in the future.
Thanks!

Hong-Revo:
@thomascwells the main issue was that `db_compute` would pull the table into memory as an R data frame before writing it back. This should now be fixed. While waiting for the PR to be merged, you can keep using the SO code.

Alternatively, if you have Microsoft R Server/Machine Learning Server, you can install my [dplyrXdf package](https://github.com/RevolutionAnalytics/dplyrXdf). It uses the same code, and also provides several other features for working with MRS.

R/db-odbc-mssql.R
```diff
@@ -100,3 +100,45 @@
   )
   DBI::dbExecute(con, sql)
 }
+
+#' @export
+`db_compute.Microsoft SQL Server` <- function(con,
```
hadley: This seems to be identical to `db_compute.DBIConnection`?

R/db-odbc-mssql.R
```diff
@@ -100,3 +100,45 @@
...
+
+
+#' @export
+`db_save_query.Microsoft SQL Server` <- function(con,
+                                                 sql,
+                                                 name,
+                                                 temporary = TRUE,
+                                                 ...)
+{
+  # check that name has prefixed '##' if temporary
```
hadley: What's the difference between a prefix of `#` vs `##`?

R/db-odbc-mssql.R
```diff
@@ -100,3 +100,45 @@
...
+  if(temporary && substr(name, 1, 1) != "#") {
+    table <- paste0("##", table)
+    message("MS SQL temp tables names require a single or double pound sign prefix. Renaming table to: ", table)
```
hadley: temp -> temporary. And I think add a "\n" after the ".".

R/db-odbc-mssql.R
```diff
@@ -100,3 +100,45 @@
...
+  }
+
+  tt_sql <- build_sql("select * into ", ident_q(name), " from (", sql, ") ", ident_q(name), con=con)
```
hadley: * I think `ident_q` should be `as.sql` now
* Can you please put spaces around the `=`

R/db-odbc-mssql.R
edgararuiz: I fixed `temporary` to TRUE in `db_save_query`

edgararuiz: In `db_compute.DBIConnection`, `db_save_query` has `temporary = temporary`

R/db-odbc-mssql.R
edgararuiz: A `#` creates a local temp table, and `##` creates a global temp table.  I'm trying to recall why I choose global,  I do remember that I tested it both ways, and I think local gave me errors. https://technet.microsoft.com/en-us/library/ms186986(v=sql.105).aspx#Anchor_1

R/db-odbc-mssql.R
hadley: Don't we want that?

R/db-odbc-mssql.R
```diff
@@ -100,3 +100,45 @@
...
+
+#' @export
+`db_save_query.Microsoft SQL Server` <- function(con,
+                                                 sql,
+                                                 name,
+                                                 temporary = TRUE,
+                                                 ...)
+{
+  # check that name has prefixed '##' if temporary
+  if(temporary && substr(name, 1, 1) != "#") {
```
hadley: I think this should be `table`, not `name`

R/db-odbc-mssql.R
edgararuiz: This change was based on my initial understanding of `compute()`.  I'll remove that customization

Hong-Revo: `db_compute.Microsoft SQL Server` should have some lines added to prefix # to the table name. See my comment above dated 18 Sep. Not sure why/when it got dropped.

This is my original version from the comment:

```r
#' @export
`db_compute.Microsoft SQL Server` <- function(con, table, sql, temporary=TRUE, unique_indexes=list(), indexes=list(), ...)
{
    # check that name has prefixed '##' if temporary
    if(temporary && substr(table, 1, 1) != "#")
        table <- paste0("##", table)

    if(!is.list(indexes))
        indexes <- as.list(indexes)

    if(!is.list(unique_indexes))
        unique_indexes <- as.list(unique_indexes)

    db_save_query(con, sql, table, temporary=temporary)
    db_create_indexes(con, table, unique_indexes, unique=TRUE)
    db_create_indexes(con, table, indexes, unique=FALSE)
    table
}
```

edgararuiz: That's correct.   If the user passes `compute()`, meaning w/o a name, then that name will generate a "##" warning.  So the prefix needs to be done silently.  Additionally, the change needs to happen in `db_compute()` because the `table` variable will now have a different name.

R/db-odbc-mssql.R
edgararuiz: Right, it was the other way around, I had to change everything to `name`

R/db-odbc-mssql.R
edgararuiz: @Hong-Revo - Can you check my latest commit to make sure I restored back your changes.  I re-tested the changes against a MS SQL connection and everything worked fine

R/db-odbc-mssql.R
```diff
@@ -100,3 +100,41 @@
...
+    name <- paste0("##", name)
+    message("MS SQL temporary tables names require a single or double pound sign prefix.\nRenaming table to: ", name)
```
hadley: And now that I look at this again, " single or double pound sign prefix." -> "A prefix of # or ##" is probably easier to understand.

R/db-odbc-mssql.R
```diff
@@ -100,3 +100,41 @@
   )
   DBI::dbExecute(con, sql)
 }
+
+#' @export
+`db_compute.Microsoft SQL Server` <- function(con, table, sql, temporary=TRUE, unique_indexes=list(), indexes=list(), ...)
+{
+  # check that name has prefixed '##' if temporary
+  if(temporary && substr(table, 1, 1) != "#")
```
hadley: I think it would be better to modify `db_compute.DBIConnection` to have:

```R
table <- db_save_query(con, sql, table, temporary = temporary)
```

edgararuiz:
The initial message was a bit confusing if I used `compute()`, so I tried a generic message:

```
t1 <- tbl(con, "mtcars") %>%
+   filter(am == 0) %>%
+   compute()
Created temporary table named: ##zldgnauvck
```

```

**Comment by edgararuiz** [comment]
PTAL @hadley

File: R/db-odbc-mssql.R (hunk 10 above)
Comment (Hong-Revo) [reply R/db-odbc-mssql.R]: Hm, could this affect other databases?

File: R/db-odbc-mssql.R [hunk 11]
```diff
@@ -100,3 +100,22 @@
...
+                                                 temporary = TRUE,
+                                                 ...)
+{
+  # check that name has prefixed '##' if temporary
+  if(temporary && substr(name, 1, 1) != "#") {
+    name <- paste0("##", name)
+    message("Created temporary table named: ", name)
+  }
+
+  tt_sql <- build_sql("select * into ", as.sql(name), " from (", sql, ") ", as.sql(name), con = con)
```
hadley: Should `name` appear twice in this query?

edgararuiz: Yes, the last instance it the table alias name for the query.  I just tried it w/o it and the operation failed

hadley: It seems slightly odd that's valid SQL, but if it works it's fine.

hadley:
Thanks everyone!

R/db-odbc-mssql.R
krlmlr: See https://github.com/r-dbi/odbc/issues/127 for the upstream issue.

PR #26, opened 2017-07-03
Translation jdbc

edgararuiz:
@hadley  - An issue in dplyr (https://github.com/tidyverse/dplyr/issues/2939) made me realize that JDBC connections are defaulting to `DBIConnection` for their translation, so  instead of supplying and error message, the change defaults to `DBIConnection`.  It looks like there are some databases that have not been in our purview, like Athena, that are fine using default translations.


R/sql-jdbc.R
```diff
@@ -0,0 +1,49 @@
+#' @export
+sql_translate_env.JDBCConnection <- function(x, ...) {
+  class <- x@jc
+
+  if (grepl("oracle.jdbc.driver", class)) {
+    sql_translate_env.OracleConnection(x, ...)
```
ianmcook: Isn't this supposed to be `.Oracle` not `.OracleConnection`?

ianmcook:
Since RJDBC does not set the class of the connection object to the DBMS name the way odbc does, this is a general problem, not only specific to Oracle. It would be good to have a more general solution.

But RJDBC has numerous other problems as well, for example the lack of a `dbExecute()` method and poor type conversion. @hadley is it your vision that dbplyr should provide workarounds for RJDBC deficiencies?

R/sql-jdbc.R
edgararuiz: Good catch, thanks @ianmcook !

hadley:
The intention is definitely not to try and shim over JDBC issues in dplyr.

Maybe a better approach is to have a custom class that lets you wrap over any DBI connection and let you supply the translation that you want to use.

edgararuiz:
Should I close this PR?

hadley:
Yes, please. And file an issue to make translation choice more flexible?

edgararuiz:
Will do

PR #27, opened 2017-07-04
Fixes Oracle LIMIT and adds ROracle support

edgararuiz:
Fixes #https://github.com/tidyverse/dplyr/issues/2928 

1. An issue was discovered while working out how to add ROracle support. The clause used currently to perform the LIMIT is only available for Oracle version 12c and is not backwards compatible.  The alternative is to a technique found here: https://oracle-base.com/articles/misc/top-n-queries.  We need to limit the rows in a WHERE clause, so had to use a subquery to fix.

2. Mapped all of the methods from the Oracle to the ROracle class (`OraConnection`).  I also mapped some `OdbcConnection` methods as an attempt to provide as much functionality as possible, meaning `copy_to()` should work now.


tests/testthat/test-translate-oracle.R
```diff
@@ -21,11 +21,6 @@ test_that("as.numeric() translated to NUMBER ", {
 df <- data.frame(x = 1, y = 2)
 df_oracle <- tbl_lazy(df, src = simulate_oracle())
 
-test_that("query uses FETCH FIRST x ROWS instead of LIMIT ", {
```
hadley: Can you update this test please?

edgararuiz:
So far, so good : https://github.com/tidyverse/dplyr/issues/2928#issuecomment-313110394

bthe:
Any news on this pull request?

R/db-roracle.R
```diff
@@ -0,0 +1,53 @@
...
+  sql_select.Oracle(con, select, from, where = where,
+                               group_by = group_by, having = having,
+                               order_by = order_by,
+                               limit = limit,
+                               distinct = distinct)
+}
+
+#' @export
+db_analyze.OraConnection <- function(con, table, ...) {
+  db_analyze.Oracle(con = con, table = table)
```
hadley: Can you please pass `...` down through all these functions? That keeps it forward compatible

hadley:
Thanks @edgararuiz!

thomastallaksen:
This fix no longer works in newer versions of dbplyr.

hadley:
I don't have the resources to support very old versions of Oracle, and so dbplyr is only compatible with versions that have still extended support. Extended support for 11.2 finished in Dec 2020.

PR #28, opened 2017-07-14
Feature: prevent select from adding names unnecessarily

DavisVaughan:
### Details

This PR addresses tidyverse/dplyr#2968, which also appeared in tidyverse/dplyr#2946. 

`names_to_as()` now prevents select from adding names unnecessarily. 

### Note

@hadley, note that this is slightly different from what you suggested. The first half of the `|` condition is different: `names == ""` compared with `names_esc == ""`. The reason for this is that when `names` is just `""`, it is escaped differently depending on whether a connection is passed to `sql_escape_ident()`. 

If an Access connection is passed: `""` -> ``` "``" ```
If no connection is passed: `""` -> `"\"\""`

The usage of `names == ""` is the same as the original implementation, maybe that was the original reasoning as well.


NEWS.md
```diff
@@ -1,5 +1,6 @@
 # dbplyr 1.1.0.9000
 
+* `names_to_as()` now prevents select from adding names when not necessary (#2968) (@DavisVaughan)
```
hadley: Can you please rewrite to be a bit more user focussed? i.e. ``Queries generated by `select()` no longer ...``

DavisVaughan: How about:

```
* Queries generated by `select()` no longer alias column names 
  unnecessarily (#2968) (@DavisVaughan).
```

Would you like to have an example to explain further?

```
* Queries generated by `select()` no longer alias column names unnecessarily:
  ``SELECT `col_name` ``, not ``SELECT `col_name` AS `col_name` `` 
  (#2968) (@DavisVaughan).
```

This renders correctly when I preview NEWS, but looks kind of ugly because of the double backticks.

hadley: I think the first is fine

DavisVaughan:
I added two tests. One checks that `select()` generates correct SQL and no aliasing is done. The other is lower level and checks that `names_to_as()` is working correctly (it basically simulates relevant pieces of `escape.ident()`).

I'm not sure what the best way to check for the old behavior is. I'm happy to add any additional tests with a bit more instruction.

tests/testthat/test-sql-escape.r
```diff
@@ -42,3 +42,27 @@ test_that("times are converted to ISO 8601", {
...
 })
+
+# Name aliasing -----------------------------------------------------------
+
+test_that("queries generated by select() don't alias unnecessarily", {
+  lf_build <- lazy_frame(x = 1) %>%
+    select(x) %>%
+    sql_build()
+
+  render_odbc <- sql_render(lf_build, con = simulate_odbc())
```
hadley: I don't think you need to test both - you're effectively just testing `sql_escape_ident()`

tests/testthat/test-sql-escape.r
```diff
@@ -42,3 +42,27 @@ test_that("times are converted to ISO 8601", {
...
+  render_dbi  <- sql_render(lf_build, con = simulate_dbi())
+
+  expect_equal(render_odbc, sql("SELECT `x`\nFROM `df`"))
+  expect_equal(render_dbi,  sql("SELECT \"x\"\nFROM \"df\""))
+})
+
+test_that("names_to_as() doesn't alias when ident name and value are identical", {
```
hadley: Could you add a similar test for when the names are missing?  And one to make sure that the names are used, when present? That's what I meant by testing the existing behaviour as well.

tests/testthat/test-sql-escape.r
DavisVaughan: I removed the odbc/dbi duplication

tests/testthat/test-sql-escape.r
DavisVaughan: Added 2 tests. I was misunderstanding what you meant by existing behavior. Thanks for the clarification!

hadley:
Thanks!!

PR #29, opened 2017-07-17
Update README.md

larspijnappel:
Install of `RSQLite` added so that the provided example can be run


hadley:
This is unfortunately in the wrong file, but generally I assume the user can install packages needed for examples.

PR #34, opened 2017-08-18
Fix na mssql

edgararuiz:
Fixes tidyverse/dplyr#3007
Fixes tidyverse/dplyr#2940


edgararuiz:
I tested `is.na()`, `!is.na()` and `head()` with the four verbs in a live connection.  Here are the cases where the solution does not work:

arrange - `!is.na()`, `head()`
summarise - `!is.na()`
mutate - `!is.na()`
filter - all worked

I decided to sent the PR with these results because a solution would probably require modifying `sql_prefix()` which seemed a bit too much because the cases where the solution fails do not seem to be a common use of those of the verb+vector function.

R/db-odbc-mssql.R
```diff
@@ -100,3 +94,14 @@
   )
   DBI::dbExecute(con, sql)
 }
+
+mssql_is_null <- function(x, sql_context){
+  if ( !is.null(sql_context$mutate) | !is.null(sql_context$arrange) ) {
```
hadley: I think it would be better to have parameters `x` and `op`.

The indenting also looks a bit off

R/sql-build.R
```diff
@@ -61,15 +61,15 @@ sql_build.op_rename <- function(op, con, ...) {
 
 #' @export
 sql_build.op_arrange <- function(op, con, ...) {
-  order_vars <- translate_sql_(op$dots, con)
+  order_vars <- translate_sql_(op$dots, con, sql_context = list(arrange = TRUE))
```
hadley: I think this would all be simpler if used one variable name everywhere - i.e. `sql_context = list(op = "arrange")`

R/db-odbc-mssql.R
hadley: Can you please also add a comment documenting the issue? (My comments in slack about bit vs boolean will probably be most helpful to me)

R/sql-build.R
```diff
@@ -86,8 +86,9 @@ sql_build.op_mutate <- function(op, con, ...) {
   new_vars <- translate_sql_(
     op$dots, con,
     vars_group = op_grps(op),
-    vars_order = translate_sql_(op_sort(op), con),
-    vars_frame = op_frame(op)
+    vars_order = translate_sql_(op_sort(op), con, sql_context = list(mutate = TRUE)),
```
hadley: Shouldn't this be "arrange"?

R/sql-build.R
hadley: Or maybe `clause = "ORDER`? Might be better to stick with SQL naming.

R/translate-sql-window.r
```diff
@@ -194,6 +197,14 @@ win_current_frame <- function() sql_context$frame
...
 sql_current_con <- function() sql_context$con
 
+# Functions to manage information for special cases
+set_current_context <- function(context) {
+  old <- sql_context$context
+  sql_context$context <- context
+  invisible(old)
+}
+
+sql_current_context <- function() sql_context$context
```
hadley: I think should be exported

R/translate-sql.r
```diff
@@ -108,7 +109,8 @@ translate_sql_ <- function(dots,
                            vars_group = NULL,
                            vars_order = NULL,
                            vars_frame = NULL,
-                           window = TRUE) {
+                           window = TRUE,
+                           sql_context = NULL) {
```
hadley: I think this should be just `context`

R/translate-sql.r
```diff
@@ -123,6 +125,11 @@ translate_sql_ <- function(dots,
   old_con <- set_current_con(con)
   on.exit(set_current_con(old_con), add = TRUE)
 
+  if ( !is.null(sql_context) ) {
```
hadley: Spacing isn't quite right here - see http://style.tidyverse.org/syntax.html#spacing

tests/testthat/test-translate-mssql.r
```diff
@@ -139,8 +125,38 @@ test_that("trimws() translates to LTRIM-RTRIM ", {
...
+    sql("SELECT *\nFROM `df`\nWHERE (NOT(((`x`) IS NULL)))"))
+})
+
+test_that("Arrange uses CONVERT-BIT for is.na ", {
+  expect_equivalent(
+    show_query(arrange(df_mssql, !is.na(x))),
+    sql("SELECT *\nFROM `df`\nORDER BY NOT(CONVERT(BIT, IIF(`x` IS NULL, 1, 0)))"))
+})
+
+
```
hadley: Can you please delete these extra lines?

R/translate-sql-window.r
edgararuiz: Ok, should I set the rdname to `win_over`

R/sql-build.R
edgararuiz: I just noticed that I named the `filter` context `where` instead of `filter`, so you're right, it makes more sense.  It's also easier to align them to the sql variable that is being loaded by `sql_translate_`, like `where_sql`

R/translate-sql-window.r
edgararuiz: I take that back.  I just added a quick description for it

hadley: It should be documented in the same file as all the other `sql_current_` functions

edgararuiz: Ok, the only other `sql_current_` function I could find is `sql_current_con` and is not exported/documented: 

https://github.com/tidyverse/dbplyr/blob/2c93c9d37cf56d7dc331df02699739f49064de50/R/translate-sql-window.r#L194-L195

hadley: Oh I was thinking of the `win_` functions. In that case, let's not export right now.

edgararuiz: Ok

juliasilge:
What's the status of this PR? I sure would love to have this updated.

edgararuiz:
Hi @juliasilge , if you like, you can install the PR's version of the package: `devtools::install_github("tidyverse/dbplyr", ref = devtools::github_pull(34))`

R/translate-sql-odbc.R
```diff
@@ -45,7 +45,10 @@ base_odbc_agg <- sql_translator(
 #' @rdname sql_variant
 #' @format NULL
 base_odbc_win <- sql_translator(.parent = base_win,
-  sd            = win_recycled("STDDEV_SAMP")
+  sd            = win_recycled("STDDEV_SAMP"),
```
hadley: Can you please remove this since it's unrelated to the rest of the PR?

R/translate-sql.r
```diff
@@ -108,7 +109,8 @@ translate_sql_ <- function(dots,
                            vars_group = NULL,
                            vars_order = NULL,
                            vars_frame = NULL,
-                           window = TRUE) {
+                           window = TRUE,
+                           context = NULL) {
```
hadley: Can you please use `list()` instead of `NULL` here?

R/translate-sql.r
```diff
@@ -36,6 +36,7 @@
 #' @param window Use `FALSE` to suppress generation of the `OVER`
 #'   statement used for window functions. This is necessary when generating
 #'   SQL for a grouped summary.
+#' @param context Use to carry information for special translation cases.  Expects a list.
```
hadley: For example, MS SQL needs a different conversion for `is.na()` in `WHERE` vs. `SELECT` clauses.

nwstephens:
I want to add my vote. This PR would be useful to have merged.

hadley:
Thanks @edgararuiz!

PR #36, opened 2017-08-27
SQL translation - Microsoft Access

DavisVaughan:
### Details

This pull request addresses tidyverse/dplyr#2946 by providing an Access SQL translation. 

This branch was branched from another pull request that is still open, #28, because it required the updated `names_to_as()` function there.

Because of problems with writing tables discussed in error 4 below, this is intended to be the _read-only_ piece of the translation until the write errors can be sorted out. Fixing these errors requires modification to the `odbc` package.

### Known errors

These still need to be addressed or added to the db.rstudio.com page.

1) When joining two tables with the same column names, you have to specify `suffix = c("_x", "_y")` over the default of `suffix = c(".x", ".y")` because a `.` cannot exist in Access column names.

2) `full_join` uses the `coalesce` function in the SQL that it builds. This is not supported in Access so it does not work. There is an alternative of using `IFF(Is.Null(x), y, x)` but the `sql_coalesce()` function in `dbplyr` is not generic and I don't think it ever sees the connection object. That alternative solution would also only work with a max of two tables being "coalesced" at once (probably fine with joins but maybe a limitation elsewhere). 

3) While a translation was provided for `dplyr::coalesce()`, `pmin()` and `pmax()`, they only work when comparing a max of 2 objects. The translations were created using variations of `IIF`, for example `pmin` is ``IIF(`field_name_x` <= `field_name_y`,`field_name_x`,`field_name_y`)`` and I am not sure how to easily generalize this.

4) While working with the database, I discovered three `db_*` functions that don't work, and I rolled my own (rough) versions that do. I've created a Gist with that code [here](https://gist.github.com/DavisVaughan/7d8dd9338afcd5c2eb94d751097f5296) and we can talk about including it too if needed. 

     1. `db_drop_table` - Originally uses `IF EXISTS` if "forcing" the drop is required. Access doesn't support this. Rewrote to ask the user if they want to drop, and require them to specify `force = TRUE` if they do. 

     2. `db_copy_to` - As of now, temporary tables aren't supported in Access, so I had to turn that option off.

     3. `db_write_table` - This is a big one. Already posted [here](https://github.com/rstats-db/odbc/issues/79), but `dbWriteTable()` does not yet work with Access `odbc` connections. Fortunately, I found Hadley's solution for [writing tables to MySQL](https://github.com/tidyverse/dbplyr/blob/master/R/db-mysql.r#L80) and Access can actually do a similar thing!


hadley:
Can you please merge/rebase?

DavisVaughan:
This should be a clean rebase on top of the changes in master since the PR was opened.

R/db-odbc-access.R
```diff
@@ -0,0 +1,240 @@
...
+                     build_sql(
+                       "RIGHT(LEFT(", x, ", ", left, "), ", right, ")")
+                   },
+                   trimws        = sql_prefix("TRIM"),
+                   # No support for CONCAT in Access
+                   paste         = sql_not_supported("paste()"),
+                   paste0        = sql_not_supported("paste0()"),
+
+                   # Logic
+                   # ISNULL() returns -1 for True and 0 for False
```
hadley: Does access use `-1` for `TRUE` everywhere?

R/db-odbc-access.R
```diff
@@ -0,0 +1,240 @@
...
+                   },
+
+                   # Coalesce doesn't exist in Access.
+                   # NZ() only works while in Access, not with the Access driver
+                   # IIF(ISNULL()) is the best way to construct this
+                   coalesce      = function(x, y) {
+                     build_sql("IIF(ISNULL(", x, "),", y, ",", x, ")")
+                   },
+
+                   # pmin/pmax for 2 columns
```
hadley: I think handling for only two cols is fine

R/db-odbc-access.R
```diff
@@ -0,0 +1,240 @@
...
+
+                   # Count
+                   # Count(Distinct *) does not work in Access
+                   # This would work but we don't know the table name when translating:
+                   # SELECT Count(*) FROM (SELECT DISTINCT * FROM table_name) AS T
+                   n_distinct    = sql_not_supported("n_distinct")
+    ),
+
+    # Window functions not supported in Access
+    sql_translator(.parent = base_odbc_win,
```
hadley: I think you can just use `base_no_win` here

tests/testthat/test-translate-access.r
```diff
@@ -0,0 +1,398 @@
+context("translate-ACCESS")
+
+# Access base_scalar conversions -----------------------------------------
+
+# Conversion
+
+test_that("as.numeric() translated to CDbl ", {
```
hadley: Can you please reduce to one test per category? i.e. put all of the coercion functions in a single test.

And generally, I don't think it's worthwhile to test the straightforward translations.

R/db-odbc-access.R
```diff
@@ -0,0 +1,240 @@
...
+                     build_sql(x, " ^ ", y)
+                   },
+
+
+                   # Trig
+                   # Most implementations come from: https://en.wikibooks.org/wiki/Visual_Basic/Simple_Arithmetic#Trigonometrical_Functions
+                   # atan / sin / cos are the only native trig functions with access. Use trig properties to generate others.
+
+                   # Catch divide by zero when x = -1 or 1 and return 0 or Pi
+                   acos          = function(x) {
```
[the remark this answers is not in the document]
DavisVaughan: Understandable. I removed them all and their tests.

R/db-odbc-access.R
DavisVaughan: Yes, it seems Access tries to mimic a `bit` field, and the two possible integer representations are either `0` for False and `1` for True. 
https://superuser.com/questions/694919/why-do-yes-no-values-in-microsoft-access-map-to-1-and-0
https://stackoverflow.com/questions/8827447/why-is-yes-a-value-of-1-in-ms-access-database

R/db-odbc-access.R
DavisVaughan: I have changed it to `base_no_win` and removed everything else, but am now getting:

```r
Warning message:
Translator is missing window functions:
cor, count, cov 
```

They don't exist in `base_no_win`. Should they?

tests/testthat/test-translate-access.r
DavisVaughan: I removed most of the straightforward tests and kept only the special cases. Also reduced to 1 per category.

R/db-odbc-access.R
```diff
@@ -0,0 +1,240 @@
...
+                   # Pull `left` chars from the left, then `right` chars from the right to replicate substr
+                   substr        = function(x, start, stop){
+                     right  <- stop - start + 1
+                     left   <- stop
+                     build_sql(
+                       "RIGHT(LEFT(", x, ", ", left, "), ", right, ")")
+                   },
+                   trimws        = sql_prefix("TRIM"),
+                   # No support for CONCAT in Access
+                   paste         = sql_not_supported("paste()"),
```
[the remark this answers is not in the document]
DavisVaughan: This is either more challenging than I originally thought, or I am missing something easy. 

```r
sql_paste <- function(..., sep = " ", collapse = NULL) {
  binary_build <- function(x, y) {
    build_sql(x, " & ", sep, " & ", y)
  }

  .dots <- list(...)
  last_dot <- length(.dots)

  if(last_dot == 1) {
    build_sql("CStr(", .dots[[1]], ")")
  } else {
    reduce(.dots, binary_build)
  }

}

sql_paste0 <- function(..., collapse = NULL) {
  ex_paste(..., sep = "")
}
```

Something like this does work. The challenge is building correct SQL for an arbitrary number of inputs. This `reduce()` method works, but if you have other suggestions I'm open to them.

R/db-odbc-access.R
DavisVaughan: Also it seems this isn't entirely true. `Avg`, `Count`, `Min`, `Max`, and `Sum` are [supported aggregate functions](https://www.techonthenet.com/access/functions/numeric/avg.php) but use `GROUP BY` in the sql generated:

```r
dbGetQuery(con, "SELECT cyl, avg(`carb`) AS avg_carb FROM mtcars GROUP BY cyl")
```

This works fine. Note that in the `SELECT` you can only include columns that are either in the `GROUP BY` or wrapped in one of the 5 aggregate functions. So this with `mpg` would not work:

```r
dbGetQuery(con, "SELECT cyl, mpg, avg(`carb`) AS avg_carb FROM mtcars GROUP BY cyl")
```

I can see that `mtcars %>% group_by(cyl) %>% mutate(mean(carb))` would produce SQL with `OVER (PARTITION ...)` like this:

```r
SELECT `mpg`, `cyl`, `disp`, `hp`, `drat`, `wt`, `qsec`, `vs`, `am`, `gear`, `carb`, avg(`carb`) OVER (PARTITION BY `cyl`) AS `mean(carb)`
FROM `mtcars`
```

Is there an easy way to change that to use the Group By syntax?

hadley: I don't think you need to supply `count()`. 

Can you please add `cov()` and `cor()` to `base_no_win`.

R/db-odbc-access.R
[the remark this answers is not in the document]
hadley: I'd just do 

```
  if (!is.null(collapse)) {
    stop("`collapse` isn't supported in SQL translation", call. = FALSE)
  }
```

Since a non-NULL `collapse` turns it from a scalar function to an aggregate function.

Otherwise, the `reduce()` looks fine. I think `last_dot` would be better called `n`.

Can you please call the function something like `sql_paste_access()` and define outside of the list?

R/db-odbc-access.R
hadley: In that case, I don't think you should translate to 1 and 0 here. Instead, I think you should define a custom `sql_escape_logical` method.

R/db-odbc-access.R
DavisVaughan: After adding `cov()` and `cor()` the `count()` warning went away as well 👍

R/db-odbc-access.R
DavisVaughan: Trying to get my head around this. Are you suggesting this?

```r
is.null       = function(x){
    build_sql("IIF(ISNULL(", x ,"), ", TRUE, ", ", FALSE, ")")
},

is.na         = function(x){
    build_sql("IIF(ISNULL(", x ,"), ", TRUE, ", ", FALSE, ")")
},
```

And a `sql_escape_logical.ACCESS` like:

```r
sql_escape_logical.ACCESS <- function(con, x) {
  y <- as.character(as.integer(x))
  y[is.na(x)] <- "NULL"
  y
}
```

Now this happens:

```r
> ct %>% mutate(x = is.na(speed))
# Database: ACCESS 14.00.0000[admin@ACCESS/C:\Users\Administrator\Desktop\R\dbplyr-testing\dplyr-test-db.accdb]
   speed  dist     x
   <dbl> <dbl> <int>
 1    NA     2     1
 2     4    10     0
 3     7     4     0
 4     7    22     0
 5     8    16     0
 6     9    10     0
 7    10    18     0
 8    10    26     0
 9    10    34     0
10    11    17     0
# ... with more rows
```

From a translation of:

```r
> ct %>% mutate(x = is.na(speed)) %>% show_query()
<SQL>
SELECT `speed`, `dist`, IIF(ISNULL(`speed`), 1, 0) AS `x`
FROM `carstest`
```

hadley: I was thinking

```R
is.null       = sql_prefix("isnull")

sql_escape_logical.ACCESS <- function(con, x) {
  y <- ifelse(x, -1, 0)
  y[is.na(x)] <- "NULL"
  y
}
```

hadley: I'm just concerned that operations like `x == y` will also yield -1 and 0, so masking for just a few ops is going to be confusing.

R/db-odbc-access.R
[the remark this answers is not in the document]
DavisVaughan: Done!

R/db-odbc-access.R
DavisVaughan: Ah I see. Now it is always consistent with what Access returns. I changed it and added a test.

DavisVaughan:
Just did a `devtools::install_github()` with the merged PR. I wasn't getting this with `load_all()` before, but it seems like you do need `count()` in the `base_no_win`?

```r
Warning message:
Translator is missing window variants of the following aggregate functions:
* count
```

PR #40, opened 2017-10-06
Enable db_write_table() to append to tables, (per deprecation)

drf5n:
This passes the `...` argument from `db_write_table(con, table, types, values, temporary = FALSE, ...)` along into the `dbWriteTable()` call to enable arguments like `append=TRUE` per https://github.com/tidyverse/dplyr/issues/3120

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # dbplyr 1.1.0.9000
 
-* Fixed `db_write_table()` to allow `append=TRUE`.  Closes tidyverse/dplyr#3120 (@drf5n).
+* Fixed `db_write_table()` to allow `append=TRUE` (@drf5n, #3120).
 
 * Fixed `n()` and `count()` for window function via `odbc` connections (@edgararuiz)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dbplyr 1.1.0.9000
 
+* Fixed `db_write_table()` to allow `append=TRUE`.  Closes tidyverse/dplyr#3120 (@drf5n).
+
 * Fixed `n()` and `count()` for window function via `odbc` connections (@edgararuiz)
 
 * SQL translation for Redshift (@edgararuiz)
--- a/R/dbi-s3.r
+++ b/R/dbi-s3.r
@@ -45,7 +45,8 @@ db_write_table.DBIConnection <- function(con, table, types, values, temporary = 
     value = values,
     field.types = types,
     temporary = temporary,
-    row.names = FALSE
+    row.names = FALSE,
+    ...
   )
 }
 


hadley:
Thanks! Can you please add a bullet to news too?

And include `Closes tidyverse/dplyr#3120` in a commit it's automatically closed when I merge

cc @bborgesr

drf5n:
I think I did it.  I initiated the change with the github 'edit' link on the main code, which automatically forked a copy, built the patch-1 branch, and initiated the pull request.  Per your note, I went back to my branch, added a bullet to https://github.com/drf5n/dbplyr/blob/patch-1/NEWS.md file and committed it, and it seems to be re-running the CI checks here--just passed.

I wasn't sure where to add the bullet, so I added under the existing first heading, but it would be easy to move elsewhere if you like.

Did you mean that I need the text `Closes tidyverse/dplyr#3120` in a commit message?

hadley:
Right - the news bullet just needs `(@drf5n, #3120)`, and the full `Closes tidyverse/dplyr#3120` needs to go in a commit message. But since you're using github and it'll be difficult to add after the fact, I can do it when I merge the PR.

drf5n:
I think the github method worked.  I edited my fork's patch branch's https://github.com/drf5n/dbplyr/blob/master/NEWS.md file once more and put that text in as the commit message, and it seem to have added it to this PR.

I'm impressed with how well this github method for simple patches seems to work.  It is a much lower bar than forking a local copy, branching, patching, pushing to gitub, pull-requesting, etc.  The trick was doing additional edits & commits directly on the github fork&branch.  And now I see a little message above the CI block that hints at the process: "Add more commits by pushing to the patch-1 branch on drf5n/dbplyr."

hadley:
Yeah, the gh interface for this is now pretty slick - hopefully we'll be able to write this up somewhere and propose it as recommended approach for people's first PR.

drf5n:
Yes, a one-file change is easy and almost intuitive to make into a Pull Request. But the additional requirement of editing the contributor's forked branch's NEWS.md file to add a properly formatted bullet like:

    * Fixed xyzzy to handle plugh (@contributor #issuenumber) 

with a commit message of:

    Closes tidyverse/dplyr#3120 

is a little trickier.  

Github's language for the additional editing is "pushing commits to the patch branch of your fork".  So, you'd navigate to your fork of the code on github, choose the patch branch (other than the default master), and then make and commit your additional edits.  When the commits are made, they are automatically added to the pull request.

bborgesr:
@hadley, this looks good, but what about the MySQL (https://github.com/tidyverse/dbplyr/blob/master/R/db-mysql.r#L80-L101) and Postgres (https://github.com/tidyverse/dbplyr/blob/master/R/db-postgres.r#L104-L120) implementations? Don't we need to change those as well?

drf5n:
Or before that, does the deprecation of `db_insert_table` and `db_create_table` in https://github.com/tidyverse/dplyr/blob/f9096e9c4d47b8da047aa88f05a064c2d521e281/docs/reference/backend_dbplyr.html#L213 make sense before these capabilities exist in `db_write_table()`?

hadley:
Yeah, I think the right solution is to un-deprecate those functions. Thanks for the discussion everyone, and sorry it's taken so long for me to get to this.

PR #41, opened 2017-10-10
URLs to Securing Credentials and PostgreSQL docs

dpprdan:
Added links to "Securing Credentials" on db.rstudio.com. 
Changed links to PostgreSQL docs from v9.3 to "current".


hadley:
Thanks!

PR #42, opened 2017-10-13
Fix error: object 'port' not found

hoxo-m:
Fix an error for OdbcConnection:

```r
> library(dplyr)
> library(dbplyr)
> library(odbc)
> con <- dbConnect(odbc(), dsn = "test")
> con
<OdbcConnection>  Database: test
> db_desc(con)
Error in paste0(":", port) : object 'port' not found
```


hadley:
Thanks!

PR #43, opened 2017-10-19
Adds Teradata translation

edgararuiz:
This change includes the new Teradata translation, and its corresponding tests.  

The new translations was tested using `dbtest`, here are the results:

teradata-test

Fixes: https://github.com/tidyverse/dplyr/issues/3040


R/translate-sql-odbc.R
```diff
@@ -9,6 +9,7 @@ base_odbc_scalar <- sql_translator(.parent = base_scalar,
   as.character  = sql_cast("STRING"),
   as.Date       = sql_cast("DATE"),
   paste0        = sql_prefix("CONCAT"),
+  sd            = sql_prefix("STDDEV_SAMP"),
```
hadley: Are you sure this is ok to add here?

R/db-odbc-teradata.R
```diff
@@ -0,0 +1,94 @@
...
+      atan2         = function(x, y){
+                          build_sql(
+                            "ATAN2(", y, ",", x, ")"
+                          )},
+      substr        = function(x, start, stop){
+                        len <- stop - start + 1
+                        build_sql(
+                          "SUBSTR(", x, ", ", start, ", ", len, ")"
+                        )},
+      paste         = sql_not_supported("paste()")
```
hadley: https://forgetcode.com/Teradata/434-CONCAT ?

R/db-odbc-teradata.R
```diff
@@ -0,0 +1,94 @@
...
+      as.numeric    = sql_cast("NUMERIC"),
+      as.double     = sql_cast("NUMERIC"),
+      as.character  = sql_cast("VARCHAR(MAX)"),
+      var           = sql_prefix("VAR_SAMP"),
+      log10         = sql_prefix("LOG"),
+      log           = function(x, base = exp(1)) {
+                        if (isTRUE(all.equal(base, exp(1)))) {
+                          build_sql("ln(", x, ")")
+                        } else {
+                          # Use log change-of-base because postgres doesn't support the
```
hadley: Is this true for teradata too? If so we should pull out into a resuable function

R/db-odbc-teradata.R
edgararuiz: Right, I use `CONCAT` for `paste0`.  `CONCAT` does not add a separator like `paste` does by default.
https://github.com/tidyverse/dbplyr/blob/d08d93071606db414b44f8feb84dae44568c0300/R/translate-sql-odbc.R#L11

R/db-odbc-teradata.R
edgararuiz: Ok, where should I place the function? I'm thinking `traslate-sql-helpers.R`

R/db-odbc-teradata.R
hadley: Ah got it. Might be better to have a more helpful error here? i.e. try `paste0()` instead

R/db-odbc-teradata.R
hadley: Yeah, sounds good.

R/db-odbc-teradata.R
edgararuiz: Good idea, I'd like to create a function for that, can I add `sql_not_supported()` a new `instead` argument? Or should I create a new function?

hadley: I think you can just create a new function inline

edgararuiz:
Ok, I moved `log` to its own function.  I noticed that `cot` was in the same situation so I moved that formula as well.  I also added, and updated,  the corresponding tests.

PTAL @hadley

R/translate-sql-helpers.r
```diff
@@ -170,3 +170,22 @@ sql_cast <- function(type) {
     build_sql("CAST(", x, " AS ", sql(type), ")")
   }
 }
+
+#' @rdname sql_variant
+#' @export
+sql_formula_log <- function(f, base = exp(1)) {
```
hadley: I think these can just be `sql_log` and `sql_cot`, and like `sql_cast()` etc should return functions

tests/testthat/test-translate-teradata.r
```diff
@@ -0,0 +1,150 @@
...
+  )
+})
+
+test_that("as.character() translate to VARCHAR(MAX) ", {
+  expect_equivalent(
+  translate_sql(as.character(field_name), con = simulate_teradata()),
+    sql("CAST(`field_name` AS VARCHAR(MAX))")
+  )
+})
+test_that("log10() translates to LOG ", {
```
hadley: Now that we have `sql_log()` used in multiple places, there's no need to test for each individual driver that uses it.

tests/testthat/test-translate-teradata.r
```diff
@@ -0,0 +1,150 @@
...
+    translate_sql(paste(field_name),
+                  window = FALSE,
+                  con = simulate_teradata()),
+    "PASTE\\(\\) is not supported in this SQL variant, try PASTE0\\(\\) instead"
+  )
+})
+
+# Teradata base_agg conversions -----------------------------------------
+
+test_that("sd() translates to STDEV ", {
```
hadley: Can you please bundle these all into one test, writing a function as needed to reduce duplication?

edgararuiz:
PTAL @hadley

PR #44, opened 2017-10-24
Typos in new-backend article

mungojam:


hadley:
Thanks!

mungojam:
I wasn't sure whether the building of the HTML was something I needed to do as part of the commit or not. I haven't done it.

PR #45, opened 2017-10-25
Fixes Oracle is.na issue, adds tests

edgararuiz:
Fixes https://github.com/tidyverse/dplyr/issues/3166

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dbplyr 1.1.0.9000
 
+* Fixed `is.na()` & `is.null()` translations for Oracle (@edgararuiz)
+
 * Added translation for Teradata (@edgararuiz)
 
 * dbplyr now uses tidyselect to power `rename()` and `select()` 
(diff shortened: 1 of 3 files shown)


edgararuiz:
PTAL @hadley

edgararuiz:
@hadley - Please hold off reviewing this PR.  I would like to check this fix against an open connection with Oralce, and using `!is.na()`.

edgararuiz:
@hadley  - This fix has the same deficiency as the one that's currently merged to master for MS SQL.  But I think it is still valuable for Oracle users.  Should I go ahead and update the code to use `sql_expr` instead of `build_sql`, or would you rather close it and find a solution that will work with normal and negated `is.na()`?

hadley:
Does `NOT(1)`/`NOT(0)` also fail in Oracle?

edgararuiz:
Yes, it looks that way:

```
> dbGetQuery(con, "SELECT NOT(0) from mtcars12")
  nanodbc/nanodbc.cpp:1344: HY000: ORA-00936: missing expression
 
> dbGetQuery(con, "SELECT NOT(1) from mtcars12")
  nanodbc/nanodbc.cpp:1344: HY000: ORA-00936: missing expression
 
> dbGetQuery(con, "SELECT 'hello'from mtcars12")
  'HELLO'
1   hello
```

edgararuiz:
I just thought about testing `1` w/o a NOT, and that worked:

```
> dbGetQuery(con, "SELECT 1 from mtcars12")
  1
1 1
```

edgararuiz:
That's correct

hadley:
Per slack discussion

PR #46, opened 2017-10-25
Fixes MS SQL issue of select after arrange

edgararuiz:
Fixes https://github.com/tidyverse/dplyr/issues/3062

It adds another layer to `sql-optimise` that checks to see if the current query is a SELECT or a FILTER and if the previous query is an ORDER BY and merges the two.  The exception is that if in the previous query history is a GROUP BY.  The exception is because implementing this optimization for grouped data seems to require a more complex solution that seems to be out of scope of the reported issue.

Here are the tests I ran:
[test_optimise.pdf](https://github.com/tidyverse/dbplyr/files/1416127/test_optimise.pdf)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dbplyr 1.1.0.9000
 
+* Improvements to `sql_optimise()` (#3062) (@edgararuiz)
+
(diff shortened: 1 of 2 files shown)


hadley:
See discussion in issue as to why this doesn't work 😞

PR #48, opened 2017-10-27
Fixes !is.na() for MS SQL

edgararuiz:
Fixes https://github.com/tidyverse/dplyr/issues/3172

- Created a custom `sql_prefix()` for MS SQL called `mssql_not_sql_prefix()` that checks the if the inside command contains code for a WHERE clause and flips the TRUE FALSE results of the IIF statement.

- Added a test for this change

### DIFF
--- a/R/db-odbc-mssql.R
+++ b/R/db-odbc-mssql.R
@@ -39,6 +39,10 @@
 `sql_translate_env.Microsoft SQL Server` <- function(con) {
   sql_variant(
     sql_translator(.parent = base_odbc_scalar,
+
+
+      `!`     = mssql_not_sql_prefix("not"),
+
       as.numeric    = sql_cast("NUMERIC"),
       as.double     = sql_cast("NUMERIC"),
       as.character  = sql_cast("VARCHAR(MAX)"),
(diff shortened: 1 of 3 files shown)


edgararuiz:
PTAL @hadley

hadley:
There are two problems with this approach:

* String manipulation means that it's very sensitive to the precise SQL that we generate elsewhere

* We could fix that, but it's an example of a general problem - we're going to have a problem any time `is.na()` is not the top-level expression.

Let me think if about this a bit more

edgararuiz:
What do you think about using `set_current_context()`  to add a new `negate` value we can use as a switch?  

Also, I'm sure i'm missing something, but I'm curious: wouldn't we have to make sure that the customization works only if `is.na()`  is a the top level?  Like for this: `!is.numeric(is.na(x))`, the negation only impacts `is.numeric()`, and should leave `is.na()` as is

hadley:
The problem is converting to the BIT data type - that seems to cause upstream problems. I wonder if a better solution is to just drop that and return `0/1`. It's not going to yield exactly the right data type in R, but it might be easier to compose elsewhere.

edgararuiz:
Right, another challenge is that the only way we can include a NULL evaluation in SELECT, ORDER BY. and GROUP BY clauses is by using IIF which cannot be prefixed with a NOT, I'm not sure how we would be able to negate `is.na()` without being able to their flip the TRUE/FALSE values, or by inserting a NOT in the expression `IIF(x IS NOT NULL... `

hadley:
Can you quickly confirm that `NOT(1)` doesn't work?

edgararuiz:
Yes, I can confirm: 

```
> dbGetQuery(con, "Select not(1)")
Error: <SQL> 'Select not(1)'
  nanodbc/nanodbc.cpp:1587: 42000: [Microsoft][ODBC SQL Server Driver][SQL Server]Incorrect syntax near the keyword 'not'. 
> dbGetQuery(con, "Select 'hello'")
       
1 hello
```

hadley:
Let's let this marinate for a while. This is going to be hard to fix.

hadley:
I wonder if we can make use of the bitwise not `~` - https://stackoverflow.com/questions/177762/boolean-not-in-t-sql-not-working-on-bit-datatype

hadley:
We also need to think about `x != 2` and `!(x == 2)` in both mutate and filter contexts. I suspect any function that yields a logical vector will need a similar transformation.  

And probably need to switch to bitwise `&` and `|` too.

PR #51, opened 2017-10-31
Fix negated is na

edgararuiz:
The new solution uses `~` for SELECT clauses and `NOT` for WHERE clauses.  It uses the `sql_currentcontext()` function to switch between the two.  

Here are the tests that worked: 

```
db_mtcars %>%
  mutate(x = !is.na(mpg))

db_mtcars %>%
  arrange(!is.na(mpg))

db_mtcars %>%
  group_by(!is.na(mpg))  %>%
  tally()

db_mtcars %>%
  filter(!is.na(mpg))

db_mtcars %>%
  filter(am != 0)

db_mtcars %>%
  filter(!am == 0)

db_mtcars %>%
  filter(!mpg < 30 )

db_mtcars %>%
  filter(!mpg < 30 | !cyl < 8 )

db_mtcars %>%
  filter(!mpg >= 30 )

db_mtcars %>%
  filter(!mpg == 21 )

db_mtcars %>%
  filter(mpg != 21 )
```

These didn't 
```
db_mtcars %>%
  mutate(x = am != 0) %>%
  show_query()

db_mtcars %>%
  mutate(x = (am == 0)) %>%
  show_query()
```

PR #53, opened 2017-11-07
Fixes ifelse after is.na eval change

edgararuiz:
Changes exclusive to MS SQL

- Adds custom a custom `mssql_is_na()` function that equates `cond` to TRUE so that CASE/WHEN works properly.  Because the options are binary, the False/No test is simplified to use ELSE
- Add custom to `if`, `ifelse` and `if_else` 
- Add tests 

Used same tests as https://github.com/tidyverse/dbplyr/pull/51 to confirm nothing originally fixed broke, and also tested these sucesfully:

```
db_mtcars %>%
  group_by(x = ifelse(mpg < 20, "good", "bad")) %>%
  select(mpg, x)

db_mtcars %>%
  group_by(x = if_else(mpg < 20, "good", "bad")) %>%
  select(mpg, x)

db_mtcars %>%
  mutate(x = ifelse(mpg <= 20 & am == 0, "good", "bad"))
```

### DIFF
--- a/R/db-odbc-mssql.R
+++ b/R/db-odbc-mssql.R
@@ -224,15 +224,15 @@ mssql_generic_infix <- function(if_select, if_filter) {
 
 mssql_sql_if <- function(cond, if_true, if_false = NULL) {
   build_sql(
-    "CASE WHEN ((", cond, ") =  'TRUE')", " THEN (", if_true, ")",
-    if (!is.null(if_false)){
-      build_sql(" ELSE (", if_false, ")")
-    } else {
-      build_sql(" ELSE ('')")
-    }
-      ,
-    " END"
-  )
+     "CASE",
+     " WHEN ((", cond, ") =  'TRUE')", " THEN (", if_true, ")",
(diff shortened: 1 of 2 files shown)


edgararuiz:
PTAL @hadley

R/db-odbc-mssql.R
```diff
@@ -218,4 +222,17 @@ mssql_generic_infix <- function(if_select, if_filter) {
   }
 }
 
+mssql_sql_if <- function(cond, if_true, if_false = NULL) {
+  build_sql(
+    "CASE WHEN ((", cond, ") =  'TRUE')", " THEN (", if_true, ")",
```
hadley: Can you please check to see if this has the same problem as fixed in tidyverse/dplyr#3157

Also can you have a go at using `sql_expr()` here? It should make the code quite a bit simpler.

edgararuiz:
PTAL @hadley

hadley:
Thanks!

PR #54, opened 2017-11-10
grepl() Postgresql mapping added

zozlak:


R/db-postgres.r
```diff
@@ -21,6 +21,16 @@ sql_translate_env.PostgreSQLConnection <- function(con) {
         digits <- as.integer(digits)
         sql_expr(round((!!x) %::% numeric, !!digits))
       },
+      grepl  = function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed = FALSE, useBytes = FALSE) {
+        if (any(c(perl, fixed, useBytes))) {
+          stop("perl, fixed and useBytes parameters are unsupported")
+        }
+        if (ignore.case) {
```
hadley: Would you mind including a link to the corresponding docs in a comment?

zozlak: Do you mean link to https://www.postgresql.org/docs/current/static/functions-matching.html#FUNCTIONS-POSIX-TABLE?

hadley: Yes, perfect. It's helpful for me when I re-read the code in the future because I forget all the details of the different SQL dialects

hadley:
Thanks!

PR #55, opened 2017-11-28
further distinguish the location of the execution

wibeasley:
Is this the best place to submit a modification?  Or should I do it on the dbplyr repo? https://github.com/rstudio/db.rstudio.com/blob/master/content/dplyr.Rmd


hadley:
Thanks!

PR #56, opened 2017-12-08
Change db_desc.PQConnection to db_desc.PqConnection in db-postgres.r

jonassundman:
Corrected db_desc.PQConnection to db_desc.PqConnection in db-postgres.r to make the RPostgres DBI implementation work correctly.

Also added db_desc.PostgreSQL, since there was already sql_translate and db_explain entries.


hadley:

hadley:
Actually, you don't need a news bullet, as this is just part of an existing one (add RPostgres support). Thanks for the PR!

PR #58, opened 2017-12-15
Adding integer64 support to escape() function

lbakerIsazi:
`integer64` vectors are converted to character and then passed to sql_vector.
Fixes tidyverse/dplyr#3230

### DIFF
--- a/R/sql-escape.r
+++ b/R/sql-escape.r
@@ -91,6 +91,13 @@ escape.integer <- function(x, parens = NA, collapse = ", ", con = NULL) {
 }
 
 #' @export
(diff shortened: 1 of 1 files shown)


hadley:
The approach in #60 looks a little simpler, so I'm going to go with that. 
Sorry I didn't merge your PR. I really appreciate the effort 😁

lbakerIsazi:
No worries. Thanks for having a look at this.

Do you have any idea when integer64 support in dplyr will filter through to the CRAN version?

hadley:
You mean in dbplyr? I'm hoping to release today, so should be on CRAN early next year.

PR #60, opened 2017-12-19
escaping for integer64

refik:
This is for enabling `integer64` support on queries. Useful for filtering on `bigint` primary keys. On its absence, this is the error:

```r
int64
#> integer64
#> [1] 223743292
dplyr::filter(postgresql_table, id == !!int64)
#> Error in UseMethod("escape") : 
#>  no applicable method for 'escape' applied to an object of class "integer64"
```

### DIFF
--- a/R/sql-escape.r
+++ b/R/sql-escape.r
@@ -91,6 +91,9 @@ escape.integer <- function(x, parens = NA, collapse = ", ", con = NULL) {
 }
 
 #' @export
+escape.integer64 <- escape.integer
+
+#' @export
 escape.NULL <- function(x, parens = NA, collapse = " ", con = NULL) {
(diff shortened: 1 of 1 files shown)


refik:
Or maybe custom class methods shouldn't be in this package?

hadley:
I think int64 is fine here, since we're going to use it throughout DBI. 

I'll add a couple of tests tomorrow before merging. In particular, need to confirm that int64 NA gets converted to NULL.

hadley:
Your approach didn't quite work, as revealed by a couple of tests: f716654dc1b810edac57d13150fa7a29047c5c93

Thanks for the idea and the initial implementation!

refik:
Thank you for showing the way :)

PR #61, opened 2017-12-22
Teradata not equal fix

edgararuiz:
- Fixes Teradata translation issue for `!=` operators, it now translates to `<>`. Reported by @weixing777.
- Adds a test for the change


hadley:

PR #63, opened 2017-12-29
fixed bug where copy_to failed for oracle-odbc because of analyze table

vitallish:
Modify translation of oracle analyze function so copy_to works correctly.

See issue here:
https://github.com/tidyverse/dplyr/issues/3265


hadley:
Can you please update the link in the comments to something that describes this behaviour?

hadley:
Looks like https://docs.oracle.com/cd/B19306_01/server.102/b14200/statements_4005.htm is good, and the "FOR TABLE" is not necessary?

hadley:
Did I myself since I really want to submit dbplyr to CRAN today. Thanks for the help!

PR #64, opened 2017-12-29
sample_n & sample_frac implementation

edgararuiz:


edgararuiz:
This PR focused on MS SQL because there was the need to create an new SQL clause to get it to work.  I'll be glad to submit separate PRs for the other backends, or extend this one, if the approach is acceptable.  

```
library(odbc)                                            
library(DBI)                                             
con <- dbConnect(odbc(), "datawarehouse")                
                                                         
airports <- tbl(con, in_schema("production", "airports"))
                                                         
tally(airports)                                          
#> # Database: Microsoft SQL Server
#> #   14.00.1000[dbo@DESKTOP-10DBTVP\SQLEXPRESS/datawarehouse]
#>       n
#>   <int>
#> 1  1458
                                                         
airports %>%                                             
group_by(dst) %>%                                        
sample_n(10) %>%                                         
show_query()                                             
#> <SQL>
#> SELECT *
#> FROM production.airports
#> TABLESAMPLE (10 ROWS)
                                                         
airports %>%                                             
sample_frac(0.1) %>%                                     
tally()                                                  
#> # Database: Microsoft SQL Server
#> #   14.00.1000[dbo@DESKTOP-10DBTVP\SQLEXPRESS/datawarehouse]
#>       n
#>   <int>
#> 1    90
                                                         
                                                         
dbDisconnect(con) 
```

edgararuiz:
ptal @hadley

R/translate-sql-clause.r
```diff
@@ -40,6 +40,25 @@ sql_clause_limit <- function(limit, con){
   }
 }
 
+sql_clause_sample <- function(sample, con){
+  if (!is.null(sample$size) && !identical(sample$size, Inf)) {
+    assert_that(is.numeric(sample$size), length(sample$size) == 1L, sample$size >= 0)
+    if(sample$type == "n"){
+      build_sql(
```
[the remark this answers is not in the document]
edgararuiz: I had to use `format()` to ensure that small fractions are not converted to scientific notation:

```
size <- 0.000000000001
sql_expr(TABLESAMPLE (!! size), con = simulate_dbi())
<SQL> TABLESAMPLE(1e-12)
```

I used `trunc()` to prevent passing `10.0` instead of `10`. 

Would you recommend a different way?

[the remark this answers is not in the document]
edgararuiz: This is my current solution, it works for both cases:

```
      size <- format(sample$size, scientific = FALSE)
      sql_expr(TABLESAMPLE (!! sql(size)), con = con)
```

edgararuiz:
Tested https://github.com/tidyverse/dbplyr/pull/64/commits/9712c72b35d30223b8df407558f2538e06e1234a against a PostgreSQL database:

```
library(DBI)
con <- dbConnect(odbc::odbc(), "Postgres Dev")
airports <- tbl(con, in_schema("datawarehouse", "airport")) 

airports %>%
  sample_frac(0.5) %>%
  tally()
#> # Database: postgres [rstudio_dev@localhost:/postgres]
#>   n              
#>   <S3: integer64>
#> 1 91

dbDisconnect(con)
```

R/db-postgres.r
```diff
@@ -1,3 +1,43 @@
...
+                  "limit", "sample")
+
+  out$select    <- sql_clause_select(select, con, distinct)
+  out$from      <- sql_clause_from(from, con)
+  out$where     <- sql_clause_where(where, con)
+  out$group_by  <- sql_clause_group_by(group_by, con)
+  out$having    <- sql_clause_having(having, con)
+  out$order_by  <- sql_clause_order_by(order_by, con)
+  out$limit     <- sql_clause_limit(limit, con)
+  size <- sample$size
```
hadley: Rather than duplicating this big function, I think it would be better to have 

```R
sql_clause_sample <- fuction(sample, con) {
  UseMethod("sql_clause_sample", con)
}

sql_clause_sample.SQLiteConnection <- function(sample, con) {
   ...
}

```

SQLite would still need special treatment, but this would reduce the duplication significantly

hadley: Does that make sense? Let me know if you need more details.

edgararuiz: Yes, it does. Thank you.

edgararuiz:
We will need to take into account `sparklyr`.  This version of `dbplyr` prevents `sample_...` from running in `sparklyr`, even though it is a supported method.  I was thinking about removing `sql_clause_sample.default()` and then explicitly add methods for each respective back end  that is translated in `dbplyr` but is not supported.

hadley:
Why does this break sparklyr?

edgararuiz:
It looks like `sparklyr` does not have a custom `sql_select()` , so it's affected by the changes I made to `sql_select.DBIConnection`

hadley:
Ah, got it, so we'll need to think about backward compatibility.

hadley:
Now tracking in #394

PR #66, opened 2018-02-12
Changed VARCHAR to VARCHAR2

washcycle:
Per

> VARCHAR Datatype
> Do not use the VARCHAR datatype. Use the VARCHAR2 datatype instead. Although the VARCHAR datatype is currently synonymous with VARCHAR2, the VARCHAR datatype is scheduled to be redefined as a separate datatype used for variable-length character strings compared with different comparison semantics.

https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements001.htm


hadley:
Can you please update the tests too?

washcycle:
Done. Let me know if there is anything else. Or if this wasn't want you meant.

hadley:
Looks good, thanks! To finish it off, can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Thanks!

PR #68, opened 2018-02-20
Switches to substring and adds test

edgararuiz:
Fixes https://github.com/tidyverse/dplyr/issues/3339

- It overrides `substr` in the PostgreSQL translation. It now uses `substring` which will work both in PostgreSQL and Redshift


edgararuiz:
PTAL @hadley

hadley:
Thanks @edgararuiz!

PR #69, opened 2018-02-20
fix sql translation of %in% with empty vectors

mgirlich:
Before the PR we have the following translation

```
dbplyr::translate_sql(col %in% !!c())
<SQL> "col" IN ()
```

which isn't valid SQL. With this PR this is simply translated to FALSE

```
dbplyr::translate_sql(col %in% !!c())
<SQL> FALSE
```

Closes tidyverse/dplyr#3375.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # dbplyr 1.2.1.9000
 
-* sql translation of `%in%` now works with empty vectors (@mgirlich)
+* the statement `x %in% y` is now translated to `FALSE`. (@mgirlich)
 
 # dbplyr 1.2.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # dbplyr 1.2.1.9000
 
-* the statement `x %in% y` is now translated to `FALSE`. (@mgirlich)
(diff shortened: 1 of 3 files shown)


tests/testthat/test-translate.r
```diff
@@ -65,6 +65,11 @@ test_that("%in% translation parenthesises when needed", {
   expect_equal(translate_sql(x %in% y), sql('"x" IN "y"'))
 })
 
+test_that("%in% with empty vector", {
+  expect_equal(translate_sql(x %in% !!c()), sql('FALSE'))
```
hadley: I think you can eliminate this test

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # dbplyr 1.2.1.9000
 
+* sql translation of `%in%` now works with empty vectors (@mgirlich)
```
hadley: Could you please make this slightly more descriptive by saying what it does?

krlmlr:
@hadley: Is this good to merge? I added a "Closes ..." entry in the issue description.

hadley:
@krlmlr I will start working on dbplyr seriously in the next week or so. There is no need to ping me on individual issues.

PR #70, opened 2018-02-21
Coerces substring vals to integer

edgararuiz:
- Fix to address an issue created by the previous Redshift fix: https://github.com/tidyverse/dplyr/issues/3339#issuecomment-367470352

- Also updates the test

- Tested my branch against an actual PostgreSQL connection to confirm that it works


edgararuiz:
PTAL @hadley

hadley:

PR #71, opened 2018-02-23
Feature/add pool

AkhilNairAmey:
Related to rstudio/pool#61. Paired with pull request rstudio/pool#62.

 - Adds pool as a dependency.
 - All `dplyr` compatibility tests from `pool` added
 - Test harness passes

If this does not break existing code, and if it respects the design philosophy of these packages, I think this is useful as the overhead in unnecessarily pulling in `dplyr` when using `pool` is far greater than the overhead of pulling in `pool` unnecessarily when using `dbplyr`.


AkhilNairAmey:
PS, this will fail with a circular dependency warning until `dbplyr` is removed from `pool` as a dependency.

edgararuiz:
Hi @AkhilNairAmey , are you using `pool` outside `shiny` apps? In other words, are you using `pool` connections for everyday analysis?

AkhilNairAmey:
Just shiny apps :)

hadley:
See discussion in original issue.

PR #72, opened 2018-02-25
Adds support for spread

edgararuiz:
Adds the following:

- Initial implementation of `tidyr::spread()`, for `tbl_sql` objects

- A test to confirm that it conforms with the original `spread()` function

- NEWS item


edgararuiz:
ptal @hadley

hadley:
Hmmmm, I'm not sure I want to start on the tidyr verbs. Is there somewhere more experimental that this could live in the interim?

edgararuiz:
Should we create an interim `tidydb` package?

R/spread.R
```diff
@@ -0,0 +1,27 @@
...
+  conds <- keys %>%
+    purrr::map(~expr(ifelse((!! key) == (!! .x), !! value, (!! fill))))
+
+  # Sets names
+  nf <- keys %>%
+    purrr::map(~expr(!! sym(.x)))
+
+  # Adds a mutate per each unique value
+  for(vals in 1:length(keys)){
+    data <- mutate(data, !! expr_text(nf[[vals]]) := !! conds[[vals]])
```
jarodmeng: Conceptually I think the operation should be a summarise() instead of mutate(). You can see your result differs from tidyr::spread() if you use the following test data.

```r
df <- tibble(
  h = c("a", "b", "c", "d", "a", "b", "c", "d"),
  x = c("A", "A", "B", "B", "B", "B", "A", "A"),
  y = c("X", "Y", "X", "Y", "X", "Y", "X", "Y"),
  z = c( 1,   2,   3,   4,   5,   6,   7,   8 )
)
```

hadley:
Yeah, I think that's a good idea. Maybe just back it your personal repo for now?

edgararuiz:
Sounds good

PR #75, opened 2018-03-13
Simplify join_vars()

krlmlr:
I'd like to make the definition of the results of a join consistent between dplyr and dbplyr, in the context of https://github.com/tidyverse/dplyr/issues/3307. I don't understand why we need to pass the join type, tests seem to pass locally without it.

### DIFF
--- a/R/sql-query.R
+++ b/R/sql-query.R
@@ -71,7 +71,7 @@ join_query <- function(x, y, vars, type = "inner", by = NULL, suffix = c(".x", "
 
 
 # Returns NULL if variables don't need to be renamed
-join_vars <- function(x_names, y_names, type, by, suffix = c(".x", ".y")) {
+join_vars <- function(x_names, y_names, by, suffix = c(".x", ".y")) {
   # Remove join keys from y
   y_names <- setdiff(y_names, by$y)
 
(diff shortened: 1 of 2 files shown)


ianmcook:
@krlmlr without knowing the join type, we can't know whether it's necessary to use a SQL `COALESCE()` to return the correct result. With this change you've made, the resulting SQL statement will call `COALESCE()` every time, including when it's not required, which has an efficiency cost.

Here's the code that determines whether or not to use `COALESCE()` based on whether the `x` and `y` values in the list returned by `join_vars()` are `NA`:

https://github.com/tidyverse/dbplyr/blob/331646383d7e1ae574b161b6a16a1d1d0582af54/R/sql-generic.R#L88-L101

krlmlr:
Thanks, makes sense. Will resubmit (and add a test ;-) ).

ianmcook:
Thanks!

ianmcook:
There are some existing tests for consistency of join results for SQL vs. data frames here:
https://github.com/tidyverse/dbplyr/blob/master/tests/testthat/test-joins-consistent.R

krlmlr:
These tests haven't fired for this PR, at least not on Travis.

ianmcook:
This PR wouldn't have caused the tests to fail; it would just have caused the presence of unnecessary `COALESCE()` calls in the SQL they generate. But the actual returned results would have been the same, because `COALESCE(x, y)` and simply `x` both return `x` when `y` is `NA`.

To create a couple of tests that confirm that `COALESCE()` is not being called unnecessarily, we would probably want to call `sql_render()` after a left join and and inner join, and use `grepl()` to confirm that there's no `coalesce` in the resulting SQL strings.

PR #77, opened 2018-03-15
Typo in vignette doc

tmastny:
focusses -> focuses

### DIFF
--- a/vignettes/dbplyr.Rmd
+++ b/vignettes/dbplyr.Rmd
@@ -108,7 +108,7 @@ The main difference is that you can see that it's a remote source in a SQLite da
 
 ## Generating queries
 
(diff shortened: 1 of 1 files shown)


tmastny:
Actually, maybe this isn't a typo, but a regional variant:
 https://english.stackexchange.com/questions/4791/focussed-or-focused-rules-for-doubling-the-last-consonant-when-adding-ed

This website says the single `s` is highly preferred: http://www.future-perfect.co.uk/grammar-tip/is-it-focussed-or-focused/

batpigandme:
I've made this change in the past as well, but it's how @hadley spells it, so it's how dplyr/dbplyr spells it! Thanks for making the fix, though.

PR #78, opened 2018-03-24
Set out.width for figure in dbplyr vignette.

batpigandme:
* Fixes https://github.com/tidyverse/dplyr/issues/3441.
* Set `echo = FALSE` and `out.width` params for windows figure in SQL-translation vignette.


hadley:
I think the dpi should also be higher? The size of the text in the image should be the same as on the page.

batpigandme:
I increased the font size and the dpi, but I can't get the article to build without it somehow reverting back to the old figure.

hadley:
You shouldn't need to change the figure in any way; just adjust the dpi.  (It doesn't need to be exactly the same, just approximately)

PR #79, opened 2018-03-26
Rebuild pkgdown site.

batpigandme:
Update pkgdown site.


batpigandme:
Closing until tidytemplate is caught up with pkgdown.

PR #80, opened 2018-04-06
Adds support for explain() for Oracle connections

edgararuiz:
Fixes https://github.com/tidyverse/dplyr/issues/3471


edgararuiz:
I tested this change against our Oracle database. Results:

```
> tbl(con, "mtcars") %>%
+   explain()
<SQL>
SELECT *
FROM "mtcars"


<PLAN>
                                                             PLAN_TABLE_OUTPUT
1                                                  Plan hash value: 1092787947
2                                                                             
3 ----------------------------------------------------------------------------
4 | Id  | Operation         | Name   | Rows  | Bytes | Cost (%CPU)| Time     |
5 ----------------------------------------------------------------------------
6 |   0 | SELECT STATEMENT  |        |    32 |  2976 |     3   (0)| 00:00:01 |
7 |   1 |  TABLE ACCESS FULL| mtcars |    32 |  2976 |     3   (0)| 00:00:01 |
8 ----------------------------------------------------------------------------
> 
> tbl(con, "mtcars") %>%
+   filter(am == 1) %>%
+   select(gear, wt) %>%
+   explain()
<SQL>
SELECT "gear", "wt"
FROM ("mtcars") 
WHERE ("am" = 1.0)


<PLAN>
                                                              PLAN_TABLE_OUTPUT
1                                                   Plan hash value: 1092787947
2                                                                              
3  ----------------------------------------------------------------------------
4  | Id  | Operation         | Name   | Rows  | Bytes | Cost (%CPU)| Time     |
5  ----------------------------------------------------------------------------
6  |   0 | SELECT STATEMENT  |        |    16 |   384 |     3   (0)| 00:00:01 |
7  |*  1 |  TABLE ACCESS FULL| mtcars |    16 |   384 |     3   (0)| 00:00:01 |
8  ----------------------------------------------------------------------------
9                                                                              
10                          Predicate Information (identified by operation id):
11                          ---------------------------------------------------
12                                                                             
13                                                   1 - filter("am"=1.0E+000D)
```

edgararuiz:
PTAL @hadley  - It passed CI but  failed code coverage. I'm not able to add a test for this change.

brianstamper:
Thanks, Edgar! Might this also need a wrapper in `R/db-roracle.R`?

edgararuiz:
You're right! Great catch! I'll update the PR

R/db-odbc-oracle.R
```diff
@@ -45,6 +45,15 @@ sql_translate_env.Oracle <- function(con) {
   )
 }
 
+#' @export
+db_explain.Oracle <- function(con, sql, ...) {
+  exsql <- build_sql("EXPLAIN PLAN FOR ", sql, con = con)
+  invisible(dbGetQuery(con, exsql))
```
hadley: What's happening here? Why do you ignore the output of this query?

Can we use `DBI::dbExecute()` instead?

edgararuiz: Yes, you're right, that would be more straightforward

PR #81, opened 2018-04-06
Fixes instr order for default translation

edgararuiz:
Fixes https://github.com/tidyverse/dplyr/issues/3397

Notes:

- Tested it against an Oracle and SQLite connections
- Re-tested the MS SQL custom translation and found that it needed no changes

### DIFF
--- a/R/translate-sql-base.r
+++ b/R/translate-sql-base.r
@@ -167,7 +167,7 @@ base_scalar <- sql_translator(
                       )},
   str_detect      = function(string, pattern){
                       build_sql(
-                        "INSTR(", pattern, ", ", string, ") > 0"
+                        "INSTR(", string, ", ", pattern, ") > 0"
                       )},
   str_trim        = function(string, side = "both"){
                       build_sql(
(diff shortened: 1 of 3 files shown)


edgararuiz:
ptal @hadley

R/translate-sql-base.r
```diff
@@ -167,7 +167,7 @@ base_scalar <- sql_translator(
                       )},
   str_detect      = function(string, pattern){
```
hadley: Can you please add a comment here pointing to the documentation fo INSTR?

edgararuiz: I have links to string functions for the different databases at the beginning of the section were we added the `stringr` functions: https://github.com/tidyverse/dbplyr/blob/81419cab7d831e5a9d7e4279a4f893fa5750e8a1/R/translate-sql-base.r#L154-L160 
Does this suffice?

hadley: Oh yeah, that's sufficient.

hadley:
Thanks!

leungi:
Thanks @edgararuiz and @hadley!

PR #83, opened 2018-05-02
Test tally with auto-naming

krlmlr:
Requires https://github.com/tidyverse/dplyr/pull/3525, CC @romainfrancois.


hadley:
This fails on purpose, right?

krlmlr:
Looks like a test where the current implementation is inconsistent with _dplyr_:

``` r
library(tidyverse)
tibble(n = 1:3) %>% tally()
#> Using `n` as weighting variable
#> # A tibble: 1 x 1
#>      nn
#>   <int>
#> 1     6
dbplyr::memdb_frame(n = 1:3) %>% tally()
#> # Database: sqlite 3.22.0 [:memory:]
#>      nn
#>   <int>
#> 1     3
```

<sup>Created on 2019-01-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

hadley:
Ah, it only works with dev dplyr

hadley:
In the original PR, you thought this should belong in dbplyr, but given that the change is to code in dplyr (and it's clearly a dbplyr bug), I feel like it belongs in dplyr.

krlmlr:
Works for me in the dev version too. Are you comfortable depending on dev dplyr here?

hadley:
I think the test has to be conditional on the dplyr version because dbplyr needs to be released before dplyr.

hadley:
I'll finish this off

hadley:
See https://github.com/tidyverse/dplyr/issues/4076

PR #131, opened 2018-07-19
Add stringr

colearendt:
Add / improve some `stringr` methods.

image

```yaml
- str_length:
    mutate: str_length(fld_character)
    filter: str_length(fld_character) > 1
    summarize: n_distinct(str_length(fld_character))
    group_by: str_length(fld_character)
    arrange: str_length(fld_character)
- str_to_upper:
    mutate: str_to_upper(fld_character)
    filter: str_to_upper(fld_character) %in% c("A","B")
    summarize: n_distinct(str_to_upper(fld_character))
    group_by: str_to_upper(fld_character)
    arrange: str_to_upper(fld_character)
- str_to_lower:
    mutate: str_to_lower(fld_character)
    filter: str_to_lower(fld_character) %in% c("a","b")
    summarize: n_distinct(str_to_lower(fld_character))
    group_by: str_to_lower(fld_character)
    arrange: str_to_lower(fld_character)
- str_detect:
    mutate: str_detect(fld_character, '[A-Z]')
    filter: str_detect(fld_character, '[A-Z]')
    summarize: n_distinct(str_detect(fld_character, '[A-Z]'))
    group_by: str_detect(fld_character, '[A-Z]')
    arrange: str_detect(fld_character, '[A-Z]')
- str_to_title:
    mutate: str_to_title(fld_character)
    filter: str_detect(str_to_title(fld_character), '[A-J]')
    summarize: n_distinct(str_to_title(fld_character))
    group_by: str_to_title(fld_character)
    arrange: str_to_title(fld_character)
- str_replace_all:
    mutate: str_replace_all(fld_character, '[A-J]', '*')
    filter: str_detect(str_replace_all(fld_character, '[A-J]', '*'), '*')
    summarize: n_distinct(str_replace_all(fld_character, '[A-J]', '*'))
    group_by: str_replace_all(fld_character, '[A-J]', '*')
    arrange: str_replace_all(fld_character, '[A-J]', '*')
- str_sub:
    mutate: str_sub(fld_character, 4,8)
    filter: str_detect(str_sub(fld_character, 4,8), '[A-J]')
    summarize: n_distinct(str_sub(fld_character, 4,8))
    group_by: str_sub(fld_character, 4,8)
    arrange: str_sub(fld_character, 4,8)
```


colearendt:
New `dbtest` YAML and results. Some noteworthy issues this has surfaced:

- oracle translation issues: https://github.com/tidyverse/dbplyr/issues/172
- Postgres ODBC serializes logical to character... have not filed issue yet

image


```yaml
- str_length:
    mutate: str_length(fld_character)
    filter: str_length(fld_character) > 1
    summarize: n_distinct(str_length(fld_character))
    group_by: str_length(fld_character)
    arrange: str_length(fld_character)
- str_to_upper:
    mutate: str_to_upper(fld_character)
    filter: str_to_upper(fld_character) %in% c("A","B")
    summarize: n_distinct(str_to_upper(fld_character))
    group_by: str_to_upper(fld_character)
    arrange: str_to_upper(fld_character)
- str_to_lower:
    mutate: str_to_lower(fld_character)
    filter: str_to_lower(fld_character) %in% c("a","b")
    summarize: n_distinct(str_to_lower(fld_character))
    group_by: str_to_lower(fld_character)
    arrange: str_to_lower(fld_character)
- str_detect:
    mutate: str_detect(fld_character, '[A-Z]')
    filter: str_detect(fld_character, '[A-Z]')
    summarize: n_distinct(str_detect(fld_character, '[A-Z]'))
    group_by: str_detect(fld_character, '[A-Z]')
    arrange: str_detect(fld_character, '[A-Z]')
- str_to_title:
    mutate: str_to_title(fld_character)
    filter: str_detect(str_to_title(fld_character), '[A-J]')
    summarize: n_distinct(str_to_title(fld_character))
    group_by: str_to_title(fld_character)
    arrange: str_to_title(fld_character)
- str_replace_all:
    mutate: str_replace_all(fld_character, '[A-J]', '*')
    filter: str_detect(str_replace_all(fld_character, '[A-J]', '*'), '*')
    summarize: n_distinct(str_replace_all(fld_character, '[A-J]', '*'))
    group_by: str_replace_all(fld_character, '[A-J]', '*')
    arrange: str_replace_all(fld_character, '[A-J]', '*')
- str_replace:
    mutate: str_replace(fld_character, '[A-J]', '*')
    filter: str_detect(str_replace(fld_character, '[A-J]', '*'), '*')
    summarize: n_distinct(str_replace(fld_character, '[A-J]', '*'))
    group_by: str_replace(fld_character, '[A-J]', '*')
    arrange: str_replace(fld_character, '[A-J]', '*')
- str_sub:
    mutate: str_sub(fld_character, 4,8)
    filter: str_detect(str_sub(fld_character, 4,8), '[A-J]')
    summarize: n_distinct(str_sub(fld_character, 4,8))
    group_by: str_sub(fld_character, 4,8)
    arrange: str_sub(fld_character, 4,8)
```

skip YAML:
```yaml
- text: str_to_title not supported by MSSQL
  db: mssql
  file: character-stringr
  context: str_to_title

```

R/translate-sql-base.r
```diff
@@ -158,9 +158,12 @@ base_scalar <- sql_translator(
   #   PostgreSQL: https://www.postgresql.org/docs/9.1/static/functions-string.html
   #   MS SQL: https://docs.microsoft.com/en-us/sql/t-sql/functions/string-functions-transact-sql
   #   Oracle: https://docs.oracle.com/middleware/bidv1221/desktop/BIDVD/GUID-BBA975C7-B2C5-4C94-A007-28775680F6A5.htm#BILUG685
-  str_length      = sql_prefix("LENGTH"),
-  str_to_upper    = sql_prefix("UPPER"),
-  str_to_lower    = sql_prefix("LOWER"),
+  #   Alt Oracle: https://docs.oracle.com/database/121/SQLRF/functions.htm#SQLRF006
+  #   Redshift: https://docs.aws.amazon.com/redshift/latest/dg/String_functions_header.html
+  str_length      = sql_prefix("LENGTH",1),
```
hadley: Can you please add a space after the comma?

R/translate-sql-base.r
```diff
@@ -175,7 +178,13 @@ base_scalar <- sql_translator(
                         sql(ifelse(side == "both" | side == "right", "RTRIM(", "(")),
                         string
                         ,"))"
-                      )}
+                      )},
+  str_sub = function(string, start = 1L, end = -1L) {
+    start <- as.integer(start)
+    length <- pmax(as.integer(end) - start + 1L, 0L)
+
+    build_sql(sql("SUBSTR"), list(string, start, length))
```
hadley: I think it's better to use `sql_expr()` for new code

tests/testthat/test-translate.r
```diff
@@ -146,6 +153,21 @@ test_that("str_trim() translates correctly ", {
...
+    translate_sql(
+      str_sub(field_name, 2L, 4L)
+    ),
+    sql("SUBSTR(\"field_name\", 2, 3)")
+  )
+  expect_equivalent(
+    translate_sql(
+      str_sub(field_name)
+    ),
+    sql("SUBSTR(\"field_name\", 1, 0)")
```
hadley: Shouldn't this just be `SUBSTR(\"field_name\", 1)` ?

R/translate-sql-base.r
hadley: I think this also needs a comment briefly explaining the parameterisation of `SUBSTR()`

colearendt:
With all of the changes to the file structure, where should tests of `backend-.R` be placed? It does not look like `test-translate.R` got preserved? Should we only be testing specific backends?

I am doing a bit of research into `SUBSTR`. The docs for Oracle / Postgres use `SUBSTR(string, start, length)`, where MSSQL uses `SUBSTRING(string, start, length)`. Do you have a standard resource that constitutes "what is base SQL" and what is backend-specific?

hadley:
The tests for code in `foo.R` should live in `tests/testthat/test-foo.R`, which will mostly be `test-backend-.R` (which now exists).

Technically the SQL 93 reference is the best bet, but it costs a bunch to buy it and I don't have a copy. Generally Postgres is closest to the spec, so I'd go with it unless you see a lot of backends taking a different approach.

hadley:
Also let me know if it be helpful to have a quick call.

R/backend-.R
```diff
@@ -186,10 +186,11 @@ base_scalar <- sql_translator(
...
   #   MS SQL: https://docs.microsoft.com/en-us/sql/t-sql/functions/string-functions-transact-sql
-  #   Oracle: https://docs.oracle.com/middleware/bidv1221/desktop/BIDVD/GUID-BBA975C7-B2C5-4C94-A007-28775680F6A5.htm#BILUG685
-  str_length      = sql_prefix("LENGTH"),
-  str_to_upper    = sql_prefix("UPPER"),
-  str_to_lower    = sql_prefix("LOWER"),
+  #   Oracle: https://docs.oracle.com/database/121/SQLRF/functions002.htm#SQLRF51180
+  str_length      = sql_prefix("LENGTH", 1),
+  str_to_upper    = sql_prefix("UPPER", 1),
+  str_to_lower    = sql_prefix("LOWER", 1),
+  str_to_title    = sql_prefix("INITCAP", 1),
```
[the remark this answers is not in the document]
colearendt: I think it's wider than just Oracle. Postgres has it, as does Hive (random other database I tested).

https://w3resource.com/PostgreSQL/initcap-function.php
https://www.qubole.com/resources/hive-function-cheat-sheet/

R/backend-postgres.R
```diff
@@ -54,6 +54,9 @@ sql_translate_env.PostgreSQLConnection <- function(con) {
       },
       str_detect  = function(string, pattern) {
         sql_expr(strpos(!!string, !!pattern) > 0L)
+      },
```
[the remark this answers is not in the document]
colearendt: Updated this one, but unfortunately the only real option (`regexp_matches`) is a set-valued function

tests/testthat/test-translate.r
colearendt: I think that you're right - I am not handling this edge case correctly. I'm struggling with how to reason through dropping a parameter in the SQL translation.

R/backend-mssql.R
```diff
@@ -97,11 +97,23 @@
       # stringr functions
 
       str_length      = sql_prefix("LEN"),
+      # mssql does not support INITCAP, toTitle, toPropCase, etc.
+      # related SO post:
+      # https://stackoverflow.com/questions/230138/sql-server-make-all-upper-case-to-proper-case-title-case
+      str_to_title    = sql_not_supported("str_to_title()"),
+      # potentially helpful in the future
+      # https://stackoverflow.com/questions/38911588/replace-first-occurrence-of-substring-in-a-string-in-sql
+      str_replace     = sql_not_supported("str_replace()"),
```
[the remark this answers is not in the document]
colearendt: Yeah, this is unfortunate 😭 In my searching, MS SQL does not seem to support regular expressions. Do we need to throw an error due to inconsistencies? Would it be sufficient to warn that there is unexpected behavior, but we got as close as we can?

https://study.com/academy/lesson/regular-expressions-in-sql-server-databases-implementation-use.html

> Note that SQL Server does not include a full-blown REGEX function, which MySQL includes. True regex operations in SQL Server require additional code and plug-ins.

colearendt:
A few outstanding questions from my point of view:

- I pulled out the `substr` function into a helper... but immediately feel like I am doing something wrong. Builds are failing (Collate something or other), it is not clear to me whether this helper is sufficient, and I'm not 100% clear on how to make `sql_expr` return `SUBSTR` in some cases and `SUBSTRING` in others.
- related to above, `str_sub("string")` needs to be handled better by dropping the third parameter (end). Not sure how to accomplish that.
- Should we warn or error for `str_detect` / `str_locate` on MSSQL, since it does not seem to support regexes out of the box? 🤕 
- anything to change on my `regexp_matches` implementation?

R/translate-sql-string.R
```diff
@@ -0,0 +1,9 @@
+# R prefers to specify start / stop or start / end
+# databases usually specify start / length
+# https://www.postgresql.org/docs/current/functions-string.html
+sql_substr <- function(x, start, stop) {
```
hadley: To be consistent with the other helper functions, this should be a function factory that takes the name of the SQL function as an input. Look at `sql_paste()` (but `sql_substr()` should be a bit simpler), and then let me know if you need more help.

R/backend-postgres.R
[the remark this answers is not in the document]
hadley: I think you could use `~` instead: https://www.postgresql.org/docs/9.1/functions-matching.html#FUNCTIONS-POSIX-REGEXP

R/translate-sql-string.R
```diff
@@ -0,0 +1,9 @@
+# R prefers to specify start / stop or start / end
```
hadley: To fix the collate problem, you need to add a line to the top of `backends-.R`. This is a weird design that I now regret, but I can't figure out how to figure it in a backward compatible way.

R/backend-mssql.R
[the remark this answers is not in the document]
hadley: Since it's only the standard "LIKE" support, I think we should drop support with an error.

R/backend-postgres.R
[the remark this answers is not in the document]
colearendt: That's perfect, thanks!

colearendt:
Still working on the `substr` helper, but it looks to me like `str_locate` will be very hard to _actually_ translate into the database since it returns `start` and `end`. Would there be any reason within `stringr` to have a function like `stringr::str_locate_start` that only returns the `start` of the regexp match? I find myself defining such a function for friendliness within `mutate` and other `dplyr` operations anyways. It seems relevant just because that is what the databases do (although it seems postgres, for instance, does not support regexp):

https://dev.mysql.com/doc/refman/8.0/en/regexp.html#function_regexp-instr
https://www.postgresql.org/docs/11/functions-string.html
https://docs.oracle.com/cd/B19306_01/B14251_01/adfns_regexp.htm

R/backend-mysql.R
```diff
@@ -21,7 +21,14 @@ sql_translate_env.MySQLConnection <- function(con) {
       },
       as.character = sql_cast("CHAR"),
       paste = sql_paste(" "),
-      paste0 = sql_paste("")
+      paste0 = sql_paste(""),
+      # https://dev.mysql.com/doc/refman/8.0/en/regexp.html
+      str_detect = function(string, pattern) {
```
hadley: Can you use `sql_infix()` here?

R/backend-postgres.R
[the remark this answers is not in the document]
hadley: And I think you can simplify even further with `sql_infix()`

colearendt:
I agree. Would a helper that does this be useful for future package support? It prints out a list of all functions in the package as "not supported." Definitely only useful in an ad-hoc / interactive sense, but useful as a baseline for adding package support that can then be walked back either generally or per-database.
```
ls("package:stringr") %>% 
  map(
    ~ if (class(eval(as.name(.x))) == "function") 
      glue::glue("{.x} = {capture.output(args(.x))[1]} sql_not_supported(\"{.x}\"),")
    ) %>% 
  as.character() %>% 
  .[! . == "NULL"] %>%
  cat(sep = "\n")
```

Example output
```
...
str_c = function (..., sep = "", collapse = NULL)  sql_not_supported("str_c"),
str_conv = function (string, encoding)  sql_not_supported("str_conv"),
str_count = function (string, pattern = "")  sql_not_supported("str_count"),
str_detect = function (string, pattern)  sql_not_supported("str_detect"),
...
```

colearendt:
Ok I have a decent pass of the helper functions, although I keep running into this error for `sql_not_supported`: `"no applicable method for 'escape' applied to an object of class \"function\""`

R/backend-.R
```diff
@@ -197,31 +193,62 @@ base_scalar <- sql_translator(
...
+  str_trim = function(string, side = c("both", "left", "right")) {
+    side <- match.arg(side)
+    switch(side,
+      left = sql_expr(LTRIM(!!string)),
+      right = sql_expr(RTRIM(!!string)),
+      both = sql_expr(LTRIM(RTRIM(!!string))),
+    )
+   },
+  str_sub = sql_str_sub("SUBSTR"),
+  str_c = function (..., sep = "", collapse = NULL) sql_not_supported("str_c"),
```
hadley: I think the not supported functions can just use `...`, since it doesn't actually matter what the arguments are, and it makes the lines a little shorter.

R/translate-sql-string.R
```diff
@@ -0,0 +1,33 @@
...
+  }
+}
+
+# if no parameters are provided, str_sub returns the whole string
+# if just start is provided, str_sub returns the rest of the string (after start)
+#' @export
+#' @rdname sql_variant
+sql_str_sub <- function(f = "SUBSTR", full_length = "drop", compute_method = "LENGTH") {
+  function(string, start = 1L, end = -1L) {
+    start <- as.integer(start)
```
hadley: I think I'll implement this a little differently, so can you please just make sure that the unit tests cover the positive and negative cases that you are thinking about? (The tests would live in `tests/testthat/test-translate-sql-string.R`)

R/backend-.R
colearendt: Was there a breaking change in `sql_not_supported`? 

I keep getting: `no applicable method for 'escape' applied to an object of class "function"` instead of the "not supported" message

colearendt: I take that back... this may be an implementation problem for `str_detect`...

colearendt:
Added list of not-yet-provided lubridate functions and cleaned things up. The code that generated the list:
```
ls("package:lubridate") %>% 
  purrr::map(
    ~ if (class(eval(as.name(.x))) == "function") 
      glue::glue("{.x} = sql_not_supported(\"{.x}()\"),")
  ) %>% 
  as.character() %>% 
  .[! . == "NULL"] %>%
  cat(sep = "\n")
```

colearendt:
Very strange note. If I used the `"DummyDBIConnection"` class in testing, I would get duplicated results... i.e. instead of `SUBSTRING('string')`, I would get `SUBSTRING('string','string','string','string','string')`. There is some part of the tests here that seems to be stateful. It was resolved by creating a unique class name (i.e. DummyDBIConnectionSQLSubstr).

hadley:
@colearendt can you please take a look at my changes, and check that the tests look correct to you? I didn't see the repeated string problem, but I made other changes as well.

colearendt:
Ahh I see - you always compute the length of the string in `str_sub` (when needing "infinite length") instead of trying to factor in whether or not the database engine supports dropping the parameter. That's an elegant cleanup! I like it. LGTM!

PR #140, opened 2018-08-10
Adds 4 tests

edgararuiz:
Added as many tests I could during my flight, it'll hopefully helps us with our coverage levels.

### DIFF
--- a/tests/testthat/test-sql-query.R
+++ b/tests/testthat/test-sql-query.R
@@ -4,3 +4,14 @@ test_that("add_suffixes works if no suffix requested", {
   expect_equal(add_suffixes(c("x", "x"), "y", ""), c("x", "x"))
   expect_equal(add_suffixes(c("x", "y"), "y", ""), c("x", "y"))
 })
+
+test_that("select_query() print method output is as expected", {
+  mf <- select_query(lazy_frame(x = 1, src = simulate_dbi()))
+  expect_equal(
+    paste0(
+      capture.output(print.select_query(mf)),
+      collapse = " "
+    ),
+    "<SQL SELECT> From:     df Select:   *"
+  )
+})
--- a/tests/testthat/test-sql-query.R
+++ b/tests/testthat/test-sql-query.R
@@ -7,11 +7,5 @@ test_that("add_suffixes works if no suffix requested", {
 
 test_that("select_query() print method output is as expected", {
   mf <- select_query(lazy_frame(x = 1, src = simulate_dbi()))
-  expect_equal(
-    paste0(
-      capture.output(print.select_query(mf)),
-      collapse = " "
-    ),
-    "<SQL SELECT> From:     df Select:   *"
-  )
+  expect_known_output(mf, test_path("test-sql-query-select.txt"), print = TRUE)
 })
--- a/tests/testthat/test-translate-window.R
+++ b/tests/testthat/test-translate-window.R
@@ -49,3 +49,19 @@ test_that("can override frame of recycled functions", {
   )
 
 })
+
+test_that("win_rank() is accepted by the sql_translator", {
+  expect_equal(
+    "<sql_variant> scalar:    test",
+    paste0(
+      capture.output(
+        sql_variant(
+          sql_translator(
+            test = win_rank("test")
+          )
+        )
+      ),
+      collapse = " "
+    )
+  )
+})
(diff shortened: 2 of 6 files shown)


edgararuiz:
ptal @hadley

tests/testthat/test-sql-query.R
```diff
@@ -4,3 +4,14 @@ test_that("add_suffixes works if no suffix requested", {
   expect_equal(add_suffixes(c("x", "x"), "y", ""), c("x", "x"))
   expect_equal(add_suffixes(c("x", "y"), "y", ""), c("x", "y"))
 })
+
+test_that("select_query() print method output is as expected", {
+  mf <- select_query(lazy_frame(x = 1, src = simulate_dbi()))
+  expect_equal(
```
hadley: Can you please use `expect_known_output()` here instead?

tests/testthat/test-translate-window.R
```diff
@@ -49,3 +49,19 @@ test_that("can override frame of recycled functions", {
   )
 
 })
+
+test_that("win_rank() is accepted by the sql_translator", {
```
hadley: This just testing the print method?

tests/testthat/test-sql-query.R
```diff
@@ -4,3 +4,14 @@ test_that("add_suffixes works if no suffix requested", {
...
+    paste0(
+      capture.output(print.select_query(mf)),
```
hadley: You should be able to just use `print()` here, I think?

tests/testthat/test-translate-window.R
edgararuiz: I think I started with `expect_silent()` but felt like we could do more.  Figured that by testing the print, we could catch unintended changes to that output.

tests/testthat/test-sql-query.R
edgararuiz: `expect_known_output()` does not seem to like the test

``` 
library(testthat)
library(dbplyr)

test_that("select_query() print method output is as expected", {
  mf <- select_query(lazy_frame(x = 1, src = simulate_dbi()))
  expect_known_output(
    paste0(
      capture.output(dbplyr:::print.select_query(mf)),
      collapse = " "
    ),
    "<SQL SELECT> From:     df Select:   *"
  )
})
#> Error: Test failed: 'select_query() print method output is as expected'
#> * cannot open the connection
#> 1: expect_known_output(paste0(capture.output(dbplyr:::print.select_query(mf)), collapse = " "), 
#>        "<SQL SELECT> From:     df Select:   *") at <text>:6
#> 2: write_lines(act$out, file)
#> 3: base::writeLines(enc2utf8(text), path, useBytes = TRUE)
#> 4: file(con, "w")
```

<sup>Created on 2019-01-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

tests/testthat/test-sql-query.R
edgararuiz: It looks like the `print()` method is overridden so the outputs won't match:

edgararuiz: library(testthat)
library(dbplyr)

test_that("select_query() print method output is as expected", {
  mf <- select_query(lazy_frame(x = 1, src = simulate_dbi()))
  expect_equal(
    paste0(
      print(dbplyr:::print.select_query(mf)),
      collapse = " "
    ),
    "<SQL SELECT> From:     df Select:   *"
  )
})
#> <SQL SELECT>
#> From:     df
#> Select:   *
#> NULL
#> Error: Test failed: 'select_query() print method output is as expected'
#> * paste0(print(dbplyr:::print.select_query(mf)), collapse = " ") not equal to "<SQL SELECT> From:     df Select:   *".
#> 1/1 mismatches
#> x[1]: ""
#> y[1]: "<SQL SELECT> From:     df Select:   *"

tests/testthat/test-sql-query.R
```diff
@@ -4,3 +4,8 @@ test_that("add_suffixes works if no suffix requested", {
   expect_equal(add_suffixes(c("x", "x"), "y", ""), c("x", "x"))
   expect_equal(add_suffixes(c("x", "y"), "y", ""), c("x", "y"))
 })
+
+test_that("select_query() print method output is as expected", {
+  mf <- select_query(lazy_frame(x = 1, src = simulate_dbi()))
+  expect_known_output(mf, test_path("test-sql-query-select.txt"), print = TRUE)
```
hadley: I just pushed a change in to this change showing how `expect_known_output()` should work. Can you please adapt the other tests to use the same approach?

tests/testthat/test-translate-window.R
hadley: I think this test should live in `test-translate-sql-helpers.r`.

PR #146, opened 2018-08-23
Generic joins

krlmlr:
Proposing a syntax to pass arbitrary join predicates, e.g. for working with spatial data on the database. Needs a test.

``` r
library(tidyverse)
library(dbplyr)
#> 
#> 
#>     ident, sql

tbl1 <- memdb_frame(a = 1:3, b = 4:2)
tbl2 <- memdb_frame(c = 1:3, b = 2:0)

tbl12 <- left_join(tbl1, tbl2, by = sql("TBL_LEFT.b < TBL_RIGHT.c"))
tbl12 %>% sql_render()
#> <SQL> SELECT `TBL_LEFT`.`a` AS `a`, `TBL_LEFT`.`b` AS `b.x`, `TBL_RIGHT`.`c` AS `c`, `TBL_RIGHT`.`b` AS `b.y`
#>   FROM `ygyjgeeqdc` AS `TBL_LEFT`
#>   LEFT JOIN `urhwkxzyib` AS `TBL_RIGHT`
#>   ON (TBL_LEFT.b < TBL_RIGHT.c)
tbl12
#> # Source:   lazy query [?? x 4]
#> # Database: sqlite 3.22.0 [:memory:]
#>       a   b.x     c   b.y
#>   <int> <int> <int> <int>
#> 1     1     4    NA    NA
#> 2     2     3    NA    NA
#> 3     3     2     3     0
```

> Created on 2018-08-23 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

### DIFF
--- a/tests/testthat/test-verb-joins.R
+++ b/tests/testthat/test-verb-joins.R
@@ -27,6 +27,20 @@ test_that("join with both same and different vars", {
   expect_equal(names(j1), c("x", "y"))
 })
 
+test_that("joining over arbitrary predicates", {
+  j1 <- collect(left_join(df1, df2, sql_on = "LHS.x = RHS.b"))
+  j2 <- collect(left_join(df1, df2, by = c("x" = "b"))) %>% mutate(b = x)
+  expect_equal(j1, j2)
+
+  j1 <- collect(left_join(df1, df3, sql_on = "LHS.x = RHS.z"))
+  j2 <- collect(left_join(df1, df3, by = c("x" = "z"))) %>% mutate(z = x.x)
+  expect_equal(j1, j2)
+
+  j1 <- collect(left_join(df1, df3, sql_on = "LHS.x = RHS.x"))
+  j2 <- collect(left_join(df1, df3, by = "x")) %>% mutate(x.y = x) %>% rename(x.x = x)
+  expect_equal(j1, j2)
+})
+
 test_that("inner join doesn't result in duplicated columns ", {
   expect_equal(colnames(inner_join(df1, df1)), c("x", "y"))
 })
--- a/NEWS.md
+++ b/NEWS.md
@@ -39,6 +39,9 @@
 
 * New `tbl_memdb()` for creating reprexes (to match `tbl_lazy()`)
 
+* All `..._join()` methods gain an `sql_on` argument that allows specifying
+  arbitrary join predicates in SQL code (#146, @krlmlr).
+
 * ORACLE: New custom translation for `paste()` and `paste0()` (@cderv, #221)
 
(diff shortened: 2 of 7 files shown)


krlmlr:
@hadley @edgararuiz: Do we want to support joins with custom join predicates? Another way to achieve this is a cross join with a filter, but

- it would work better if we had a `cross_join()` verb (`full_join()` with empty `by` doesn't work in edge cases where column names overlap)
- perhaps some databases perform better with `JOIN ... ON` than with a filter on a cross join in a subquery

hadley:
I'd prefer an approach based around `cross_join()`

krlmlr:
`left_join(tbl1, tbl2, on = sql("TBL_LEFT.b < TBL_RIGHT.c"))`

?

krlmlr:
`left_join(tbl1, tbl2, sql_on = "TBL_LEFT.b < TBL_RIGHT.c")`

hadley:
Yeah, I like that as a temporary stop gap.

krlmlr:
A proper `on` argument that would take an expression would be more in line with the rest of the interface, but is not as useful for spatial joins due to #181.

I wonder if we should rename the aliases to `x` and `y`, or if this would cause too much trouble.

hadley:
I don't think we need any extra work on this interface, as it's mostly meant as a stop gap until we come up with a general interface. But could you please add a couple of very simple tests of the SQL generation? (Just to make sure everything is plumbed together correctly)

krlmlr:
Rebased on current `master`, added a few integration tests. Do we still need tests for the SQL generation?

hadley:
I think it might be better to turn the test into a regression test on the generated SQL? (i.e. using `expect_known_output()`)

tests/testthat/test-verb-joins.R
```diff
@@ -27,6 +27,20 @@ test_that("join with both same and different vars", {
   expect_equal(names(j1), c("x", "y"))
 })
 
+test_that("joining over arbitrary predicates", {
```
hadley: I think this would be better converted to this style:

```R
test_that("generated sql doesn't change unexpectedly", {
  lf <- lazy_frame(x = 1, y = 2)
  reg <- list(
    inner = inner_join(lf, lf),
    left = left_join(lf, lf),
    right = right_join(lf, lf),
    full = full_join(lf, lf)
  )
  expect_known_output(print(reg), test_path("sql/join.sql"))
})
```

krlmlr:
Done, also adapted to the new naming.

hadley:
Thanks!

lorenzwalthert:
This is really nice but I think more people would use it if this was documented. To me, it looks like it is currently missing in [the docs](https://dplyr.tidyverse.org/reference/join.tbl_df.html).

PR #170, opened 2018-10-01
as.double should cast to double precision

ramnathv:
Currently, `as.double` casts to NUMERIC. This leads to an issue in PostGreSQL where aggregate functions like `mean` return integers rather than doubles. This behavior is consistent with the [docs]( https://www.postgresql.org/docs/10/static/datatype-numeric.html), since `NUMERIC` fields need a scale that defaults to 0 for integers.

This PR switches `as.double` to cast to `DOUBLE PRECISION`. Given that there is already an `as.numeric` that casts to `NUMERIC`, I believe this should be alright.


ramnathv:
I am closing this PR as this may not be a generic PostgresSQL issue.

Issue #178, opened 2018-10-26
SQL translation turns aggregation into window function


javierluraschi:
`dbplyr` is adding an unnecessary `OVER ()` hint with a repro similar to:

```r
library(sparklyr)
sc <- spark_connect(master = "local")
sdf_len(sc, 10) %>%
    dplyr::mutate(average = mean(id)) %>%
    dplyr::show_query()
```
```sql
<SQL>
SELECT `id`, avg(`id`) OVER () AS `average`
FROM `sparklyr_tmp_afe063482375`
```

Which causes performance degradation in Spark cluster, see: https://github.com/rstudio/sparklyr/issues/1738

However, this behavior has been around since 2017, probably even earlier, so not a regression at all.

hadley:
This is deliberate behaviour — asking clarifying questions on original issue.

PR #179, opened 2018-10-29
Avoid explicit over() translation when translating SQL

javierluraschi:
Fix for https://github.com/tidyverse/dbplyr/issues/178, some SQL engines, like Spark, treat `OVER()` as a hint to process data differently. In general, I believe generating the minimal SQL statement required to perform an operation makes the most sense; however, I'm not sure if somehow, other SQL engines might require `OVER()` to be explicitly added, I wouldn't think so, but just in case, added a `dplyr.explicit.over` to revert this fix.

CC: @edgararuiz

### DIFF
--- a/R/translate-sql-window.r
+++ b/R/translate-sql-window.r
@@ -58,8 +58,13 @@ win_over <- function(expr, partition = NULL, order = NULL, frame = NULL) {
     frame <- build_sql("ROWS ", frame)
   }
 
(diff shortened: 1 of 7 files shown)


hadley:
I don't believe that this will generate correct SQL. Lets continue discussion in https://github.com/rstudio/sparklyr/issues/1738.

PR #182, opened 2018-11-11
Fix parameter name in vignette

tdsmith:
The name of the parameter that describes the path to the database is `dbname` [1]. Unrecognized keys are ignored. If no dbname is received, an in-memory database is created by default, so the example worked anyway. Fixing this usage avoids misleading users trying to build on this example.

[1] https://github.com/r-dbi/RSQLite/blob/33d3631f2809df4eb2ab01460ef3ce73788e1488/R/connect.R#L46


hadley:
Ooops, thanks!

PR #220, opened 2019-01-17
Remove toupper() from sql_prefix()

### LINKED ISSUE #181: Case sensitivity when translating to SQL
Some SQL functions may be case-sensitive, for instance spatial methods in MSSQL. The syntax is a bit weird, e.g. `colname.STBuffer(1)` for [`STBuffer()`](https://docs.microsoft.com/en-us/sql/t-sql/spatial-geometry/stbuffer-geometry-data-type?view=sql-server-2017), the call `COLNAME.STBUFFER(1)` fails.

``` r
library(tidyverse)
library(dbplyr)
memdb_frame(a = 1) %>%
  mutate(x = a.STBuffer(1)) %>%
  sql_render()
#> <SQL> SELECT `a`, A.STBUFFER(1.0) AS `x`
#> FROM `opozhizxky`
```

<sup>Created on 2018-11-05 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

(This example works identically with MSSQL, using SQLite for the reprex.)

What's the reason for converting all function calls to uppercase?

mtoto:
This fixes #181, I wonder though whether more of `translate-sql` functions should be similarly updated.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dbplyr (development version)
 
+* `sql_prefix()` no longer turns SQL functions into uppercase, allowing for correct translation of case-sensitive SQL functions (#181, @mtoto).
+
 * SQL simulation has been overhauled. It's now better documented, and it always
   uses `` ` `` for field names and `'` for string.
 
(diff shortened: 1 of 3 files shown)


hadley:
Let's merge this PR and then if you want to upper case more of SQL translation, we can do it in another PR.

Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

mtoto:
Thanks for the quick reply, I've added a line to  NEWS.

hadley:
Thanks!

PR #221, opened 2019-01-19
add paste and paste0 translation for ORACLE

### LINKED ISSUE #155: translation of paste for Oracle should be CONCAT
It appears that `paste()` is translated as `CONCAT_WS()`. This function does not exist in Oracle Database, which implements the concatenation operator `||` as well as `CONCAT()` (the latter is why `paste0()` is translated correctly). Accordingly, the translation is not working (as of **`dbplyr`** 1.2.2).

``` r

lf <- dbplyr::lazy_frame(x = "a", src = dbplyr::simulate_oracle())
lf %>% 
  mutate(y = paste(x, "b")) %>% 
  show_query()
#> <SQL> SELECT `x`, CONCAT_WS(' ', `x`,'b') AS `y`
#> FROM (`df`)
```

Created on 2018-09-06 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

Because `paste()` takes an arbitrary number of arguments, the translation could be a little tricky. The below suggestion (inspired by `dbplyr::replace_expr()`) could be a start, though it would need to be extended to handle column names and be more `paste()`-like, e.g., accept `sep` and `collapse`.

``` r
ora_paste_ws <- function(...) {
  x <- list(...)
  
  if (length(x) == 1) {
    dbplyr::escape(x[[1]])
  } else {
    y <- do.call(ora_paste_ws, x[-1])
    dbplyr::build_sql(dbplyr::escape(x[[1]]), " || ", dbplyr::sql(y[[1]]))
  }
}

ora_paste_ws("a", "b", "c")
#> <SQL> 'a' || 'b' || 'c'
```

Created on 2018-09-06 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

cderv:
closes #155 

Following these documentation: 

* https://docs.oracle.com/cd/B19306_01/server.102/b14200/operators003.htm#i997789
* https://blogs.oracle.com/oraclemagazine/working-with-strings

translate `paste` and `paste0` using the infix operator `||`

### DIFF
--- a/tests/testthat/test-backend-oracle.R
+++ b/tests/testthat/test-backend-oracle.R
@@ -20,3 +20,14 @@ test_that("queries translate correctly", {
     sql("^SELECT [*] FROM [(]SELECT [*]\nFROM [(]`df`[)] [)] `[^`]*` WHERE ROWNUM [<][=] 6")
   )
 })
+
+test_that("vectorised translations", {
+  trans <- function(x) {
+    translate_sql(!!enquo(x), con = simulate_oracle(), window = FALSE)
+  }
+
+  expect_equal(trans(paste(x, y)), sql("`x` || ' ' || `y`"))
+  expect_equal(trans(paste(x, y, z)), sql("`x` || ' ' || `y` || ' ' || `z`"))
+  expect_equal(trans(paste(x, y, sep = "#")), sql("`x` || '#' || `y`"))
+  expect_equal(trans(paste0(x, y)), sql("`x` || `y`"))
+})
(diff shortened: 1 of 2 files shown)


tests/testthat/test-backend-oracle.R
```diff
@@ -20,3 +20,14 @@ test_that("queries translate correctly", {
     sql("^SELECT [*] FROM [(]SELECT [*]\nFROM [(]`df`[)] [)] `[^`]*` WHERE ROWNUM [<][=] 6")
   )
 })
+
+test_that("vectorised translations", {
```
hadley: This feels like slightly too much testing because you're effectively testing `sql_paste_infix()`, which should be tested elsewhere. I think a single "sanity" check for `paste()` and `paste0()` should be adequate.

Issue #201, opened 2019-01-03
Sort order of NAs


krlmlr:
NAs sort first in (some) SQL databases (at least MySQL, SQLite and SQL Server), but last in tibbles and PostgreSQL. I wonder what's the best way to document this, we don't seem to have `?dbplyr::arrange` yet.

``` r
library(tidyverse)

dbplyr::memdb_frame(a = c(1, NA)) %>%
  arrange(a)
#> # Source:     table<szlvekywfo> [?? x 1]
#> # Database:   sqlite 3.22.0 [:memory:]
#> # Ordered by: a
#>       a
#>   <dbl>
#> 1    NA
#> 2     1

tibble(a = c(1, NA)) %>%
  arrange(a)
#> # A tibble: 2 x 1
#>       a
#>   <dbl>
#> 1     1
#> 2    NA

dbplyr::memdb_frame(a = c(1, NA)) %>%
  arrange(if_else(is.na(a), 1, 0), a)
#> # Source:     table<liromxakig> [?? x 1]
#> # Database:   sqlite 3.22.0 [:memory:]
#> # Ordered by: if_else(is.na(a), 1, 0), a
#>       a
#>   <dbl>
#> 1     1
#> 2    NA
```

<sup>Created on 2019-01-03 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

mkearney:
DIBS

hadley:
@mkearney just in case it's not obvious, you'll need to document `arrange.tbl_lazy` since currently it's not documented. The closest existing docs would be `copy_to.src_sql`.

yutannihilation:
FYI, `dplyr::arrange()` now documents about `NA`, which is yet different from databases https://github.com/tidyverse/dplyr/pull/4083.

PR #222, opened 2019-01-19
document arrange.tbl_lazy

mkearney:
Add documentation that includes description of sorting behavior for missing values


R/verb-arrange.R
```diff
@@ -1,3 +1,28 @@
+#' Arrange rows by variables in a local lazy tibble
```
hadley: I _think_ it's better to say "remote database table" rather than "local lazy tibble" (just because that's exposing internal implementation details which users shouldn't have to care about)

R/verb-arrange.R
```diff
@@ -1,3 +1,28 @@
...
+#'
+#' Order tbl_lazy rows by an expression involving its variables.
```
hadley: This needs to mention `desc()`, like `?arrange()` does

hadley: And shouldn't mention `tbl_lazy` since that won't mean anything to most people

R/verb-arrange.R
```diff
@@ -1,3 +1,28 @@
...
+#'
+#' @section Missing values:
+#' When applied to `tbl_lazy`, the [arrange()] method sorts NAs first (at the
```
hadley: I think this needs some caveat about "most databases"

R/verb-arrange.R
```diff
@@ -1,3 +1,28 @@
...
+#' beginning) compared to the default sort behavior on data frames, which places
+#' NA values last (at the end).
```
hadley: You can override this behaviour by explicitly sorting on `is.na(x)`

R/verb-arrange.R
```diff
@@ -1,3 +1,28 @@
...
+#' library(dplyr)
+#'
+#' dbplyr::memdb_frame(a = c(3, 4, 1, 2)) %>%
+#'   arrange(a)
+#'
+#' # tbl_lazy: NAs sorted first
+#' dbplyr::memdb_frame(a = c(3, 4, NA, 2)) %>%
+#'   arrange(a)
+#'
+#' # tbl_df: NAs sorted last
```
hadley: I wouldn't show this, but instead show `arrange(is.na(a), a)` to force NAs to sort first.

R/verb-arrange.R
mkearney: I had a feeling this would be the case. It's fixed [here](https://github.com/tidyverse/dbplyr/pull/222/commits/6873d6482a1cfa076b4d49d0787485ab996cc3cd#diff-12fbbf8bf4196bb24a276dff57ec3116R1)

R/verb-arrange.R
mkearney: Fixed [here](https://github.com/tidyverse/dbplyr/pull/222/commits/6873d6482a1cfa076b4d49d0787485ab996cc3cd#diff-12fbbf8bf4196bb24a276dff57ec3116R22)

R/verb-arrange.R
mkearney: I've added a [note about `desc()`](https://github.com/tidyverse/dbplyr/pull/222/commits/6873d6482a1cfa076b4d49d0787485ab996cc3cd#diff-12fbbf8bf4196bb24a276dff57ec3116R7) and [removed `tbl_lazy`](https://github.com/tidyverse/dbplyr/pull/222/commits/6873d6482a1cfa076b4d49d0787485ab996cc3cd#diff-12fbbf8bf4196bb24a276dff57ec3116R6)

R/verb-arrange.R
mkearney: Docs now [say for "most database tables"](https://github.com/tidyverse/dbplyr/pull/222/commits/6873d6482a1cfa076b4d49d0787485ab996cc3cd#diff-12fbbf8bf4196bb24a276dff57ec3116R7).

hadley:
Fixes #201

PR #225, opened 2019-01-21
Consistently remove `toupper()` & uppercase SQL functions

mtoto:
This PR is a follow up on #220 to remove `toupper()` from relevant translate functions and uppercase all SQL (function) strings in the rest of the repo for a consistent style. 

Translate functions that lacked `toupper()` were similarly updated with uppercase SQL strings for consistency,  however as a result I needed to modify a couple of tests that expected standard SQL functions as lowercase.


hadley:
Thank you!

PR #235, opened 2019-02-06
Support bitwise operators

davidchall:
This PR adds support for the base R functions: `bitwNot()`, `bitwAnd()`, `bitwOr()`, `bitwXor()`, `bitwShiftL()`, `bitwShiftR()`.

I've added support for most backends based on their documentation and some manual testing. However, I didn't consider the ODBC and Access backends - are these also needed?

Also, is there some standard way to systematically compare the results produced by each db connection type to those produced by `dplyr`? I saw the figure from https://github.com/tidyverse/dbplyr/pull/131 and was wondering if this was produced by a standard script?


edgararuiz:
HI @davidchall , thank you, this PR is looking great to me!  The way are are running systematic tests against individual database back ends is via infrastructure and a testing package that @colearendt has created.  

@colearendt - Do you think we may be able to run this PR through your `dbtest` setup?

colearendt:
Sure thing! @davidchall do you have some examples of the kinds of mutate, filter, or select statements that might be used with these verbs? I have to admit I am not super familiar with them. The input for the tests is a YAML file with these types of statements. I think some of the currently open PRs have some examples of this YAML syntax, but I am happy to build the file if you have some handy expressions that are good test cases.

R/backend-.R
```diff
@@ -84,6 +84,23 @@ base_scalar <- sql_translator(
...
+  #   Hive: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF#LanguageManualUDF-ArithmeticOperators
+  #   Impala: https://www.cloudera.com/documentation/enterprise/5-9-x/topics/impala_bit_functions.html
+  #   PostgreSQL: https://www.postgresql.org/docs/7.4/functions-math.html
+  #   MS SQL: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/bitwise-operators-transact-sql?view=sql-server-2017
+  #   MySQL https://dev.mysql.com/doc/refman/5.7/en/bit-functions.html
+  #   Oracle: https://docs.oracle.com/cd/E19253-01/817-6223/chp-typeopexpr-7/index.html
+  #   SQLite: https://www.tutorialspoint.com/sqlite/sqlite_bitwise_operators.htm
+  #   Teradata: https://docs.teradata.com/reader/1DcoER_KpnGTfgPinRAFUw/h3CS4MuKL1LCMQmnubeSRQ
+  bitwNot    = function(x) sql_expr(~ !!x),
+  bitwAnd    = function(x, y) sql_infix("&")(x, as.integer(y)),
```
hadley: Shouldn't just this be `bitwAnd = sql_infix("&")`? I don't think you should be coercing the right hand side to an integer.

R/backend-.R
```diff
@@ -84,6 +84,23 @@ base_scalar <- sql_translator(
...
+  # SQL Syntax reference links:
+  #   Hive: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF#LanguageManualUDF-ArithmeticOperators
+  #   Impala: https://www.cloudera.com/documentation/enterprise/5-9-x/topics/impala_bit_functions.html
+  #   PostgreSQL: https://www.postgresql.org/docs/7.4/functions-math.html
+  #   MS SQL: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/bitwise-operators-transact-sql?view=sql-server-2017
+  #   MySQL https://dev.mysql.com/doc/refman/5.7/en/bit-functions.html
+  #   Oracle: https://docs.oracle.com/cd/E19253-01/817-6223/chp-typeopexpr-7/index.html
+  #   SQLite: https://www.tutorialspoint.com/sqlite/sqlite_bitwise_operators.htm
+  #   Teradata: https://docs.teradata.com/reader/1DcoER_KpnGTfgPinRAFUw/h3CS4MuKL1LCMQmnubeSRQ
+  bitwNot    = function(x) sql_expr(~ !!x),
```
hadley: This needs extra parentheses around `x` because the operator precedent rules for R and SQL are not the same. Because unquoting automatically strips on set of parens, you'll need to do `sql_expr(~ ((!!x)))`

tests/testthat/test-backend-impala.R
```diff
@@ -6,7 +6,13 @@ test_that("custom scalar functions translated correctly", {
     translate_sql(!!enquo(x), con = simulate_impala())
   }
 
-  expect_equal(trans(as.Date(x)), sql("CAST(`x` AS VARCHAR(10))"))
-  expect_equal(trans(ceiling(x)), sql("CEIL(`x`)"))
+  expect_equal(trans(bitwNot(x)),       sql("BITNOT(`x`)"))
```
hadley: Can you please put in a separate test?

tests/testthat/test-backend-teradata.R
```diff
@@ -6,19 +6,25 @@ test_that("custom scalar translated correctly", {
...
-  expect_equal(trans(cot(x)),          sql("1 / TAN(`x`)"))
-  expect_equal(trans(nchar(x)),        sql("CHARACTER_LENGTH(`x`)"))
-  expect_equal(trans(ceil(x)),         sql("CEILING(`x`)"))
-  expect_equal(trans(ceiling(x)),      sql("CEILING(`x`)"))
-  expect_equal(trans(atan2(x, y)),     sql("ATAN2(`y`, `x`)"))
-  expect_equal(trans(substr(x, 1, 2)), sql("SUBSTR(`x`, 1.0, 2.0)"))
-
-  expect_error(trans(paste(x)),        sql("not supported"))
+  expect_equal(trans(x != y),           sql("`x` <> `y`"))
+  expect_equal(trans(bitwNot(x)),       sql("BITNOT(`x`)"))
```
hadley: And these should go in a new test too.

R/backend-.R
davidchall: Thanks for catching this! I've made the change 👍

tests/testthat/test-backend-impala.R
davidchall: 👍

tests/testthat/test-backend-teradata.R
davidchall: 👍

R/backend-.R
davidchall: Bitwise operations require that both arguments are integers. My thoughts were:
- If `x` is a column then I don't want to implicitly cast this to an integer (preventing the user from accidentally applying bitwise logic to floating point columns).
- If `y` is a single integer (e.g. `128`) then I want to coerce this to an integer (so the user doesn't need to type `bitwAnd(x, as.integer(128))` or `bitwAnd(x, 128L)`).

Of course, now I realize that we should actually treat both arguments the same. This leaves us with 3 possibilities:
1. Coerce both arguments
2. Coerce neither argument
3. Coerce arguments only if they aren't referring to a column

The base R functions have chosen option 1.

davidchall:
Thank you all for your feedback! Let's wait until I've resolved the final comment from @Hadley before running `dbtest`.

@colearendt Looking at your examples, I came up with the following YAML file. I've not seen the values in the `fld_integer` column, so it's possible that some of the hardcoded values are inappropriate.

Also, it's important to note that the results can depend on whether we're using a signed or unsigned integer data type. Will be interesting to see if anything is flagged due to this.

```yaml
- bitwNot:
    mutate: bitwNot(fld_integer)
    filter: bitwNot(fld_integer) < 0
    summarize: n_distinct(bitwNot(fld_integer))
    group_by: bitwNot(fld_integer)
    arrange: bitwNot(fld_integer)
- bitwAnd:
    mutate: bitwAnd(fld_integer, 8)
    filter: bitwAnd(fld_integer, 8)
    summarize: n_distinct(bitwAnd(fld_integer, 8))
    group_by: bitwAnd(fld_integer, 8)
    arrange: bitwAnd(fld_integer, 8)
- bitwOr:
    mutate: bitwOr(fld_integer, 8)
    filter: bitwOr(fld_integer, 8)
    summarize: n_distinct(bitwOr(fld_integer, 8))
    group_by: bitwOr(fld_integer, 8)
    arrange: bitwOr(fld_integer, 8)
- bitwXor:
    mutate: bitwXor(fld_integer, 8)
    filter: bitwXor(fld_integer, 8)
    summarize: n_distinct(bitwXor(fld_integer, 8))
    group_by: bitwXor(fld_integer, 8)
    arrange: bitwXor(fld_integer, 8)
- bitwShiftL:
    mutate: bitwShiftL(fld_integer, 1)
    filter: bitwShiftL(fld_integer, 1) > 10
    summarize: n_distinct(bitwShiftL(fld_integer, 1))
    group_by: bitwShiftL(fld_integer, 1)
    arrange: bitwShiftL(fld_integer, 1)
- bitwShiftR:
    mutate: bitwShiftR(fld_integer, 1)
    filter: bitwShiftR(fld_integer, 1) > 10
    summarize: n_distinct(bitwShiftR(fld_integer, 1))
    group_by: bitwShiftR(fld_integer, 1)
    arrange: bitwShiftR(fld_integer, 1)
```

R/backend-.R
hadley: I think you should coerce neither, and rely on the database to do it (or otherwise complain).

davidchall: Ok, I've made the change 👍

davidchall:
@colearendt Ok, I think it should be good to test now - and I've updated the YAML file above to account for the latest changes. Thanks!

hadley:
Thanks!

colearendt:
@davidchall Thanks for writing the YAML file! I will include that into our `dbtest` suite. Overall, it looks really good. Just a few problems (on Postgres / MSSQL, which is what I currently have handy to test):

image

Looks like semantics of "where" clauses are different in SQL (i.e. WHERE integer) does not get handled appropriately. I don't know what would possibly fix this. It's almost like you need a context aware translation - didn't you run into this somewhere @edgararuiz ?

```
> t %>% filter(bitwAnd(fld_integer, 8L))
  nanodbc/nanodbc.cpp:1344: 42804: [RStudio][PostgreSQL] (30) Error occurred while trying to execute a query: [SQLState 42804] ERROR:  argument of WHERE must be type boolean, not type integer
LINE 3: WHERE ("fld_integer" & 8)
               ^
 
> t %>% filter(bitwAnd(fld_integer, 8L)) %>% show_query()
<SQL>
SELECT *
FROM "blah"
WHERE ("fld_integer" & 8)
```

The other MSSQL "failures" are just "not available on this SQL variant"

davidchall:
@colearendt Ah, this does make sense. The result of bitwise operations is also an integer, but the WHERE clause expects to receive a boolean value. I've adjusted the YAML file to account for this.

PR #130, opened 2018-07-19
Add lubridate

colearendt:
Add some `lubridate` functionality.  Several tests are failing, so we still need a bit of cleanup.

```r
> prepdata %>% mutate(as_date(fld_datetime)) %>% pull()
 [1] "2015-01-01" "2015-02-01" "2015-03-01" "2015-04-01" "2015-05-01" "2015-06-01" "2015-07-01" "2015-08-01" "2015-09-01"
[10] "2015-10-01"
> testdata %>% mutate(as_date(fld_datetime)) %>% pull()
 [1] "1-01-15"  "2-01-15"  "3-01-15"  "4-01-15"  "5-01-15"  "6-01-15"  "7-01-15"  "8-01-15"  "9-01-15"  "10-01-15"
```

Not quite sure what the best way to cause this test to succeed is... `dbtest` needs improvement for date and datetime types (`copy_to` turns them into text values, even if they are dates at the outset)

PR #248, opened 2019-02-25
Add main lubridate function support to Oracle

rlh1994:
Hi,

I have added as many of the (at least from my experience) key lubridate functions to be translated into Oracle SQL as possible (they may work for other SQL variants but I do not have experience with those). This includes most of the getting and setting functions (those not included are mainly due to Oracle not having those measures and the date math just gets too complex, at least for me), floor/ceiling/round_date, and one or two of the other functions as I needed a break from the bigger ones.

There's currently 5 outstanding issues with the code:
1) I cannot work out how to have round_date call floor/ceiling_date in the same definition, it should be a one liner but I am not sure how to do it.
2) I have had to re-introduce the dependency on stringR to parse the units, again this was because I couldn't work out how to access the function that now does this for the actual lubridate package so took the original parsing code. As the parsing is done R-side it should be able to be removed I again just don't know how.
3) There is little to no support for timestamp datatypes and therefore timezones. Most of the functions have been built explicitly with a call to to_date or date related functions that will not retain timestamp info. In part this was due to trying to keep it simple, and in part due to my lack of experience with Oracle timestamps and timezones. In addition to this, a lot of timestamp info is related to the NLS setting of the particular database which make it very hard to verify via testing.
4) There are very few tests; I've been mostly focused on making sure that the SQL actually returns the same results as a pure lubridate call would, so I have been using a table in a database to compare calls, which means I haven't yet written the actual tests for the SQL text itself into the tests file, but all of my tests are now passing. I have also been unable to use dbtest as I only have access to the Oracle DB though an ROracle connection which is not currently supported. 
5) as.Date() (which has to be defined not using cast as this is NLS setting dependant in Oracle at least) currently doesn't return the expected error it should. I still need to look into this one and do more testing for that function as a whole as the sheer volume of possible combinations mean I kept putting that off.

Finally, sorry for all the commits, it's my first PR so didn't realise I should have made another branch first.


cderv:
Just for reference, I think this PR follows a comment on https://github.com/tidyverse/dbplyr/pull/130#issuecomment-457499445 - this linked issue being the main PR for adding lubridate support.

hadley:
There seems to be a lot of differences, so you would mind starting again with a fresh PR? Please start with the core lubridate functions that we're planning to support in the other backends, i.e. `today()`, `now()`,  `year()`, `month()` (numeric value only), `day()`, `hour()`, `minute()`, and `second()`.

rlh1994:
Yes can do. Year-day should be identical although I need to verify the column type dbplyr pulls this as as I thought it had treated it as chr when I tried it before from Oracle however this may not be the case (and is not the case in-db). 
For hour-second these will work for timestamp type columns but will fail for date type columns (no datetime object in Oracle, all date objects have a time). Are you looking to support date columns as well or only timestamp?

hadley:
If the type of the output is wrong, it's unlikely to be resolvable in dbplyr, and will instead need to happen upstream somewhere.

Ideally we should support both date and timestamp columns, but you can assume the database will return an error if (e.g.) you attempt to extract the minute from a date.

Issue #261, opened 2019-03-13
Provide lubridate translations for SQL server


hadley:
https://docs.microsoft.com/en-us/sql/t-sql/functions/datepart-transact-sql?view=sql-server-2017

hadley:
Done in #271

PR #271, opened 2019-03-28
preliminary support for lubridate for SQL Server

derekmorr:
This adds preliminary support for lubridate functions for SQL Server (re: #261 ).

### DIFF
--- a/R/backend-mssql.R
+++ b/R/backend-mssql.R
@@ -70,6 +70,7 @@
       ifelse         = function(test, yes, no) mssql_sql_if(test, yes, no),
 
       as.logical    = sql_cast("BIT"),
+      as.Date       = sql_cast("DATETIME2"),
       as.numeric    = sql_cast("NUMERIC"),
       as.double     = sql_cast("NUMERIC"),
       as.character  = sql_cast("VARCHAR(MAX)"),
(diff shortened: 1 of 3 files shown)


R/backend-mssql.R
```diff
@@ -96,7 +97,48 @@
       str_c = sql_paste_infix("", "+", function(x) sql_expr(cast(!!x %as% text))),
       # no built in function: https://stackoverflow.com/questions/230138
       str_to_title = sql_not_supported("str_to_title()"),
-      str_sub = sql_str_sub("SUBSTRING", "LEN")
+      str_sub = sql_str_sub("SUBSTRING", "LEN"),
+
+      # lubridate ---------------------------------------------------------------
+      # https://en.wikibooks.org/wiki/SQL_Dialects_Reference/Functions_and_expressions/Date_and_time_functions
+      as_date = sql_cast("DATE"),
+      as_datetime = sql_cast("DATETIME2"),
```
hadley: Can you provide a comment that describes why this is `DATETIME2`?

R/backend-mssql.R
```diff
@@ -96,7 +97,48 @@
...
+      second = function(x) sql_expr(DATEPART(second, !!x)),
+
+      month = function(x, label = FALSE, abbr = TRUE) {
+        if (!label) {
+          sql_expr(DATEPART(month, !!x))
+        } else {
+          if (!abbr) {
+            sql_expr(DATENAME(month, !!x))
+          } else {
+            stop("abbr not supported for month() on SQL Server", call. = FALSE)
```
hadley:
```suggestion
            stop("`abbr` is not supported in SQL Server translation", call. = FALSE)
```

R/backend-mssql.R
```diff
@@ -96,7 +97,48 @@
...
+          }
+        }
+      },
+
+      quarter = function(x, with_year = FALSE, fiscal_start = 1) {
+        if (fiscal_start != 1) {
+          stop("fiscal_start parameter is not supported. Must be 1.", call. = FALSE)
```
hadley:
```suggestion
          stop("`fiscal_start` is not supported in SQL Server translation. Must be 1.", call. = FALSE)
```

hadley:
Could you please edit `NEWS.md` to include this in the description of the new lubridate translations?

R/backend-mssql.R
derekmorr: Sure. I'm not a SQL Server expert, but my understanding is that DATETME2 is the newer preferred alternative to DATETIME.

derekmorr: However, `DATETIME2` is only supported in SQL Server 2008 and newer. I don't know what version of SQL Server dbplyr supports. If lack of support for pre-2008 versions is a problem, we can change it to `DATETIME`.

hadley: A link to a discussion of the differences would be helpful either way. Is SQL server 2008 in widespread use? It is 10 years old.

derekmorr:
Looks like I need to add tests for this (which I should have done originally). I'll get to those tonight.

derekmorr:
I believe I've addressed your requested changes.

I used `DATETIME2` because in Microsoft's [SQL Server docs](https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-2017), they say:

> Use the time, date, datetime2 and datetimeoffset data types for new work. These types align with the SQL Standard. They are more portable. time, datetime2 and datetimeoffset provide more seconds precision. datetimeoffset provides time zone support for globally deployed applications.

I linked to that page in the code.

PR #272, opened 2019-03-29
Fix explain() for Postgres data sources

krlmlr:
What's a good place to add tests, also for other data sources?

### DIFF
--- a/tests/testthat/sql/postgres-explain.sql
+++ b/tests/testthat/sql/postgres-explain.sql
@@ -1,0 +1,6 @@
+<SQL>
+SELECT "x", "x" + 1.0 AS "y"
+FROM "dbplyr_002"
+
+<PLAN>
+Seq Scan on dbplyr_002  (cost=0.00..1.04 rows=3 width=36)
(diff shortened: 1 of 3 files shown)


hadley:
In `test-backend-postgres.R` etc

krlmlr:
`db_explain()` currently builds the query (where the error occurred) and runs it against the database. Testing this seems to require a live connection to the database. Should we add a new `test-verb-explain.R` ?

hadley:
Tests should live in the test file corresponding to the R file. It is fine to use an actual database connection in the backend tests.

krlmlr:
Added test. I still think it's worthwhile to test `explain()` on all sources.

tests/testthat/sql/postgres-explain.sql
```diff
@@ -0,0 +1,6 @@
+<SQL>
+SELECT "x", "x" + 1.0 AS "y"
+FROM "dbplyr_001"
+
+<PLAN>
+Seq Scan on dbplyr_001  (cost=0.00..1.04 rows=3 width=36)
```
hadley: It just occurs to me that this might cause problems if postgres changes the output format across versions. But we can cross that bridge when we come to it.

hadley:
Tests for other EXPLAIN implementations would also be appreciated; just my experience is that it's easier to keep all the database specific stuff together, rather than keeping (e.g.) all the explain stuff together.

PR #277, opened 2019-04-09
Use TOP 100 PERCENT as stop-gap to allow subqueries with ORDER BY

### LINKED ISSUE #275: Support ORDER BY in subqueries for MSSQL
MSSQL is sensitive to subqueries that have an `ORDER BY` clause -- they are not allowed. A [suggested workaround](https://stackoverflow.com/a/985953/946850) is to use `TOP 100 PERCENT`~~, but I believe we can also solve this by more accurate optimization~~.

This is an actual problem if window functions are involved, in this case it's not easy to move the `arrange()` as noted in https://github.com/tidyverse/dbplyr/issues/94#issuecomment-460789909.

``` r
library(tidyverse)
library(dbplyr)

df <- lazy_frame(x = 1:4, g = rep(c(1, 2), each = 2), con = simulate_mssql())

df %>%
  group_by(g) %>%
  arrange(x) %>%
  mutate(r = cumsum(x))
#> <SQL>
#> SELECT `x`, `g`, SUM(`x`) OVER (PARTITION BY `g` ORDER BY `x` ROWS UNBOUNDED PRECEDING) AS `r`
#> FROM (SELECT *
#> FROM `df`
#> ORDER BY `x`) `dbplyr_001`
```

<sup>Created on 2019-04-09 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

krlmlr:
for MSSQL.

Closes #275.

### DIFF
--- a/R/query-select.R
+++ b/R/query-select.R
@@ -100,8 +100,12 @@ sql_render.select_query <- function(query, con, ..., bare_identifier_ok = FALSE)
   )
 
   sql_select(
-    con, query$select, from, where = query$where, group_by = query$group_by,
-    having = query$having, order_by = query$order_by, limit = query$limit,
+    con, query$select, from,
+    where = query$where,
+    group_by = query$group_by,
+    having = query$having,
+    order_by = query$order_by,
+    limit = query$limit,
     distinct = query$distinct,
     ...,
     bare_identifier_ok = bare_identifier_ok
--- a/R/query-select.R
+++ b/R/query-select.R
@@ -103,7 +103,8 @@ sql_render.select_query <- function(query, con, ..., bare_identifier_ok = FALSE)
     con, query$select, from, where = query$where, group_by = query$group_by,
     having = query$having, order_by = query$order_by, limit = query$limit,
     distinct = query$distinct,
-    ...
+    ...,
+    bare_identifier_ok = bare_identifier_ok
   )
 }
 
--- a/tests/testthat/test-backend-mssql.R
+++ b/tests/testthat/test-backend-mssql.R
@@ -166,7 +166,6 @@ test_that("ORDER BY in subqueries uses TOP 100 PERCENT (#175)", {
 
   expect_equal(
     mf %>% mutate(x = -x) %>% arrange(x) %>% mutate(x = -x) %>% sql_render(),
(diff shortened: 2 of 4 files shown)


hadley:
Re-building

tests/testthat/test-backend-mssql.R
```diff
@@ -160,3 +160,12 @@ FROM `df`")
     sql("SELECT `x`, CASE\nWHEN (((`x`) IS NULL)) THEN (1.0)\nWHEN (NOT(((`x`) IS NULL))) THEN (2.0)\nELSE (3.0)\nEND AS `z`\nFROM `df`")
   )
 })
+
+test_that("ORDER BY in subqueries uses TOP 100 PERCENT (#175)", {
+  mf <- lazy_frame(x = 1:3, con = simulate_mssql())
+
+  expect_equal(
```
hadley: Maybe it's better to use `expect_known_output()` here?

R/query-select.R
```diff
@@ -103,7 +103,8 @@ sql_render.select_query <- function(query, con, ..., bare_identifier_ok = FALSE)
     con, query$select, from, where = query$where, group_by = query$group_by,
     having = query$having, order_by = query$order_by, limit = query$limit,
     distinct = query$distinct,
-    ...
+    ...,
```
hadley: Since you touched this function, can you please put each of the other arguments on its own line?

tests/testthat/test-backend-mssql.R
krlmlr: This code is consistent with the rest of the file. I filed #285 to address this.

R/query-select.R
krlmlr: Done. I kept the unnamed arguments on one line.

alexkyllo:
Azure Data Warehouse doesn't support TOP [...] PERCENT so this broke dbplyr for ADW. 

Would it be feasible to emit this only when the subquery actually has an order by clause?

tschager:
Any workaround for Azure Data Warehouse besides sticking to an older version?

filipwastberg:
This also breaks code where the connection is a Parallell Data Warehouse (PDW) with MSSQL, doesn't support `top ... percent` either.

PR #278, opened 2019-04-09
Use as.data.frame() instead of overriding class

krlmlr:
Slightly cleaner.


hadley:
Re-building with cleared cache (to hopefully avoid remotes problem)

hadley:
Thanks

PR #50, opened 2017-10-29
Adds custom db_copy_to method for MS SQL

edgararuiz:
Fixes https://github.com/tidyverse/dplyr/issues/3174

It adds a custom `db_copy_to()` method for MS SQL connections that is a copy of the base `odbc` `db_copy_to()` but it prefixes the `##` if needed and executes the `dbWriteTable()` command with a `temporary` argument set to `FALSE`:

Test results: 

```
> test <- copy_to(con, mtcars, "mtcars2")
Created a temporary table named: ##mtcars2
> test
# Source:   table<##mtcars2> [?? x 12]
# Database: Microsoft SQL Server 14.00.1000[dbo@DESKTOP-10DBTVP\SQLEXPRESS/master]
           row_names   mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
               <chr> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
 1         Mazda RX4  21.0     6 160.0   110  3.90 2.620 16.46     0     1     4     4
 2     Mazda RX4 Wag  21.0     6 160.0   110  3.90 2.875 17.02     0     1     4     4
 3        Datsun 710  22.8     4 108.0    93  3.85 2.320 18.61     1     1     4     1
 4    Hornet 4 Drive  21.4     6 258.0   110  3.08 3.215 19.44     1     0     3     1
 5 Hornet Sportabout  18.7     8 360.0   175  3.15 3.440 17.02     0     0     3     2
 6           Valiant  18.1     6 225.0   105  2.76 3.460 20.22     1     0     3     1
 7        Duster 360  14.3     8 360.0   245  3.21 3.570 15.84     0     0     3     4
 8         Merc 240D  24.4     4 146.7    62  3.69 3.190 20.00     1     0     4     2
 9          Merc 230  22.8     4 140.8    95  3.92 3.150 22.90     1     0     4     2
10          Merc 280  19.2     6 167.6   123  3.92 3.440 18.30     1     0     4     4
# ... with more rows
```

PR #280, opened 2019-04-09
Fix copy_to() for MSSQL databases

krlmlr:
A regression.

The partial `NextMethod()` calls didn't work for me, only when the full list of arguments is specified.

### DIFF
--- a/R/backend-mssql.R
+++ b/R/backend-mssql.R
@@ -144,7 +144,14 @@ mssql_temp_name <- function(name, temporary){
                             analyze = TRUE, ...) {
   NextMethod(
     table = mssql_temp_name(table, temporary),
-    temporary = FALSE
+    values = values,
+    overwrite = overwrite,
+    types = types,
+    temporary = FALSE,
+    unique_indexes = unique_indexes,
+    indexes = indexes,
+    analyze = analyze,
+    ...
   )
 }
 
--- a/R/backend-mssql.R
+++ b/R/backend-mssql.R
@@ -161,7 +161,6 @@ mssql_temp_name <- function(name, temporary){
   name <- mssql_temp_name(name, temporary)
 
   tt_sql <- build_sql(
-    #"SELECT ", escape(ident(vars), collapse = ", ", con = con), " ",
     "SELECT * ",
     "INTO ", as.sql(name), " ",
     "FROM (", sql, ") AS temp",
--- a/R/backend-mssql.R
+++ b/R/backend-mssql.R
@@ -160,6 +160,7 @@ mssql_temp_name <- function(name, temporary){
                                                  temporary = TRUE, ...){
   name <- mssql_temp_name(name, temporary)
 
+  # Different syntax for MSSQL: https://stackoverflow.com/q/16683758/946850
   tt_sql <- build_sql(
     "SELECT * ",
     "INTO ", as.sql(name), " ",
--- a/R/backend-mssql.R
+++ b/R/backend-mssql.R
@@ -144,14 +144,9 @@ mssql_temp_name <- function(name, temporary){
                             analyze = TRUE, ...) {
   NextMethod(
     table = mssql_temp_name(table, temporary),
-    values = values,
-    overwrite = overwrite,
     types = types,
-    temporary = FALSE,
-    unique_indexes = unique_indexes,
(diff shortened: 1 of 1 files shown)


R/backend-mssql.R
```diff
@@ -144,24 +144,42 @@ mssql_temp_name <- function(name, temporary){
...
   )
 }
 
 #' @export
 `db_save_query.Microsoft SQL Server` <- function(con, sql, name,
                                                  temporary = TRUE, ...){
-  NextMethod(
-    name = mssql_temp_name(name, temporary),
-    temporary = FALSE
+  name <- mssql_temp_name(name, temporary)
```
hadley: Is there some reason you can no longer use `NextMethod()` here?

krlmlr: The syntax is different in MSSQL. Originally implemented in #50, f283e03 reverted to using the parent method.

https://stackoverflow.com/q/16683758/946850

Adding a comment to the code.

krlmlr:
I have reduced the `NextMethod()` calls somewhat.

This isn't exactly what I've seen, but something weird is going on:

``` r
foo <- function(x, y = 1, ...) UseMethod("foo")

foo.bar <- function(x, y = 2, z = 2, ...) NextMethod(z = z + 1, b = 5)
foo.baz <- function(x, y = 3, z = 3, ...) moo(x, y, z, ...)

moo <- function(x, y, z, ...) UseMethod("moo")
moo.bar <- function(x, y, z, ...) NextMethod(y = 6)
moo.baz <- function(x, y, z, ...) list(y = y, z = z, ...)

foo(structure(list(), class = c("bar", "baz")), y = 2)
#> $y
#> [1] 6
#> 
#> $z
#> [1] 2
#> 
#> [[3]]
#> [1] 3
#> 
#> $b
#> [1] 5
```

<sup>Created on 2019-04-11 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

krlmlr:
For reference, this is what I see on a live database:

``` r
library(tidyverse)
library(DBI)
library(dbplyr)

mssql_con <- function() {
  DBI::dbConnect(
    odbc::odbc(),
    Driver = "SQL Server Driver",
    ...,
    TDS_Version = 7.4
  )
}

con <- mssql_con()

dbExecute(con, "SET NOCOUNT ON")
#> [1] 0

src <- src_dbi(con)

iris_tbl <- copy_to(src, iris, name = "iris", overwrite = TRUE, temporary = FALSE)
#> Error in (function (classes, fdef, mtable) : unable to find an inherited method for function 'dbWriteTable' for signature '"Microsoft SQL Server", "SQL", "character"'
```

<sup>Created on 2019-04-11 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

hadley:
Hmmm, maybe it's some weird interaction with S4.

R/backend-mssql.R
```diff
@@ -144,23 +144,35 @@ mssql_temp_name <- function(name, temporary){
...
+
+  # Different syntax for MSSQL: https://stackoverflow.com/q/16683758/946850
+  tt_sql <- build_sql(
+    "SELECT * ",
+    "INTO ", as.sql(name), " ",
+    "FROM (", sql, ") AS temp",
+    con = con
   )
+  dbExecute(con, tt_sql)
```
hadley: Could we test this without an activate MS SQL backend by mocking `dbExecute()` to just return the sql, and then using `expect_known_output()`?

krlmlr: Perhaps with an approach similar to dbi-log. But then we still need to double-check the generated SQL.

Would you like me to open a new issue to discuss?

krlmlr:
The reprex in https://github.com/tidyverse/dbplyr/pull/280#issuecomment-482290759 doesn't use S4.

R/backend-mssql.R
hadley: No, I just want a simple way to test this code without adding additional complexity or new dependencies.

krlmlr: Filed an issue: #284. Suggesting to rely on user testing for this PR, discuss/implement #284 for the v1.4.0 release and add some tests too.

PR #283, opened 2019-04-12
Relax output test

krlmlr:
Follow-up to #272.


tests/testthat/test-backend-postgres.R
```diff
@@ -62,11 +62,12 @@ test_that("postgres can explain (#272)", {
 
   df1 <- data.frame(x = 1:3)
 
-  expect_known_output(
+  expect_output(
```
hadley: I think `expect_error(..., NA)` would better express the intent here

PR #288, opened 2019-04-24
Fix typos in NEWS, vignettes, documentation

dpprdan:
This fixes some typos. I renamed "mysql", "postgresql" and "sqlite" to "MySQL", "PostgreSQL" and "SQLite" where appropriate. I hope that's okay. 

While reviewing, I was wondering whether [this line](https://github.com/tidyverse/dbplyr/blob/2f17ed749e4f634c64cf1f99f1a9634f7d4378a5/vignettes/dbplyr.Rmd#L200) in the comparison of PostgreSQL with SQLite is still appropriate in its current wording, given that SQLite now supports window functions.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -50,6 +50,7 @@ Suggests:
 VignetteBuilder: 
     knitr
 Encoding: UTF-8
+Language: en-gb
 LazyData: yes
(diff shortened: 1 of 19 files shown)


hadley:
Thanks! 

I'll remove that line.

PR #297, opened 2019-05-08
WIP: Get rid of dbi_quote()

krlmlr:
Replace only instance by existing generic.


hadley:
I think unfortunately the fix is more complicated than that, and it doesn't seem very high priority to fix.

PR #298, opened 2019-05-08
Avoid unnecessary calls to dbQuoteIdentifier()

krlmlr:
to reduce noise when logging DBI calls.


hadley:
It seems odd to log calls to `dbQuoteIdentifier()`?

krlmlr:
In general yes, but it may be helpful for those DBI tests that test quoting and unquoting.

I'll fix in DBIlog.

Issue #284, opened 2019-04-14
DDL tests for simulated DBI


krlmlr:
Methods like `db_compute()` and `db_copy_to()` currently require a real database to be tested against. Suggesting the following approach:

1. Define a `db_execute()` generic, implement it only for the `"TestConnection"` class (which would print the SQL) and for `"DBIConnection"` (which forwards to `DBI::dbExecute()`)
2. Replace all calls to `DBI::dbExecute()` with calls to `db_execute()`
3. Run tests with `simulate_dbi(...)`, capture output

Reference: https://github.com/tidyverse/dbplyr/pull/280/files#r275091349

hadley:
If we going to rearchitect stuff, I think it would be better to extract out a separate method that just generates the SQL.

hadley:
Looks like these are the main generics that generate then execute SQL:

* `db_explain`
* `db_analyze`
* `db_save_query`
* `db_create_index` — but only used by MySQL and now I don't see why* 
* `db_drop_table`

I'm going to add matching `sql_*()` generics, and change the default `db_*()` to call use `sql_*` + `dbExecute()`.

hadley:
`db_save_query()` is a little trickier because it has to return the name, and MS SQL needs to mangle it. So probably needs another new generic. (Which I think SAP HANA will also use)

PR #308, opened 2019-05-28
make explain work with redshift

ramnathv:
Redshift EXPLAIN does not accept a `(FORMAT ***)` statement following it, in contrast to PostgresSQL. This PR allows the user to pass `format = NULL` explicitly to allow `explain()` to work with `redshift`.

I am unable to provide a reproducible example, since I don't have a public instance of a redshift cluster I can connect to.

### DIFF
--- a/R/backend-postgres.R
+++ b/R/backend-postgres.R
@@ -199,7 +199,6 @@ db_explain.PostgreSQLConnection <- function(con, sql, format = "text", ...) {
     sql,
     con = con
   )
-  print(exsql)
   expl <- dbGetQuery(con, exsql)
 
   paste(expl[[1]], collapse = "\n")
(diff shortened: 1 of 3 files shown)


hadley:
What connection class does RedShift use? I'd prefer to fix this by providing a more specific method.

ramnathv:
The connection class used by RedShift `tbl_PqConnection`. I am happy to rewrite my PR in terms of a method if you can provide some guidance.

hadley:
Ok then that won't work because it's a postgres class. In that case, I think the place to start is to get that class tweaked. How exactly are you connecting to Redshift?

ramnathv:
I am connecting to redshift using `DBI::dbConnect` using the driver `RPostgres::Postgres()`. Redshift for all practical purposes supports Postgres syntax, but there are a few unsupported features, and a few extensions. Do you think the way forward is to define a separate class for `redshift` and then extend the dbplyr API for postgres? I would be happy to take a stab at this if you think this is the design direction that should be taken.

hadley:
I've added the basic layer for redshift translation both here and in RPostgres. Can you please make a new PR (since it'll be very different to this one) creating a new `db_explain.RedshiftConnection` method? There's no need to test it until we get #284 working.

Issue #337, opened 2019-07-24
TOP 100 PERCENT is issued for MSSQL in absence of ORDER BY, breaking support for Azure DW


alexkyllo:
PR #277 causes the mssql backend to issue `SELECT TOP 100 PERCENT` in sub queries even if there is no `ORDER BY` clause. This breaks dbplyr for Azure Data Warehouse, which is a version of MSSQL that does not support `SELECT [...] PERCENT` syntax. I'm working on a PR to change this so that `TOP 100 PERCENT` at least only gets issued if there is actually an `ORDER BY`, so that ADW users can use dbplyr again.

```r
mf <- lazy_frame(x = 1:3, con = simulate_mssql())
mf %>% mutate(x = -x) %>% mutate(x = -x) %>% sql_render()
#> <SQL> SELECT -`x` AS `x`
#> FROM (SELECT TOP 100 PERCENT -`x` AS `x`
#> FROM `df`) `dbplyr_001`
```

filipwastberg:
#277 also breaks code for Parallell Data Warehouse for MSSQL which doesn't support TOP ... PERCENT

filipwastberg:
We hade to roll back to `dbplyr` `1.4.0` to be able to use `dplyr` with our database. 

If anyone wonder how: `devtools::install_version("dbplyr", version = "1.4.0", repos = "http://cran.us.r-project.org")`

Issue #276, opened 2019-04-09
Rethink optimization of ORDER BY


krlmlr:
`ORDER BY` clauses in subqueries are problematic:

- MSSQL outrights forbids them (#275)
- Other DBMS sometimes ignore them (see reprex) and are not obliged to follow either (https://mariadb.com/kb/en/library/why-is-order-by-in-a-from-subquery-ignored/, https://dba.stackexchange.com/a/83500/6934)

As a stop gap, we could remove `ORDER BY` clauses from inner queries if we don't combine the inner and the outer query. With a warning. This would also solve #275. I wonder if this generates spurious warnings, e.g., for window functions.

Alternatively, we could allow `arrange()` only at the very last step in the pipe, and transform it to `window_order()` if additional verbs are added, with a warning. This will be a problem later when we want to support lazy operations for data frames.

For a true solution, we cannot simply delay the computation due to aliasing (https://github.com/tidyverse/dbplyr/issues/94#issuecomment-395599454). A few examples, and a reprex illustrating the problem (using #277), are below. This might well be out of scope.

```r
# Input:
lazy_frame(a = 1:3, b = 4:2) %>%
  mutate(a = -a) %>%
  arrange(a) %>%
  mutate(a = -a)

# Equivalent with `arrange()` delayed:
lazy_frame(a = 1:3, b = 4:2) %>%
  mutate(a = -a) %>%
  mutate(zzz_001 = a) %>%
  mutate(a = -a) %>%
  arrange(zzz_001) %>%
  select(-zzz_001)

# Input:
lazy_frame(a = 1:3, b = 4:2) %>%
  arrange(a) %>%
  group_by(b) %>%
  mutate(c = cumsum(a)) %>%
  ungroup()

# Equivalent with `arrange()` delayed:
lazy_frame(a = 1:3, b = 4:2) %>%
  window_order(a) %>%
  group_by(b) %>%
  mutate(c = cumsum(a)) %>%
  ungroup() %>%
  arrange(a)
```

``` r
library(tidyverse)
devtools::load_all("~/git/R/dbplyr")
#> Loading dbplyr
#> 
#> Attaching package: 'testthat'
#> The following object is masked from 'package:dplyr':
#> 
#>     matches
#> The following object is masked from 'package:purrr':
#> 
#>     is_null
#> Registering testing src: df OK
#> Registering testing src: sqlite OK
#> Registering testing src: mysql OK
#> Registering testing src: MariaDB OK
[...]

PR #338, opened 2019-07-24
mssql backend: only emit TOP 100 PERCENT if there is an ORDER BY

alexkyllo:
This PR is to address Issue #337


R/backend-mssql.R
```diff
@@ -22,7 +22,7 @@
       # e.g: SELECT TOP 100 * FROM my_table
       assert_that(is.numeric(limit), length(limit) == 1L, limit > 0)
       build_sql("TOP(", as.integer(limit), ") ", con = con)
-    } else if (!is.null(order_by) && bare_identifier_ok) {
+    } else if (length(order_by) > 0 && bare_identifier_ok) {
```
krlmlr:
```suggestion
    } else if (has_length(order_by) && bare_identifier_ok) {
```

krlmlr:
Closes #337.

alexkyllo:
Another possibility is instead of emitting `TOP 100 PERCENT` we could emit `TOP 9223372036854775807` (or `TOP 0x7fffffffffffffff`), which is the largest possible value that can be supplied to `TOP n`.

Also rather hacky, but would entirely resolve the issue of Azure Data Warehouse not supporting `TOP [...] PERCENT`. @krlmlr what do you think?

A theoretical better solution, I think, would require changing dbplyr so that it never puts `ORDER BY` in a subquery in the first place.

Calling `arrange` twice currently results in a subquery, like this:

```r
mf <- lazy_frame(x = c(2, 1, 2), y = c(3, 3, 4), con = simulate_mssql())
mf %>% arrange(x) %>% arrange(y)
#> <SQL>
#> SELECT *
#> FROM (SELECT TOP 100 PERCENT *
#> FROM `df`
#> ORDER BY `x`) `dbplyr_001`
#> ORDER BY `y`
```
Instead dbplyr would need to be smart enough to walk the query syntax tree, find any `order_by` operations, and combine them into a single `ORDER BY` in the outer query, emitting SQL like this:
```r
#> <SQL>
#> SELECT *
#> FROM `df`
#> ORDER BY `y`, `x`
```

krlmlr:
I'm not sure we can fully support the dplyr semantics of `arrange()` here -- specifically we can't guarantee stability of sort order.

#276 has some ideas, can we continue the discussion there?

alexkyllo:
Yes, it's an open question of how to correctly translate `arrange()` to SQL in a semantic-preserving and compatible manner. Meanwhile, is it OK to go ahead and merge this PR, which is basically just a one-line bug fix?

pbahr:
I would love to see this PR approved and merged into the package. In the meantime, I have to install the package from this branch to be able to interact with Azure SQL DW. Thanks to @alexkyllo !

krlmlr:
Do we still need `TOP 92xxxxx` instead of `TOP 100 PERCENT`?

R/backend-mssql.R
```diff
@@ -22,7 +22,7 @@
       # e.g: SELECT TOP 100 * FROM my_table
       assert_that(is.numeric(limit), length(limit) == 1L, limit > 0)
       build_sql("TOP(", as.integer(limit), ") ", con = con)
-    } else if (!is.null(order_by) && bare_identifier_ok) {
+    } else if (has_length(order_by) && bare_identifier_ok) {
```
hadley, line 25: This would be clearer as `length(order_by) > 0`

tests/testthat/test-backend-mssql.R
```diff
@@ -164,9 +164,30 @@ FROM `df`")
...
-    mf %>% mutate(x = -x) %>% arrange(x) %>% mutate(x = -x) %>% sql_render(),
-    sql("SELECT -`x` AS `x`\nFROM (SELECT TOP 100 PERCENT *\nFROM (SELECT TOP 100 PERCENT -`x` AS `x`\nFROM `df`) `dbplyr_001`\nORDER BY `x`) `dbplyr_002`")
+  rendered_query <-
+    mf %>%
+    mutate(x = -x) %>%
+    arrange(x) %>%
+    mutate(x = -x) %>%
+    sql_render()
+
+  expect_match(
```
hadley, line 174: Needs to use `verify_output()`

krlmlr:
@alexkyllo: I can take care of that if you don't have the time.

alexkyllo:
Hi @krlmlr, it's been some time since I've looked at this so just to refresh my memory and confirm understanding: 

- dbplyr's mssql backend intended to issue `SELECT TOP 100 PERCENT` in subqueries that have an `ORDER BY` as a workaround, as described in this comment: https://github.com/tidyverse/dbplyr/blob/ddec0e870f067ad94693fd09cda69fa29df1b6fb/R/backend-mssql.R#L26
- A bug caused by `!is.null(order_by)` returning true even when there was no ORDER BY clause, resulted in the `TOP 100 PERCENT` to always be emitted. https://github.com/tidyverse/dbplyr/blob/ddec0e870f067ad94693fd09cda69fa29df1b6fb/R/backend-mssql.R#L24
- Azure Data Warehouse (still) doesn't support PERCENT here, so the aforementioned bug makes the package wholly unusable for ADW.
- My PR as written fixes that specific bug, but dbplyr would still break for ADW, only when there _is_ actually an `ORDER BY` in the subquery.
- That problem can be resolved by emitting `TOP 9223372036854775807` (64-bit signed integer max) instead of `TOP 100 PERCENT` (and making sure this is clearly commented and documented).

So I will go ahead and add that last change (the `TOP 9223372036854775807` part) to this PR, and also address @hadley's two review comments in the process, does that sound good?

krlmlr:
Thanks. I agree with your analysis.

Sorry for the overlap -- I looked at #373 once again and think we can fix properly and eliminate the need for `TOP 100 PERCENT` altogether.

I'll respond here if it turns out we can't fix #373.

tests/testthat/test-backend-mssql.R
krlmlr: We settled on the following structure:

```r
test_that("...", {
  local_options(dbplyr_table_num = 0)
  verify_output(test_path("sql/<filename>.txt"), {
     ...
  }
}
```

The `local_options()` is important to keep the names of subqueries stable. Can you please update only this test as part of this pull request?

krlmlr:
While we can eliminate most `TOP 100 PERCENT`, it doesn't look like we can eliminate all of them. We need to keep the code as a fallback for a little while, it's fine to fix it here.

I still agree with your analysis, it's fine to use `TOP 9223372036854775807`, with a comment. Apparently this worked in 2011 already: https://stackoverflow.com/a/4971263/946850.

Thanks again!

tests/testthat/test-backend-mssql.R
krlmlr: Done in f1e0f47, can you cherry-pick from there?

tests/testthat/test-backend-mssql.R
```diff
@@ -161,13 +161,12 @@ FROM `df`")
...
-
-  expect_equal(
-    mf %>% mutate(x = -x) %>% arrange(x) %>% mutate(x = -x) %>% sql_render(),
-    sql("SELECT -`x` AS `x`\nFROM (SELECT TOP 100 PERCENT *\nFROM (SELECT TOP 100 PERCENT -`x` AS `x`\nFROM `df`) `dbplyr_001`\nORDER BY `x`) `dbplyr_002`")
-  )
+test_that("ORDER BY in subqueries uses TOP 9223372036854775807 (#337)", {
+  local_options(dbplyr_table_num = 0)
+  verify_output("sql/mssql-order-by.txt", {
+    mf <- lazy_frame(x = 1:3, con = simulate_mssql())
+    mf %>% mutate(x = -x) %>% arrange(x) %>% mutate(x = -x) %>% sql_render()
```
krlmlr, line 168: Let's add another test, this will be fixed by #373.

Do lazy frames still need `sql_render()` in output tests?

```suggestion
    mf %>% mutate(x = -x) %>% arrange(x) %>% mutate(x = -x)

    mf2 <- lazy_frame(x = 3:1, con = simulate_mssql())
    mf %>% arrange(x) %>% left_join(mf2 %>% arrange(x))
```

PR #342, opened 2019-08-05
Adds db_pivot_longer function and tests

edgararuiz-zz:


edgararuiz:
PTAL @hadley 

It looks like the Travis failure is not related to the changes in the PR: https://travis-ci.org/tidyverse/dbplyr/jobs/568099285#L3499

hadley:
Would you mind briefly outlining how it works?

edgararuiz:
1. Gets the unique combinations of each non-selected columns (`group_by()` + `summarise()`)
1. Takes the results of step 1, and joins them with the original table, and selecting each "selected" column at a time (`map()` + `inner_join()`)
1. Within each `map()` iteration, it `renames()` the "selected" column "value" (or the value in `values_to`) and creates a column (`mutate()`) with the name of the column, and names that column "names" (or the value on `names_to()`)
1. Uses `union_all()` to paste all of the results, and make sure we send only one query

hadley:
Can you include a small example (just here in the comments) showing the generated SQL?

edgararuiz:
Sure, here's an example with SQLite.  I just realized that the JOIN repeats in the resulting query instead of being a sub-query, not sure if we could help that:

```r
library(DBI)
library(dplyr)
library(dbplyr)
con <- DBI::dbConnect(RSQLite::SQLite(), path = ":dbname:")
copy_to(con, mtcars) %>%
  select(am, mpg, wt) %>%
  db_pivot_longer(-am) %>%
  show_query()
#> <SQL>
#> SELECT `am`, 'mpg' AS `name`, `value`
#> FROM (SELECT `am`, `mpg` AS `value`
#> FROM (SELECT `LHS`.`am` AS `am`, `RHS`.`mpg` AS `mpg`, `RHS`.`wt` AS `wt`
#> FROM (SELECT `am`
#> FROM (SELECT `am`, `mpg`, `wt`
#> FROM `mtcars`)
#> GROUP BY `am`) AS `LHS`
#> INNER JOIN (SELECT `am`, `mpg`, `wt`
#> FROM `mtcars`) AS `RHS`
#> ON (`LHS`.`am` = `RHS`.`am`)
#> ))
#> UNION ALL
#> SELECT `am`, 'wt' AS `name`, `value`
#> FROM (SELECT `am`, `wt` AS `value`
#> FROM (SELECT `LHS`.`am` AS `am`, `RHS`.`mpg` AS `mpg`, `RHS`.`wt` AS `wt`
#> FROM (SELECT `am`
#> FROM (SELECT `am`, `mpg`, `wt`
#> FROM `mtcars`)
#> GROUP BY `am`) AS `LHS`
#> INNER JOIN (SELECT `am`, `mpg`, `wt`
#> FROM `mtcars`) AS `RHS`
#> ON (`LHS`.`am` = `RHS`.`am`)
#> ))
dbDisconnect(con)
```

PR #356, opened 2019-09-06
Remove order after compute()

krlmlr:
The returned object refers to a database table which doesn't have an inherent order, even if it may apper this way. (In particular, MSSQL seems to store in arbitrary order.) If we want to retain the order, we need to `arrange()`.

In the reprex below (taken with current dev version f3529c3f), the `"Ordered by: a"` is misleading.

``` r
library(tidyverse)
library(dbplyr)

memdb_frame(a = 1) %>%
  arrange(a) %>%
  compute()
#> # Source:     lazy query [?? x 1]
#> # Ordered by: a
#>       a
#>   <dbl>
#> 1     1
```

<sup>Created on 2019-09-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


krlmlr:
Superseded by #466.

PR #372, opened 2019-11-01
Respect current window frame in first() et al.

### LINKED ISSUE #366: last() and nth() don't work correctly for SQLite and perhaps other databases
Window frame `window_frame()` is currently ignored for `last()` and `nth()`:

``` r
library(tidyverse)
library(dbplyr)
tbl <- memdb_frame(a = 4:1, g = rep(1:2, 2))

print.tbl_dbi <- function(x, ...) {
  message(sql_render(x))
  NextMethod()
}

# first() seems to work.
tbl %>%
  arrange(a) %>%
  group_by(g) %>%
  mutate(l = first(a))
#> SELECT `a`, `g`, FIRST_VALUE(`a`) OVER (PARTITION BY `g` ORDER BY `a`) AS `l`
#> FROM (SELECT *
#> FROM `dbplyr_001`
#> ORDER BY `a`)
#> # Source:     lazy query [?? x 3]
#> # Groups:     g
#> # Ordered by: a
#>       a     g     l
#>   <int> <int> <int>
#> 1     2     1     2
#> 2     4     1     2
#> 3     1     2     1
#> 4     3     2     1

# last() doesn't:
tbl %>%
  group_by(g) %>%
  arrange(a) %>%
  mutate(l = last(a))
#> SELECT `a`, `g`, LAST_VALUE(`a`) OVER (PARTITION BY `g` ORDER BY `a`) AS `l`
#> FROM (SELECT *
#> FROM `dbplyr_001`
#> ORDER BY `a`)
#> # Source:     lazy query [?? x 3]
#> # Groups:     g
#> # Ordered by: a
#>       a     g     l
#>   <int> <int> <int>
#> 1     2     1     2
#> 2     4     1     4
#> 3     1     2     1
#> 4     3     2     3

# We need "ROWS BETWEEN CURRENT AND UNBOUNDED FOLLOWING":
tbl %>%
  mutate(l = sql(!!win_over(sql("LAST_VALUE(a)"), "g", "a", c(0, Inf), con = tbl$src$con)))
#> SELECT `a`, `g`, LAST_VALUE(a) OVER (PARTITION BY `g` ORDER BY `a` ROWS BETWEEN CURRENT ROW AND UNBOUNDED FOLLOWING) AS `l`
#> FROM `dbplyr_001`
#> # Source:   lazy query [?? x 3]
#> # Database: sqlite 3.29.0 [:memory:]
#>       a     g     l
#>   <int> <int> <int>
#> 1     2     1     4
#> 2     4     1     4
#> 3     1     2     3
#> 4     3     2     3
```

<sup>Created on 2019-10-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

@hannesmuehleisen: Do you know if the default range specification for window functions is standardized across databases? [SQLite](https://www.sqlite.org/windowfunctions.html) has:

> `RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW EXCLUDE NO OTHERS` 

which explains why the example fails, but does it necessarily fail for other databases? Should we always pass an unbounded range to mimic dplyr semantics?

krlmlr:
Affected functions:

- `first()`, `last()`, `nth()`
- `lead()` and `lag()`

Closes #366.


krlmlr:
I think `first()`, `last()` and `nth()` need a `window_frame` argument. Will adapt.

ramnathv:
@krlmlr Any idea when this would get merged? Let me know if I can help in any way.

hadley:

Issue #379, opened 2019-11-30
**as.double** casts to integer in MSSQL


DavidPatShuiFong:
Currently in `dbplyr 1.4.2`, both **as.numeric** and **as.double** is translated to **CAST("x" as NUMERIC)** in MSSQL.

This doesn't give the intended effect for string representations of floating point numbers.

In the example below, the *wt* column of *mtcars* is converted to character, and stored in a temporary MSSQL table. When converted back to a numeric with either **as.numeric** or **as.double**, the column is changed to an integer.

```r
> con <- pool::poolCheckout(a$emr_db$conn()) # or whatever else is required to get a DB connection to a MSSQL database
> DBI::dbWriteTable(con, "##mtcars", mtcars %>% dplyr::mutate(wt = as.character(wt)))
> con %>% dplyr::tbl("##mtcars")

# Source:   table<##mtcars> [?? x 11]
# Database: Microsoft SQL Server 12.00.2000[bpsrawdata@DAVIDLENOVOPC\BPSINSTANCE/BPSSAMPLES]
     mpg   cyl  disp    hp  drat wt     qsec    vs    am  gear  carb
   <dbl> <dbl> <dbl> <dbl> <dbl> <chr> <dbl> <dbl> <dbl> <dbl> <dbl>
 1  21       6  160    110  3.9  2.62   16.5     0     1     4     4
 2  21       6  160    110  3.9  2.875  17.0     0     1     4     4
 3  22.8     4  108     93  3.85 2.32   18.6     1     1     4     1
 4  21.4     6  258    110  3.08 3.215  19.4     1     0     3     1
 5  18.7     8  360    175  3.15 3.44   17.0     0     0     3     2
 6  18.1     6  225    105  2.76 3.46   20.2     1     0     3     1
 7  14.3     8  360    245  3.21 3.57   15.8     0     0     3     4
 8  24.4     4  147.    62  3.69 3.19   20       1     0     4     2
 9  22.8     4  141.    95  3.92 3.15   22.9     1     0     4     2
10  19.2     6  168.   123  3.92 3.44   18.3     1     0     4     4
# ... with more rows

> con %>% dplyr::tbl("##mtcars") %>% dplyr::mutate(wt = as.numeric(wt))
# Source:   lazy query [?? x 11]
# Database: Microsoft SQL Server 12.00.2000[bpsrawdata@DAVIDLENOVOPC\BPSINSTANCE/BPSSAMPLES]
[...]

PR #387, opened 2019-12-14
cast "as.double" and "as.numeric" to FLOAT instead of NUMERIC for MSSQL 

DavidPatShuiFong:
casting as NUMERIC in MSSQL results in an integer being returned, even if the character string is representing a floating point number. casting as FLOAT results in a float being returned.
addresses issue https://github.com/tidyverse/dbplyr/issues/379


DavidPatShuiFong:
sorry, this contains a file change which is not part of this issues

Issue #376, opened 2019-11-27
add SQL translation for weighted.mean


saadaslam:
I often encounter the need to perform weighted average calculations. R has a neat functionality to perform this with `weighted.mean`. It's even more useful when there are missing values, in which I can provide `na.rm = TRUE`. 

I think it's worthwhile providing a `weighted.mean` translation for dbplyr. Mainly because, the method in which we produce weighted average calculations is verbose:

```r
library(tidyverse)
library(babynames)
library(dbplyr)
baby_db <- tbl_memdb(babynames)

## R and dplyr
babynames %>% 
  summarize(wavg_prop = weighted.mean(prop, n))
## dbplyr
baby_db %>% 
  summarize(wavg_prop = sum(prop * n) / sum(n))
```
This becomes a little more verbose and repetitive when missing values are apparent:

```r
set.seed(1337)
babynames_missing <- babynames
babynames_missing[sample(1:nrow(babynames), 1e5, replace = F), "prop"] <- NA
baby_missing_db <- tbl_memdb(babynames_missing)

babynames_missing %>% 
  summarize(wavg_prop = weighted.mean(prop, n, na.rm = TRUE))
baby_missing_db %>% 
  summarize(
    wavg_prop = sum(prop * n) / sum(if_else(! is.na(prop), n, 0)),
    wavg_prop2 = sum(prop * n) / sum(n)
  )
```

Note the difference in between `wavg_prop` and `wavg_prop2` when missing values are apparent. 

I wrote code in `backend-.R` to supply a translator. If you think it's worthwhile, I can add that on for a pull request.

AdeelK93:
This is great stuff! Would love to see it in dbplyr

hadley:
A PR would be great! 

I think you also need to check that `n` isn't missing?

saadaslam:
Yes, I would need to add a condition for the weighting variable to be missing, as well. Thanks for pointing it out.

ahcyip:
I'm getting "operator does not exist" "no operator matches given name and argument types" when I try to do the manual weighted average as above. `show_query` seems to produce the right query, but `collect` fails.

Comparing to `baby_db %>% summarize(wavg_prop = sum(prop * n) / sum(n))`, I see that my names are all quoted with " - could that be causing a problem?

I've posted on StackOverflow to see if anyone else can help: https://stackoverflow.com/questions/65545756/r-dbplyr-translation-produces-operator-does-not-exist-error

```
[...]

PR #389, opened 2019-12-15
SQL translation for `weighted.mean`

saadaslam:
See issue #376 

Creates a translation for the R function `weighted.mean`.


hadley:
Thanks for having a go at this, but as discussed in the linked issue, I'm worried that this will work often enough to give people false confidence. I really appreciate the effort you've put into this, and I hope you'll contribute again in the future 😄

Issue #384, opened 2019-12-13
Non standard handling of iso-date strings in PrestoDB/AWS Athena


OssiLehtinen:
Most databases are able to interpret iso-date strings as dates in, e.g., "where" clauses, such as

`select * from table where date_column = '2019-01-01'`

This is not the case with PrestoDB/AWS Athena, however. Instead, the strings need to be cast as dates first like:
`select * from table where date_column = date('2019-01-01')`,
`select * from table where date_column = date'2019-01-01'` or
`select * from table where date_column = cast('2019-01-01' as date)`

However, when a date variable is used as a parameter id dplyr/dbplyr, is is translated to a iso-date string in the generated sql, so a command such as:

```r
today <- Sys.Date()
tbl(AthenaCon, "table") %>% filter(date_column <= today)
```

produces and error, as the generated sql does not have the explicit cast to date of the iso-date string. One needs to manually cast the date variable as a date as in
`... %>% filter(date_column <= as.Date(today))`, but this feels unnecessary, as R knows today is a date already.

A DBI backend for AWS Athena is being developed at https://github.com/DyfanJones/RAthena, and I'm asking for some guidance on how to address this issue there.

My question is: is there a method for overriding the default behavior of dbplyr, where dates are translated to quoted strings and instead have them translated to, e.g., "date'2019-01-01'"?

I'm sorry if this seems like a trivial question, but I'm drawing a blank from reading the database specific source files and packages. All I can come up with is to use assignInNameSpace to replace the escape.Date -method, but this seems pretty brutal.

Thanks for any hints!

hadley:
It's not super elegant, but you can use the technique below to work around the problem in the interim:

``` r
library(dbplyr)
library(dplyr, warn.conflicts = FALSE)

athena_date <- function(x) sql(paste0("date('", x, "')"))

today <- Sys.Date()

tbl_lazy(mtcars) %>% 
  filter(date_column <= !!athena_date(today))
#> <SQL>
#> SELECT *
#> FROM `df`
#> WHERE (`date_column` <= date('2019-12-13'))
```

<sup>Created on 2019-12-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Otherwise, you have to wait for a custom backend to provide the correct escaping behaviour.

OssiLehtinen:
Thanks for the response!

[...]

PR #391, opened 2019-12-16
Add generics and implementations for custom Date and POSIXt escaping

OssiLehtinen:
These modifications address the issue described in https://github.com/tidyverse/dbplyr/issues/384


OssiLehtinen:
Travis CI was not too happy about my pull request. I'm not familiar with Travis CI so not sure how to proceed.

OssiLehtinen:
I added export of sql_escape_string.

This would enable optionally setting, _e.g._,
```r
sql_escape_string.AthenaConnection <- function(con, x) {
  if(try(as.Date(x), silent=T)  == x) {
    paste0('date ', dbQuoteString(con, x))
  } else {
    dbQuoteString(con, x)
  }
}
```
meaning a call, such as `tbl(con, "calendar") %>% filter(date_col == "2019-01-01")`, would work, as the string gets identified as an isodate and the 'date' prefix gets added.
This is obviously very much subjective, but in pretty much all of our ad-hoc analysis use cases we need to do filtering of data by date, so any streamlining of this would go a long way.

However, I'm uncertain whether automatic detection of isodates should be the default behaviour. This change would enable doing this if wanted.

OssiLehtinen:
A bit of a mess, but I think something like this would actually be needed to have date vectors work out and have only iso-dates cast to dates. Automatically casting something like "1" to a date is not a good thing.

```r
sql_escape_string.AthenaConnection <- function(con, x) {
  all_dates <- all(try(as.Date(x, tryFormats = "%Y-%m-%d"), silent=T) == x)
  if(all_dates & !is.na(all_dates)) {
    paste0('date ', DBI::dbQuoteString(con, x))
  } else {
    DBI::dbQuoteString(con, x)
  }
}
```

In any case, I don't think this should be the default behaviour of dbplyr, but exporting the generic would allow a user to shoot him/herself in the foot, if that is what he/she wishes to do.

OssiLehtinen:
One more note:

if one would define sql_escape_string.AthenaConnection as described above, the escape.Date definition would not be necessary, as currently the dates get cast to strings first and the sql_escape_string function is called on those, which would then detect them as dates. 

However, still I don't think the sql_escape_string should be defined like that as default, so having the Athena-specific escape.Date (and escape.POSIXt) would be a good thing imho.

p.s. If I'm not mixing up things (still learning about s3-classes etc), exporting sql_escape_string is actually not necessary to be able to define the Athena specific version, right?

R/escape.R
```diff
@@ -69,14 +69,13 @@ escape.factor <- function(x, parens = NA, collapse = ", ", con = NULL) {
 
 #' @export
 escape.Date <- function(x, parens = NA, collapse = ", ", con = NULL) {
-  x <- as.character(x)
-  escape.character(x, parens = parens, collapse = collapse, con = con)
+  x = as.character(x)
```
hadley, line 72: I think you should delete this line; `sql_escape_date()` should be in charge of converting from a date to a character

R/escape.R
```diff
@@ -229,13 +228,37 @@ sql_escape_logical <- function(con, x) {
   UseMethod("sql_escape_logical")
 }
 
+#' @keywords internal
+#' @export
+sql_escape_Date <- function(con, x) {
```
hadley, line 233: I'd prefer to call these `sql_escape_date()` and `sql_escape_datetime()`.

hadley:
(Also since it's a been a while since you submitted this PR, please let me know if you're not interested, and I'll finish it)

R/escape.R
```diff
@@ -229,13 +228,37 @@ sql_escape_logical <- function(con, x) {
...
 # DBIConnection methods --------------------------------------------------------
 
 #' @export
 sql_escape_string.DBIConnection <- function(con, x) {
   dbQuoteString(con, x)
 }
 
+#' @export
+sql_escape_Date.DBIConnection <- function(con, x) {
+  dbQuoteString(con, x)
```
krlmlr, line 253: DBI has this, DBI backends can override.

```suggestion
  dbQuoteLiteral(con, x)
```

R/escape.R
```diff
@@ -229,13 +228,37 @@ sql_escape_logical <- function(con, x) {
...
+}
+
+#' @export
+sql_escape_POSIXt.DBIConnection <- function(con, x) {
+  x <- strftime(x, "%Y-%m-%dT%H:%M:%OSZ", tz = "UTC")
+  dbQuoteString(con, x)
```
krlmlr, line 259: Should this be `%OS6`? DBI needs to be fixed here -- I think passing the time zone might be correct here.

```suggestion
  dbQuoteLiteral(con, x)
```

R/escape.R
OssiLehtinen: Moved to coverting bit to sql_escape_date.

R/escape.R
OssiLehtinen: Renamed the functions.

R/escape.R
OssiLehtinen: I'm sorry, could you explain a bit what would be the benefit of this change? dbQuoteLiteral seems to format a date as something like '2020-04-15 UTC' which does not work at least in Athena (don't have other databases up right now to test against). So if I would use dbQuoteLiteral, I would still need to first cast the date as character.

Surely I'm just missing the point here, so please help me out :)

R/escape.R
OssiLehtinen: This specific formatting is what is in place in dbplyr at the moment for POSIXt, so I wouldn't want to touch that.

OssiLehtinen:
I made changes to the code according to the recommendations. I hope I got it right. Could you @hadley take a look at the news bullet, as I'm a bit uncertain about the terminology? Thanks!

hadley:
Thanks @OssiLehtinen — I finished it by updating your branch with the latest changes on master, tweaking the news bullet, and making a couple of small documentation and test fixes. Thanks for all your work on this!

OssiLehtinen:
> Thanks @OssiLehtinen — I finished it by updating your branch with the latest changes on master, tweaking the news bullet, and making a couple of small documentation and test fixes. Thanks for all your work on this!

I'm very happy to be able to contribute even a little bit to this great stack of tools!

PR #392, opened 2019-12-17
postgres backend: str_detect, str_replace and str_replace_all

### LINKED ISSUE #362: str_replace_all mimics str_replace
The current implementation of `str_replace_all` which was introduced in PR #131 for postgres backends has no global parameter, thus replacing only the first occurence of any match in a string:

File R/backend-postgres.R, line 58ff:

```r
str_replace_all = function(string, pattern, replacement){
        sql_expr(regexp_replace(!!string, !!pattern, !!replacement))
      }
```

To my mind, the correct implementation would be
```r
str_replace_all = function(string, pattern, replacement){
        sql_expr(regexp_replace(!!string, !!pattern, !!replacement, 'g'))
    }
```

Also, the fix of str_detect has disappeared from master and it is not possible to grep for regular expressions on Postgres databases at the moment: Line 55 in beforementioned file quotes

```r
      str_detect  = function(string, pattern) {
        sql_expr(strpos(!!string, !!pattern) > 0L)
      }
```

instead of the line that was implemented Pull request #131:

```r
str_detect = sql_infix("~"),
```

Thanks for the great effort and package!
sandro

shosaco:
corrected str_detect so it is possible to use regular expressions, introduced str_replace and corrected str_replace_all with 'g' parameter

fixes #362

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dbplyr (development version)
 
+* `str_detect()`, `str_replace()` and `str_replace_all()` are now correct for postgres backends (@shosaco, #362).
+
 # dbplyr 1.4.2
 
 * Fix bug when partially evaluating unquoting quosure containing a single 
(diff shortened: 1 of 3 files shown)


hadley:
Thanks! Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

shosaco:
done

hadley:
Thanks!

PR #397, opened 2020-01-22
Fix bracket typo in error message

alistaire47:
Just replacing a paren with a bracket so the code in the error is syntactic.


man/memdb_frame.Rd
```diff
@@ -16,7 +16,7 @@ src_memdb()
 \item{...}{A set of name-value pairs. Arguments are evaluated sequentially,
 so you can refer to previously created elements. These arguments are
 processed with \code{\link[rlang:quos]{rlang::quos()}} and support unquote via \code{\link{!!}} and
-unquote-splice via \code{\link{!!!}}. Use \verb{:=} to create columns that start with a dot.}
+unquote-splice via \code{\link{!!!}}. Use \code{:=} to create columns that start with a dot.}
```
alistaire47: Rebuilt the docs out of habit, and I'm not sure why this happened...can revert it if you like

hadley: It looks like you need to upgrade roxygen2 (but don't worry about it for this PR, I'll take care of it)

hadley:
Thanks!

PR #390, opened 2019-12-15
Move in dplyr generics

hadley:
Fixes #385 

Currently this generates a rather large set of conflicts:

``` r
library(dbplyr)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:dbplyr':
#> 
#>     db_analyze, db_begin, db_commit, db_create_index,
#>     db_create_indexes, db_create_table, db_data_type, db_desc,
#>     db_drop_table, db_explain, db_has_table, db_insert_into,
#>     db_list_tables, db_query_fields, db_query_rows, db_rollback,
#>     db_save_query, db_write_table, ident, sql, sql_escape_ident,
#>     sql_escape_string, sql_join, sql_select, sql_semi_join, sql_set_op,
#>     sql_subquery, sql_translate_env
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
```

<sup>Created on 2019-12-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #407, opened 2020-03-30
Avoid import overwrite message on load

krlmlr:
This removes the annoying message when running dplyr tests.


hadley:
Thanks!

hadley:
Oh bummer, this fails on 3.2.0 — https://github.com/tidyverse/dbplyr/pull/422/checks?check_run_id=587249783#step:9:22

hadley:
I'm going to revert; instead I'll eventually switch to the selective approach used in #390

PR #424, opened 2020-04-15
Support collect(n = 0) and head(0)

### LINKED ISSUE #415: Support collect(n = 0)
This is useful for getting column metadata, and seems to be supported across databases:

- https://stackoverflow.com/q/6033957/946850
- https://stackoverflow.com/q/37495497/946850

If not supported, the database will yell at us. The only risk I see is a misbehaving DBI driver, we could subclass.

krlmlr:
Not sure how to test.

Closes #415.


R/backend-access.R
```diff
@@ -22,7 +22,7 @@ sql_select.ACCESS <- function(con, select, from,
     # TOP is expected after DISTINCT and not at the end of the query
     # e.g: SELECT TOP 100 * FROM my_table
     if (!is.null(limit) && !identical(limit, Inf)) {
-      assert_that(is.numeric(limit), length(limit) == 1L, limit > 0)
+      assert_that(is.numeric(limit), length(limit) == 1L, limit >= 0)
```
hadley, line 25: I wonder if we should remove these checks in favour of a single check in `head()`?

R/verb-compute.R
```diff
@@ -94,7 +94,7 @@ db_compute.DBIConnection <- function(con,
 #' @param warn_incomplete Warn if `n` is less than the number of result rows?
 #' @export
 collect.tbl_sql <- function(x, ..., n = Inf, warn_incomplete = TRUE) {
-  assert_that(length(n) == 1, n > 0L)
+  assert_that(length(n) == 1, n >= 0L)
```
hadley, line 97: Maybe convert the if statement to `identical(n, Inf)` then `head()` can take care of the validation?

PR #427, opened 2020-04-15
Use new database backends

### LINKED ISSUE #423: Switch default packages to MariaDB and RPostgres

hadley:
Fixes #423


hadley:
Thanks — I will eliminate withr since we don't need it

hadley:
(Thanks for spotting that open connection; it had been bothering me for a while but I never had the heart to run down the source)

PR #433, opened 2020-04-23
Escape raw vectors and support list of raw vectors

okhoma:
Hi,

Please, consider my pull request to support escape of raw vectors and lists of raw vectors. 

The motivation is to be able to pass UTF-8 strings into dplyr `filter` method using a workaround described in https://github.com/r-dbi/DBI/issues/215#issuecomment-356376133. Here, UTF-8 character vectors are converted to list of UTF-16 raw vectors and then the data is properly written into the database. However, when I attempt to filter the table passing raw vectors or list of raw vectors in the same manner, I am getting an error `Error in UseMethod("escape") :  no applicable method for 'escape' applied to an object of class "raw"`.


R/escape.R
```diff
@@ -214,7 +218,16 @@ sql_quote <- function(x, quote) {
...
-
+#' Helper function to convert raw vector to hex string
+#'
+#' @param x Raw vector
+#' @export
+#' @keywords internal
+#' @examples
+#' sql_raw_to_hex(charToRaw("abc"))
+sql_raw_to_hex <- function(x) {
+  paste(sprintf("%02X", as.integer(x)), collapse = "")
```
krlmlr, line 229: Would that work?

```suggestion
  paste(format(x), collapse = "")
```

R/escape.R
```diff
@@ -214,7 +218,16 @@ sql_quote <- function(x, quote) {
...
 
-
+#' Helper function to convert raw vector to hex string
+#'
+#' @param x Raw vector
+#' @export
+#' @keywords internal
+#' @examples
+#' sql_raw_to_hex(charToRaw("abc"))
+sql_raw_to_hex <- function(x) {
```
hadley, line 228: Does this need to be exported?

R/escape.R
```diff
@@ -273,4 +293,6 @@ sql_escape_logical.DBIConnection <- function(con, x) {
   y
 }
 
-
+sql_escape_raw.DBIConnection <- function(con, x) {
```
hadley, line 296: This needs to be exported

R/sql-expr.R
```diff
@@ -77,3 +77,8 @@ replace_expr <- function(x, con) {
   }
 
 }
+
+# copied from DBI
+is_raw_list <- function(x) {
```
hadley, line 82: Shouldn't this be using [blob](https://github.com/tidyverse/blob)?

R/escape.R
hadley: Can you also add a comment with link to some docs about this sql feature?

R/escape.R
```diff
@@ -108,6 +108,11 @@ escape.integer64 <- function(x, parens = NA, collapse = ", ", con = NULL) {
   sql_vector(x, parens, collapse, con = con)
 }
 
+#' @export
+escape.raw <- function(x, parens = NA, collapse = ", ", con = NULL) {
```
hadley, line 112: This feels a bit weird to me because x is going to be a vector, but this function will always return a single value. That makes it different to the other methods.

R/escape.R
okhoma: Removed this function altogether as too trivial

R/escape.R
okhoma: Added @export and comments

R/sql-expr.R
okhoma: Updated to use blob instead of list of raw.

R/escape.R
okhoma: It does, but raw is actually different from other types.

R/escape.R
```diff
@@ -108,6 +108,17 @@ escape.integer64 <- function(x, parens = NA, collapse = ", ", con = NULL) {
...
+  sql_vector(sql_escape_raw(con, x), parens, collapse, con = con)
+}
+
+#' @export
+escape.blob <- function(x, parens = TRUE, collapse = ", ", con = NULL) {
+  pieces <- vapply(x, escape, character(1), con = con)
```
krlmlr, line 118: Do we really need `escape.raw()`? Could we call into `sql_escape_raw()` directly instead?

```suggestion
  pieces <- vapply(x, sql_escape_raw, character(1), con = con)
```

R/backend-mssql.R
```diff
@@ -177,6 +177,13 @@
 
   )}
 
+#' @export
+`sql_escape_raw.Microsoft SQL Server` <- function(con, x) {
+  # SQL Server binary constants should be prefixed with 0x
+  # https://docs.microsoft.com/en-us/sql/t-sql/data-types/constants-transact-sql?view=sql-server-ver15#binary-constants
+  paste0(c("0x", format(x)), collapse = "")
```
krlmlr, line 184: I benchmarked performance of some variants and a C/C++ implementation in https://github.com/r-lib/rlang/issues/978. No need to act here.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # dbplyr (development version)
 
+* `escape()` now support `raw` and `blob` vectors using new `sql_escape_raw()` 
+  generic. It eables using [blob](https://blob.tidyverse.org/) variables in
```
krlmlr, line 4:
```suggestion
* `escape()` now support `blob` vectors using new `sql_escape_raw()` 
  generic. It enables using [blob](https://blob.tidyverse.org/) variables in 
```

krlmlr:
I also sent a pull request to remove the extraneous magrittr import: https://github.com/okhoma/dbplyr/pull/1/files.

R/escape.R
```diff
@@ -192,7 +203,6 @@ names_to_as <- function(x, names = names2(x), con = NULL) {
 #' If the quote character is present in the string, it will be doubled.
 #' `NA`s will be replaced with NULL.
 #'
-#' @export
```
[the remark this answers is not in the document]
okhoma: @export was used twice in this comment.

R/escape.R
okhoma: Thanks, removed escape.raw

hadley:
Thanks!

PR #435, opened 2020-04-28
Use immediate = TRUE

krlmlr:
for statements. Important on SQL Server, should not hurt other databases.


krlmlr:
All tests with MSSQL succeed locally when I add the test source. 

Tests fail without development version of {odbc}. For FreeTDS we also need https://github.com/r-dbi/odbc/pull/360.

krlmlr:
Tests are now run on GitHub Actions too.

DESCRIPTION
```diff
@@ -41,13 +41,16 @@ Suggests:
...
     nycflights13,
+    odbc (>= 1.2.2.9000),
     RMariaDB (>= 1.0.2),
     rmarkdown,
     RPostgres (>= 1.1.3),
     RSQLite (>= 2.1.0),
     testthat (>= 2.0.0)
 VignetteBuilder: 
     knitr
+Remotes:
```
hadley, line 52: Can you not use remotes, and skip if the version isn't high enough? That way I don't have to worry about updating it when I release to CRAN?

krlmlr:
I'll also add explicit mssql tests.

Issue #373, opened 2019-11-01
Retaining too optimistic order information when manipulating after arrange()


krlmlr:
It seems that `op_sort()` is a tad too optimistic in some cases.

Related: #338 #337. I think `arrange()` should only have an effect if it's the last verb in a pipe (or can be bubbled down safely), or to define the window order for window functions.

``` r
library(tidyverse)
library(dbplyr)

tbl <- memdb_frame(a = 1:3, b = 3:1)
tbl
#> # Source:   table<dbplyr_001> [?? x 2]
#> # Database: sqlite 3.29.0 [:memory:]
#>       a     b
#>   <int> <int>
#> 1     1     3
#> 2     2     2
#> 3     3     1

tbl_a <- tbl %>% arrange(a)
tbl_a
#> # Source:     table<dbplyr_001> [?? x 2]
#> # Database:   sqlite 3.29.0 [:memory:]
#> # Ordered by: a
#>       a     b
#>   <int> <int>
#> 1     1     3
#> 2     2     2
#> 3     3     1

# Sort order wrong, the database won't guarantee lexical ordering
tbl_ab <- tbl %>% arrange(a) %>% arrange(b)
tbl_ab %>% sql_render()
#> <SQL> SELECT *
#> FROM (SELECT *
#> FROM `dbplyr_001`
#> ORDER BY `a`)
#> ORDER BY `b`

# a no longer exists, confusing
tbl_aba <- tbl %>% arrange(a) %>% select(-a)
tbl_aba
#> # Source:     lazy query [?? x 1]
#> # Database:   sqlite 3.29.0 [:memory:]
#> # Ordered by: a
#>       b
#>   <int>
#> 1     3
#> 2     2
#> 3     1

# How can I un-arrange?
tbl_aa <- tbl_aba %>% arrange(a) %>% arrange()
tbl_aa
#> # Source:     lazy query [?? x 1]
#> # Database:   sqlite 3.29.0 [:memory:]
#> # Ordered by: a, a
#>       b
#>   <int>
#> 1     3
#> 2     2
#> 3     1

# The ordering leaks to the lag() call
tbl_ac <- tbl_aba %>% collapse() %>% mutate(c = lag(b))
tbl_ac
#> Error: no such column: a
tbl_ac %>% sql_render()
#> <SQL> SELECT `b`, LAG(`b`, 1, NULL) OVER (ORDER BY `a`) AS `c`
#> FROM (SELECT `b`
#> FROM (SELECT *
[...]

PR #459, opened 2020-05-19
Refactorings

krlmlr:
as preparation for #373. The commits are self-contained, the third builds upon the second.


tests/testthat/test-backend-mssql.R
```diff
@@ -200,3 +89,52 @@ test_that("custom lubridate functions translated correctly", {
   expect_equal(trans(quarter(x, with_year = TRUE)), sql("(DATENAME(YEAR, `x`) + '.' + DATENAME(QUARTER, `x`))"))
   expect_error(trans(quarter(x, fiscal_start = 5)))
 })
+
+verify_lazy_output("sql/mssql.sql", {
```
hadley, line 93: I think these should just be .txt files since they're not really SQL

tests/testthat/helper-lazy.R
```diff
@@ -0,0 +1,3 @@
+verify_lazy_output <- function(...) {
```
hadley, line 1: Rather than creating a new function, I would recommend using `verify_output()` + `local_options(dplyr_table_num = 0)`

krlmlr: `local_options()` doesn't work as expected in `verify_output()`, the option, the helper seems to be the easiest way. Do you prefer a different name?

``` r
testthat::verify_output("foo.txt", {
  rlang::local_options(foo = 1)
  getOption("foo")
})
#> Warning: Creating reference output

cat(readLines("foo.txt"), sep = "\n")
#> > rlang::local_options(foo = 1)
#> > getOption("foo")
#> NULL
```

<sup>Created on 2020-05-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley: The `local_options()` would go outside the `verify_output()`:

``` r
local({
  rlang::local_options(foo = 1)
  testthat::verify_output("foo.txt", {
    getOption("foo")
  })
})
#> Warning: Creating reference output
cat(readLines("foo.txt"), sep = "\n")
#> > getOption("foo")
#> [1] 1
```

<sup>Created on 2020-05-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
Done, in a more compact way. I don't understand how this is better than defining a helper function, we'll be using this in ~20 places eventually.

tests/testthat/test-backend-mssql.R
```diff
@@ -200,3 +89,52 @@ test_that("custom lubridate functions translated correctly", {
   expect_equal(trans(quarter(x, with_year = TRUE)), sql("(DATENAME(YEAR, `x`) + '.' + DATENAME(QUARTER, `x`))"))
   expect_error(trans(quarter(x, fiscal_start = 5)))
 })
+
+with_options(dbplyr_table_num = 0, verify_output("sql/mssql.txt", {
```
hadley, line 93: The structure of these should be:

```
test_that("...", {
  local_options(dbplyr_table_num = 0)
  verify_output(test_path("path"), {
     ...
  }
}
```

That way each test follows the same form, works interactively and from `devtools::test()`, and is easily understood at a glance. 

Copy and paste is much more acceptable in tests because it's more important than each test be independent and easily understood, than have the smallest amount of code possible.

tests/testthat/test-backend-mssql.R
```diff
@@ -200,3 +89,52 @@ test_that("custom lubridate functions translated correctly", {
...
+  "# filter and mutate translate is.na correctly"
+  mf %>% head()
+
+  mf %>% mutate(z = is.na(x))
+
+  mf %>% mutate(z = !is.na(x))
+
+  mf %>% filter(is.na(x))
+
+  "Invalid, can we reliably detect that the return value is logical?"
```
hadley, line 105: These should stay as separate tests; no need to cram unrelated tests into a single `verify_output()`.

tests/testthat/test-query-join.R
```diff
@@ -16,7 +16,7 @@ test_that("generated sql doesn't change unexpectedly", {
     right = right_join(lf, lf),
     full = full_join(lf, lf)
   )
-  expect_known_output(print(reg), test_path("sql/join.sql"))
+  expect_known_output(print(reg), test_path("sql/join.txt"))
```
hadley, line 19: I think you should leave these as is because they are not related to the current PR. 

When you're remodelling a house, you only need to bring "up to code" the bits of the house that you are touching. The same applies to refactoring.

krlmlr:
Let's split this PR.

PR #466, opened 2020-05-20
arrange() and window_order() override previously set order

krlmlr:
with a warning.

This is a breaking change with two failure modes:

- For queries of the form `. %>% arrange(...) %>% ... %>% mutate(some_window_function())`, the order no longer propagates to the window function. This should lead to an error or at least a warning.
- Queries like `. %>% arrange(...) %>% arrange(...)` have a different order, with a warning.

The `ORDER BY` in subqueries is still emitted in any case. This will be fixed in future PRs.

Addresses part of #373.

Closes #465 (includes it).


krlmlr:
Will submit compound PR.

PR #471, opened 2020-05-23
arrange() only effective at end of pipe, window_order() overrides

### LINKED ISSUE #276: Rethink optimization of ORDER BY

### LINKED ISSUE #373: Retaining too optimistic order information when manipulating after arrange()

krlmlr:
- `arrange()` can now only be used at the end of a pipe or followed by `head()`, otherwise it has no effect and raises a warning.
- `window_order()` can be used repeatedly but only affects the verb that follows immediately.
- All other verbs reset the sort order, with warning.
- An empty `arrange()` can be used to undo the effects of the last `arrange()`.
- This change leads to fewer subqueries and fewer redundant `ORDER BY` clauses.
- The documentation for `arrange()` and `window_order()` has been expanded.

Each commit is self-contained and passes tests. Happy to split into separate pull requests if it helps.

Closes #276, closes #373.


R/verb-window.R
```diff
@@ -1,5 +1,24 @@
 #' Override window order and frame
 #'
+#' @description
+#' `window_order()` sets the order for subsequent invocations
+#' of window functions on a lazy table.
+#' The order is active only for the next verb.
```
hadley, line 6: Is that also true for `window_frame()`? 

Maybe it's easier to just leave the `window_order()` persistence as is?

R/verb-window.R
```diff
@@ -51,6 +70,15 @@ sql_build.op_order <- function(op, con, ...) {
 
 # Frame -------------------------------------------------------------------
 
+#' @description
```
hadley, line 73: Can you please put all these docs up with `winder_order()`?

tests/testthat/sql/arrange-join.txt
```diff
@@ -5,10 +5,13 @@
 Message: Joining, by = "a"
 
 <SQL>
+
+Warning: Ignoring sort order.
+Hint: `arrange()` only has an effect if used at the end of a pipe or after `head()`. See `?arrange.tbl_lazy` for details.
```
hadley, line 10: _before_ head?

tests/testthat/sql/arrange-single.txt
```diff
@@ -35,41 +34,51 @@ ORDER BY `a`
 > # complex mutate
 > lf %>% arrange(a) %>% mutate(a = b) %>% arrange(a)
```
hadley, line 35: Why do we have so many cases here? They all seem rather similar.

R/verb-window.R
krlmlr: `window_frame()` is different because it never stacks. I agree we should remove it there too: #475.

krlmlr: If we leave `window_order()` persistence, we need to provide a way to reset it. This also leaves us vulnerable to `. %>% window_order(a) %>% select(-a)`.

I suggest we make a clean cut here, perhaps after the dplyr CRAN release.

hadley: But clearly `window_order()` and `window_frame()` need to persist the status of the other. That's starting to seem rather complicated to me.

krlmlr: I don't mind persisting the status of the order in `window_order()` if we provide a way to reset it. What would be a good API -- `window_order(.add = FALSE)` ?

krlmlr: `window_order()` and `window_frame()` now respect each other, added tests. `window_order()` still overrides the previously set order.

tests/testthat/sql/arrange-single.txt
krlmlr: Reduced.

PR #478, opened 2020-06-07
Allow using blobs with !! and !!! operators in filter()

okhoma:
This is an addition to #433 to be able to use blobs in `filter()` with !! and !!! operators


okhoma:
@krlmlr Could you please, review this pull request as a continuation of #433. The only fail in checks is from #1109, which, I believe, should be fixed with newer roxygen2.

hadley:

hadley:
Thanks!

PR #483, opened 2020-07-04
allow argument na.rm in median() 

### LINKED ISSUE #442: `na.rm` argument in median() for consistency with stats::median()
I want to compute a median in-memory and on the DB with the same code. Because my data may have missing values, I set `na.rm = TRUE` in `median()`. However, this does not work on the database.

``` r
dbplyr::translate_sql(median(x, na.rm = TRUE))
#> Error in median(x, na.rm = TRUE): unused argument (na.rm = TRUE)
median
#> function (x, na.rm = FALSE, ...) 
#> UseMethod("median")
#> <bytecode: 0x7f98e111c070>
#> <environment: namespace:stats>
```

<sup>Created on 2020-05-04 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I expected it to work the same as with `mean`, i.e. 
```r
dbplyr::translate_sql(mean(x, na.rm = TRUE))
#> <SQL> AVG(`x`) OVER ()
```

Don't know what happens with missing values currently and if that generalizes to all databases, but I thought probably the missing values are removed anyways as stated in the above warning for `mean`? 
```r
dbplyr::translate_sql(mean(x))
#> Warning message:
#> Missing values are always removed in SQL.
#> Use `AVG(x, na.rm = TRUE)` to silence this warning
#> This warning is displayed only once per session. 
#> <SQL> AVG(`x`) OVER ()
```

If that's the case, the in-memory version would by default not remove the NAs, the DB version would, and there is no way to set the `na.rm` argument to align them.

Can we add an argument `na.rm` for consistency to make the same code running with the same result on both engines?

lorenzwalthert:
Closes #442 by extending the codebase as suggested there.  Let's see if tests pass...


hadley:
Thanks! Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

lorenzwalthert:
Done.

hadley:
Thanks!

PR #493, opened 2020-09-07
Postgres: Translations for `str_squish`, `str_remove`, `str_remove_all`

shosaco:
Added Postgres translations for these three stringr functions.


hadley:
Thanks!

Issue #380, opened 2019-11-30
use **try_cast** instead of **cast** with Microsoft SQL (2012 and later)


DavidPatShuiFong:
A failure to **cast** (e.g. **as.numeric** or **as.double**, when the string contains a non-numeric character) results in an error:

```r
> library(magrittr)
> x <- mtcars %>% dplyr::mutate(wt = as.character(wt))
> x[1,6] <- "<2.62" # adds a non-numeric character '<' to an element
> DBI::dbWriteTable(con, "##mtcars", x) # where 'con' is an already defined connection to a MSSQL database
> con %>% dplyr::tbl("##mtcars") %>% dplyr::mutate(wt = as.numeric(wt))
Error: <SQL> 'SELECT TOP(11) "mpg", "cyl", "disp", "hp", "drat", CAST("wt" AS NUMERIC) AS "wt", "qsec", "vs", "am", "gear", "carb"
FROM "##mtcars"'
  nanodbc/nanodbc.cpp:1587: 42000: [Microsoft][ODBC SQL Server Driver][SQL Server]Error converting data type varchar to numeric. 
```

For Microsoft SQL (2012 and later) **try_cast** allows a more graceful return of *NA*. In `dbplyr 1.4.2`, this can be done by changing lines 225-232 in `translate-sql-helpers.R` (but this will probably only work for Microsoft SQL 2012 and later):

```r
#' @rdname sql_variant
#' @export
sql_cast <- function(type) {
  type <- sql(type)
  function(x) {
    sql_expr(try_cast(!!x %as% !!type))
  }
}
```

...which results in a more graceful failure (returns *NA*):

```r
> con %>% dplyr::tbl("##mtcars") %>% dplyr::mutate(wt = as.numeric(wt))
# Source:   lazy query [?? x 11]
# Database: Microsoft SQL Server 12.00.2000[bpsrawdata@DAVIDLENOVOPC\BPSINSTANCE/BPSSAMPLES]
     mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
   <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
 1  21       6  160    110  3.9     NA  16.5     0     1     4     4
 2  21       6  160    110  3.9      3  17.0     0     1     4     4
 3  22.8     4  108     93  3.85     2  18.6     1     1     4     1
[...]

PR #496, opened 2020-09-20
Use try_cast instead of cast for MSSQL version 11+ (2012+)

DavidPatShuiFong:
try_cast results in a more elegant failure (returns NA) if a cast is not possible ('cast' will normally throw an error).
Address issue #380

If Microsoft SQL is version 11 or greater (2012+) then call `try_cast` instead of `cast` when converting character to

* `logical`
* `Date`
* `POSIXct`
* `numeric`
* `double`
* `integer`
* `integer64`
* `character`
* `date`
* `datetime`

In line `155` of `R/backend-mssql.R`, `try_cast` is *also* used if `con` is not a `DBIObject`. This should not happen in normal operation, but is the case if the code is called during use of `testthat` in `tests/testthat/test-backend-mssql.R`.


DavidPatShuiFong:
A failure to **cast** (e.g. **as.numeric** or **as.double**, when the string contains a non-numeric character) results in an error:

```r
> library(magrittr)
> x <- mtcars %>% dplyr::mutate(wt = as.character(wt))
> x[1,6] <- "<2.62" # adds a non-numeric character '<' to an element
> DBI::dbWriteTable(con, "##mtcars", x) # where 'con' is an already defined connection to a MSSQL database
> con %>% dplyr::tbl("##mtcars") %>% dplyr::mutate(wt = as.numeric(wt))
Error: <SQL> 'SELECT TOP(11) "mpg", "cyl", "disp", "hp", "drat", CAST("wt" AS NUMERIC) AS "wt", "qsec", "vs", "am", "gear", "carb"
FROM "##mtcars"'
  nanodbc/nanodbc.cpp:1587: 42000: [Microsoft][ODBC SQL Server Driver][SQL Server]Error converting data type varchar to numeric. 
```

For Microsoft SQL (2012 and later) **try_cast** allows a more graceful return of *NA*. 

This pull request results in a more graceful failure (returns *NA*):

```r
> con %>% dplyr::tbl("##mtcars") %>% dplyr::mutate(wt = as.numeric(wt))
# Source:   lazy query [?? x 11]
# Database: Microsoft SQL Server 12.00.2000[bpsrawdata@DAVIDLENOVOPC\BPSINSTANCE/BPSSAMPLES]
     mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
   <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
 1  21       6  160    110  3.9     NA  16.5     0     1     4     4
 2  21       6  160    110  3.9      3  17.0     0     1     4     4
 3  22.8     4  108     93  3.85     2  18.6     1     1     4     1
 4  21.4     6  258    110  3.08     3  19.4     1     0     3     1
 5  18.7     8  360    175  3.15     3  17.0     0     0     3     2
 6  18.1     6  225    105  2.76     3  20.2     1     0     3     1
 7  14.3     8  360    245  3.21     4  15.8     0     0     3     4
 8  24.4     4  147.    62  3.69     3  20       1     0     4     2
 9  22.8     4  141.    95  3.92     3  22.9     1     0     4     2
10  19.2     6  168.   123  3.92     3  18.3     1     0     4     4
# ... with more rows
```

R/backend-mssql.R
```diff
@@ -150,8 +150,31 @@
           sql_expr(DATEPART(QUARTER, !!x))
         }
       },
+    )
 
-    ),
+  if (!is(con, "DBIObject") || sub("\\..*", "", DBI::dbGetInfo(con)$db.version) >= 11) {
```
hadley, line 155: Can you please pull this out into separate function? I think that will make the main thrust of the code easier to understand.

hadley (commented):

R/backend-mssql.R
DavidPatShuiFong: Thanks Hadley. Just to clarify, would you like me to place everything inside the curly braces { *re-assignment of mssql_scalr* } as a function? Or even the 'test condition' (*!is.con(...) and *version >= 11*) should also be in a separate function?

hadley: The test of whether or not an connection has an adequate version of SQL server

DavidPatShuiFong: okay, non-exported function `mssql_version()` returns a list of `ProductVersion_Major`/`ProductVersion_Minor`/`ProductVersion_Revision` (returns list of NULLs if argument `con` is not a valid `DBIObject`). `mssql_version()` is used to determine whether the MSSQL server version is >=11, in which case `sql_translate_env.Microsoft SQL Server` uses `TRY_CAST` instead of `CAST` where possible.

Also added to `tests/testthat/test-backend-mssql.R` to test the small non-exported function `mssql_interpret_version()` which is used to interpret the `$db.version` string returned by `DBI::dbGetInfo()`.

DavidPatShuiFong:
> 
> 
> Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

okay, done.

hadley:
I just pushed a simplification to your logic — for the real database case we can rely on `numeric_version()` to do most of the work, and for the testing case I added an explicit version component to the simulation.

DavidPatShuiFong:
> I just pushed a simplification to your logic — for the real database case we can rely on `numeric_version()` to do most of the work, and for the testing case I added an explicit version component to the simulation.

Great! Thanks for fixing for what even a non-programmer like me realized was rather ugly code. All whilst I was listening to Taiwan Digital Minister Tang talk about open data and governance for a sustainable development lecture. An honour to be involved in data and application! :)

R/backend-mssql.R
```diff
@@ -150,8 +150,27 @@
...
+  if (mssql_version(con) >= "11.0") { # MSSQL 2012
+    mssql_scalar <- sql_translator(
+      .parent = mssql_scalar,
+      as.logical = sql_try_cast("BIT"),
+      as.Date = sql_try_cast("DATE"),
+      as.POSIXct = sql_try_cast("TIMESTAMP"),
+      as.numeric = sql_try_cast("FLOAT"),
+      as.double = sql_try_cast("FLOAT"),
+      as.integer = sql_try_cast("NUMERIC"),
+      # in MSSQL, NUMERIC converts to integer
```
krlmlr, line 164: I wonder what's the reasoning behind this. When I do this, I get a `numeric` via FreeTDS:

``` r
con <- DBI::dbConnect(
  odbc::odbc(),
  "mssql-test",
  uid = "kirill", pwd = keyring::key_get("mssql", "kirill")
)

library(tidyverse)

tbl <-
  tibble(a = 1L) %>%
  copy_to(con, ., name = "tbl_a") %>%
  mutate(b = as.integer(a))
#> Created a temporary table named #tbl_a

tbl
#> # Source:   lazy query [?? x 2]
#> # Database: Microsoft SQL Server 15.00.4073[@tp/test]
#>       a     b
#>   <int> <dbl>
#> 1     1     1
dbplyr::sql_render(tbl)
#> <SQL> SELECT "a", TRY_CAST("a" AS NUMERIC) AS "b"
#> FROM "#tbl_a"
```

<sup>Created on 2020-11-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DavidPatShuiFong: In https://github.com/DavidPatShuiFong/dbplyr/commit/4933037fcdea29606ac29badefecee9e753b6eef I describe how I found that "in MSSQL casting to NUMERIC casts to integer, and is more flexible/graceful ... [NUMERIC] returns NA if unable to convert individual rows. Whereas casting to  INTEGER/INT results in the entire column becoming NA if a single row cannot be converted to integer."

krlmlr: Yeah, a double cast might work -- would still all rows be converted to `NULL` if the numbers are too large or too small?

DavidPatShuiFong: yes, double casting works! and doesn't drop the entire column if a single row has an invalid value.

Happy to know if there is a more elegant solution, but here is what I did (from https://github.com/tidyverse/dbplyr/compare/master...DavidPatShuiFong:feature/doubletrycastMSSQL )

changed `try_cast` (`translate_sql_helpers.R`) to accept a second optional parameter to do a second casting

```r
#' @rdname sql_variant
#' @export
sql_try_cast <- function(type1, type2 = NULL) {
  type1 <- sql(type1)
  if (is.null(type2)) {
    function(x) {
      sql_expr(try_cast(!!x %as% !!type1))
      # try_cast available in MSSQL 2012+
    }
  } else {
    type2 <- sql(type2)
    # if type2 is defined, then double-cast ('type1' then 'type2')
    function(x) {
      sql_expr(try_cast(try_cast(!!x %as% !!type1) %as% !!type2))
      # try_cast available in MSSQL 2012+
    }
  }
}
```

Change the casting of `as.integer` in `backend-mssql.R` for `mssql_version` >= 11.0

```r
 as.integer = sql_try_cast("NUMERIC", "INT"),
```

For the example I've given (I couldn't set up your example to work, perhaps because I am using an old version of `dplyr`?)

```r
> x <- con %>% dplyr::tbl("##nchild") %>% dplyr::mutate(n = as.integer(n))
> dbplyr::sql_render(x)
<SQL> SELECT "Year", "Country", TRY_CAST(TRY_CAST("n" AS NUMERIC) AS INT) AS "n"
FROM "##nchild"
> x
# Database: Microsoft SQL Server 12.00.2000[bpsrawdata@DESKTOP-BH8LGMO\BPSINSTANCE/bpssamples]
   Year                Country     n
   <chr>               <chr>   <int>
 1 1970-01-01 10:32:50 China      NA
 2 xyz                 China       4
 3 1970-01-01 10:33:00 China       3
 4 1970-01-01 10:33:05 China       3
 5 1970-01-01 10:33:10 China       2
 6 1970-01-01 10:33:15 China       2
 7 1970-01-01 10:33:20 China       2
 8 1970-01-01 10:33:25 China       2
 9 1970-01-01 10:33:30 China       2
10 1970-01-01 10:32:50 Taiwan      4
# ... with more rows
```

krlmlr: Thanks, this looks good. I don't know an easier solution off the top of my head.

DavidPatShuiFong: > 
> 
> Thanks, this looks good. I don't know an easier solution off the top of my head.

Thanks @krlmlr . After a bit more testing on the 'production code' which uses `dbplyr`, should I open a new issue and submit the fix as a pull request?

krlmlr: Thanks. A PR referring to this conversation should be enough.

gschone-data:
Hello, 
Would it be possible to also replace CAST by TRY_CAST for duckdb_connection ? 




```
#prepare connection 
drv <- duckdb::duckdb(config = list(allow_unsigned_extensions = "true"))
conDB <- duckdb::dbConnect(drv)
conDB |>  duckdb::duckdb_register(
  name = "iris", 
  df = iris,
  overwrite = TRUE)


# example 
x <- conDB %>% dplyr::tbl("iris") %>% dplyr::mutate(Petal.Width = as.numeric(Petal.Width))
dbplyr::sql_render(x)
```
Thanks

krlmlr:
@gschone-data: Good idea. Can you please file an issue with the duckdb R package? The translation lives there.

https://github.com/duckdb/duckdb-r

Comments to closed issues and PRs are likely to go unnoticed, comments to existing issues (or new issues) are generally preferred.

PR #514, opened 2020-10-07
add str_like() function

### LINKED ISSUE #509: Bring str_like to dbplyr
Currently have `str_like()` in `{stringr}`, will add a PR

rjpat:
closes #509 

Made the decision that for Hive and MSSQL it would provide a warning when trying to be case sensitive and just continue on insensitively

``` r
library(dbplyr)
library(stringr)

lf <- lazy_frame(x = c("abc", "def"), con = simulate_postgres())

lf %>% 
  filter(str_like(x, "a%")) 
#> <SQL>
#> SELECT *
#> FROM `df`
#> WHERE (`x` LIKE 'a%')
```

### DIFF
--- a/R/backend-.R
+++ b/R/backend-.R
@@ -237,6 +237,7 @@ base_scalar <- sql_translator(
   str_glue = sql_not_supported("str_glue()"),
   str_glue_data = sql_not_supported("str_glue_data()"),
   str_interp = sql_not_supported("str_interp()"),
+  str_like = sql_not_supported("str_like()"),
   str_locate = sql_not_supported("str_locate()"),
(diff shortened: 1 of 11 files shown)


hadley:
You can already do this with infix function translation:

``` r
library(dbplyr)

lf <- lazy_frame(x = c("abc", "def"))

lf %>% 
  filter(x %like% "a%") 
#> <SQL>
#> SELECT *
#> FROM `df`
#> WHERE (`x` like 'a%')
```

<sup>Created on 2020-10-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

I'm not sure adding support for `str_like()` helps much.

PR #521, opened 2020-10-16
Add back specific RPoostgreSQL methods and tests

### LINKED ISSUE #516: RPostgreSQL column names regression
I've been testing the development branch of dbplyr and I believe there might be a regression with RPostgreSQL backends: it appears the column names are not being added to the `ops` of the `tbl` object so subsequent verbs that reference those columns (`select()`) don't work.

I believe this is related to [removing the special-case](https://github.com/tidyverse/dbplyr/commit/f51eed49d85c71cf849541465d1943d5b4a172ce) `db_query_fields.PostgreSQLConnection()` called [from `tbl_sql()`](https://github.com/tidyverse/dbplyr/blob/master/R/tbl-sql.R#L19). I'm happy to submit a PR that brings back `db_query_fields.PostgreSQLConnection()` or try a different approach.

```r
test_user = "[redacted]"
test_pw = "[redacted]"

library(dbplyr)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:dbplyr':
#> 
#>     ident, sql
#> 
#>     filter, lag
#> 

# RPostgreSQL
con_rpsql <- DBI::dbConnect(
  RPostgreSQL::PostgreSQL(),
  dbname = "travelling",
  host = "127.0.0.1",
  user = test_user,
  password = test_pw
)
DBI::dbWriteTable(con_rpsql, "mtcars", mtcars)
#> [1] TRUE
DBI::dbDisconnect(con_rpsql)
#> [1] TRUE

con_rpsql <- DBI::dbConnect(
  RPostgreSQL::PostgreSQL(),
  dbname = "travelling",
  host = "127.0.0.1",
  user = test_user,
  password = test_pw
)
str(tbl(con_rpsql, "mtcars"))
#> List of 2
#>  $ src:List of 2
#>   ..$ con  :Formal class 'PostgreSQLConnection' [package "RPostgreSQL"] with 1 slot
#>   .. .. ..@ Id: int [1:2] 92428 1
#>   ..$ disco: NULL
#>   ..- attr(*, "class")= chr [1:4] "src_PostgreSQLConnection" "src_dbi" "src_sql" "src"
#>  $ ops:List of 2
#>   ..$ x   : 'ident' chr "mtcars"
#>   ..$ vars: chr(0) 
#>   ..- attr(*, "class")= chr [1:3] "op_base_remote" "op_base" "op"
#>  - attr(*, "class")= chr [1:5] "tbl_PostgreSQLConnection" "tbl_dbi" "tbl_sql" "tbl_lazy" ...
tbl(con_rpsql, "mtcars") %>%
  select(mpg)
#> Error: Can't subset columns that don't exist.
#> x Column `mpg` doesn't exist.
DBI::dbDisconnect(con_rpsql)
#> [1] TRUE

<sup>Created on 2020-10-08 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>
```

jonkeane:
This resolves #516 and adds back `db_query_fields.PostgreSQLConnection()` and `db_write_table.PostgreSQLConnection()` so that RPostgreSQL backends work again. I also added tests (which required RPostgreSQL to be added to Suggests). The one test might not be worth the addition to Suggests, in which case I’m happy to take it out, but wanted to submit with to confirm it works.

I’m also not certain that the change to `dbplyr_fallback()` is the right place to special-case RPostgreSQL.


R/backend-postgres.R
```diff
@@ -229,6 +229,36 @@ sql_expr_matches.PqConnection <- sql_expr_matches.PostgreSQLConnection
 
 # DBI methods ------------------------------------------------------------------
 
+#' @export
+db_write_table.PostgreSQLConnection <- function(con, table, types, values,
```
hadley, line 233: I think maybe it might be best to put this in a separate file, to make it more clear that its for backward compatibility with RPostgresSQL. Then the test could be specific to RPostgreSQL (no longer using the `test_register_con()`) and exercise the full pipeline (i.e. copy data in, write simple query, and retrieve data).

hadley:
Thanks @jonkeane!

PR #533, opened 2020-11-09
Add pivot_longer()

### LINKED ISSUE #532: Add `pivot_longer()`

mgirlich:
Closes #532.
After nothing seems to have happened in PR #342 for more than a year I thought it would be best to create a new PR for this.

I tried to support all the features from `tidyr::pivot_longer()` but there are some open questions:

1. `pivot_longer()` in `tidyr` uses S3 dispatch but `build_longer_spec()` and `pivot_longer_spec()` do not. For now, I simply overwrite both these functions which of course is not the way to go. Would you prefer to change both of them to use S3 dispatch as well or should the functions here have different names?
2. The values are not interleaved in the same order as in `tidyr`. Should we a) not care about the order, b) order them explicitly with `arrange()`, or c) add a flag so the user can decide?
3. What if the resulting column is mixed? Use the db default (which is quite nasty for [RSQLite](https://github.com/r-dbi/RSQLite/issues/323))? Try to check compatibility locally? Try to coerce similarly to vctrs?
4. Should `values_transform` be supported? This seems quite reasonable to me.
5. Should `values_ptypes` be supported? Here only a subset of ptypes would make sense so I am not sure about this one.
6. There is quite a lot of code copy pasted from tidyr just for the string processing. This is required for the support of the `names_sep` and `names_pattern` argument. How do you want to deal with that?
7. For now, I just copy pasted most of the documentation from `tidyr`. How do you want to handle the documentation?

Example for the generated SQL
```r
pivot_longer(
  memdb_frame(!!!anscombe),
  everything(),
  names_to = c(".value", "set"),
  names_pattern = "(.)(.)"
) %>% 
  show_query()
```

```sql
SELECT '1' AS `set`, `x1` AS `x`, `y1` AS `y`
FROM `dbplyr_148`
UNION ALL
SELECT '2' AS `set`, `x2` AS `x`, `y2` AS `y`
FROM `dbplyr_148`
UNION ALL
SELECT '3' AS `set`, `x3` AS `x`, `y3` AS `y`
FROM `dbplyr_148`
UNION ALL
SELECT '4' AS `set`, `x4` AS `x`, `y4` AS `y`
FROM `dbplyr_148`
```

The generated SQL code is quite a bit shorter than in the other PR as I didn't use `inner_join()`. I don't understand why @edgararuiz used the `inner_join()` so I left it out and it seems to work anyways...

I also checked for the examples in the `pivot_longer()` documentation whether the results match (up to the ordering)

<details>
<summary>
Test with tidyr examples
</summary>

```r
devtools::load_all(".")
con <- DBI::dbConnect(RSQLite::SQLite(), path = ":dbname:")

relig_income <- tidyr::relig_income
billboard <- tidyr::billboard
who <- tidyr::who


equal_pivot_longer <- function(x, ...) {
  x_long <- tidyr::pivot_longer(x, ...) %>% 
    dplyr::arrange(dplyr::across(everything()))
  
  y_long <- pivot_longer(
    copy_to(con, x, name = "tmp_table", overwrite = TRUE),
    ...
  ) %>% 
    collect() %>% 
    dplyr::arrange(dplyr::across(everything()))
  
  all.equal(
    x_long,
    y_long
  )
}

equal_pivot_longer(
  relig_income,
  !religion, names_to = "income", values_to = "count"
)

equal_pivot_longer(
  billboard %>% mutate(date.entered = as.character(date.entered)),
  cols = starts_with("wk"),
  names_to = "week",
  names_prefix = "wk",
  values_to = "rank",
  values_drop_na = TRUE
)

# Multiple variables stored in column names
equal_pivot_longer(
  who,
  cols = new_sp_m014:newrel_f65,
  names_to = c("diagnosis", "gender", "age"),
  names_pattern = "new_?(.*)_(.)(.*)",
  values_to = "count"
)

# Multiple observations per row
equal_pivot_longer(
  anscombe,
  everything(),
  names_to = c(".value", "set"),
  names_pattern = "(.)(.)"
)
```

</details>


R/verb-pivot-longer.R
```diff
@@ -0,0 +1,517 @@
...
+    idx <- pieces$val[[i]]
+    copy[idx] <- seq_along(idx)
+  }
+
+  spec$.seq <- copy
+  spec
+}
+
+str_separate <- function(x, into, sep, convert = FALSE, extra = "warn", fill = "warn") {
+  # COPIED FROM tidyr
```
hadley, line 332: Can you please stick all the copied functions into a separate file, with a comment header at top saying that they came from tidyr?

tests/testthat/test-verb-pivot-longer.R
```diff
@@ -0,0 +1,258 @@
+test_that("can pivot all cols to long", {
+  df <- tibble(x = 1:2, y = 3:4)
+  pv <- pivot_longer(memdb_frame(!!!df), x:y) %>% collect()
```
hadley, line 3: You can use `tbl_memdb()` here

tests/testthat/test-verb-pivot-longer.R
```diff
@@ -0,0 +1,258 @@
+test_that("can pivot all cols to long", {
```
hadley, line 1: Again, I'd prefer to start with some simple tests, mostly focussing on the SQL generation, and with only a few checking the actual execution. I think we'll also have to tackle the tests in pieces so that I can progressively build up an understanding of what's going on.

R/verb-pivot-longer.R
```diff
@@ -0,0 +1,517 @@
...
+    names[[col]] <- f(names[[col]])
+  }
+
+  out <- tibble(.name = names(cols))
+  out[[".value"]] <- values_to
+  out <- vctrs::vec_cbind(out, names)
+  out
+}
+
+#' Pivot data from wide to long using a spec
```
hadley, line 193: I don't think think these need to be exported (at least for now), and I'd call them `dbplyr_pivot_longer_spec()` just to make it very clear.

hadley (commented):
> pivot_longer() in tidyr uses S3 dispatch but build_longer_spec() and pivot_longer_spec() do not. For now, I simply overwrite both these functions which of course is not the way to go. Would you prefer to change both of them to use S3 dispatch as well or should the functions here have different names?
  
See comments
  
> The values are not interleaved in the same order as in tidyr. Should we a) not care about the order, b) order them explicitly with arrange(), or c) add a flag so the user can decide?
  
I'd say start with database order, and then we'll see what people think.
  
> What if the resulting column is mixed? Use the db default (which is quite nasty for RSQLite)? Try to check compatibility locally? Try to coerce similarly to vctrs?

I think we just have to stick with the database default here; otherwise we have to simulate the database type system locally, which will be hard.

> Should values_transform be supported? This seems quite reasonable to me.

Sure, if you want to implement it.

> Should values_ptypes be supported? Here only a subset of ptypes would make sense so I am not sure about this one.

I don't think so; as above, I think we just have to rely on type inference in the database for this. Fortunately (apart from SQLite), databases seem handle this in a pretty sane way.

> There is quite a lot of code copy pasted from tidyr just for the string processing. This is required for the support of the names_sep and names_pattern argument. How do you want to deal with that?

See comment - put in a separate file for now.

> For now, I just copy pasted most of the documentation from tidyr. How do you want to handle the documentation?

I think the documentation should assume that the user already understands `tidyr::pivot_longer()`/`tidyr::pivot_wider()` and can focus on the DB side of things. It can be brief to start with, but I'd appreciate a sketch of how the SQL translation works, since I've already forgotten since the last time Edgar explained it to me.

Thanks for working on this! I'll try and stay on top of getting feedback to you, but it's likely to take a few days. Please feel free to ping me (with `@hadley`) if you don't hear from me for a while.

R/verb-pivot-longer.R
```diff
@@ -0,0 +1,302 @@
...
+  dbplyr_pivot_longer_spec(data, spec,
+    names_repair = names_repair,
+    values_drop_na = values_drop_na,
+    values_transform = values_transform
+  )
+}
+
+# this is a nearly identical copy of `tidyr::build_longer_spec` adapted to
+# work with `tbl_lazy` objects.
+dbplyr_build_longer_spec <- function(data, cols,
```
hadley, line 105: Do you think there's some way that we could use `tidyr::build_longer_spec()`, possibly by creating a dummy dataframe that matches the structure of the database table? I've admittedly only skimmed `dbplyr_build_longer_spec()`, but I don't see anything obviously database specific.

R/verb-pivot-longer.R
```diff
@@ -0,0 +1,302 @@
...
+  # .seq col needed if different input columns are mapped to the same output
+  # column
+  spec <- deduplicate_spec(spec, data)
+
+  id_cols <- syms(setdiff(colnames(data), spec$.name))
+
+  spec_split <- vctrs::vec_split(spec, spec[, -(1:2)])
+  data_long_list <- purrr::map(
+    vctrs::vec_seq_along(spec_split),
+    function(idx) {
```
hadley, line 193: I think this is complicated enough that I'd want to pull it out into a top-level function

R/verb-pivot-longer.R
```diff
@@ -0,0 +1,302 @@
...
+    . <- NULL
+
+    data_long <- dplyr::filter_at(
+      data_long,
+      value_cols,
+      dplyr::all_vars(!is.na(.))
+    )
+  }
+
+  return(data_long)
```
hadley, line 240:
```suggestion
  data_long
```

R/verb-pivot-longer.R
```diff
@@ -0,0 +1,302 @@
...
+      )
+    }
+  )
+
+  data_long <- purrr::reduce(data_long_list, union_all)
+
+
+  if (values_drop_na) {
+    value_cols <- unique(spec$.value)
+    # add "." variable locally so that cran doesn't complain about
```
hadley, line 229: Handle with `globalVariables()` like everything else?

hadley (commented):
I think I basically see the overall flow here. I'll try and dig into it more next week.

R/verb-pivot-longer.R
mgirlich: Yes, it actually was trivial.

R/verb-pivot-longer.R
mgirlich: I instead extracted parts of the function into a top level function so there is no map() called directly. I think it looks quite simple now.

R/verb-pivot-longer.R
```diff
@@ -0,0 +1,235 @@
...
+        .x
+      } else {
+        expr((!!f_trans)(!!.x))
+      }
+    }
+  )
+}
+
+# The following is copy-pasted from tidyr
+# with the exception of `simplifyPieces()` which I have quickly hacked in R
```
hadley, line 175: Can you add a comment pointing to https://github.com/tidyverse/tidyr/issues/1087 please?

hadley: And use exclusion comments to make sure that it's not counted by test coverage: https://covr.r-lib.org/#exclusion-comments

hadley (commented):
Two tiny last comments, and then if you merge/rebase, I'll accept.

mgirlich:
Thanks a lot for your effort of reviewing!

PR #537, opened 2020-11-10
Support sqlite full join

mgirlich:
Closes #536

PR #539, opened 2020-11-10
Add expand() and complete()

### LINKED ISSUE #538: add translations for expand() and complete()

mgirlich:
Closes #538 

* `complete()` needs the full join fix to work in SQLite (PR #537)
* should the output of `expand()` be sorted explicitly like in `tidyr`?
* what should `expand()` return in case of empty dots?
* should `dbplyr` depend on `tidyr`? or should `tidyr` be a suggested package and the S3 methods are registered on loading?

**nesting**
* is there a use case for nesting outside of `expand()`?
* the way I catch and handle `nesting()` is probably not the best.


DESCRIPTION
```diff
@@ -37,7 +37,9 @@ Imports:
     tidyselect (>= 0.2.4),
     blob (>= 1.2.0),
     utils,
-    withr
+    withr,
+    tidyr,
```
hadley, line 41: I think these probably should just be in suggests, and do conditional method registration.

tests/testthat/test-verb-expand.R
```diff
@@ -0,0 +1,122 @@
+test_that("expand completes all values", {
+  df <- memdb_frame(x = 1:2, y = 1:2)
+  out <- expand(df, x, y) %>% collect()
+  expect_equal(nrow(out), 4)
+})
+
+test_that("multiple variables in one arg doesn't expand", {
```
hadley, line 7: Can you remove all the tidyr specific tests and start fresh with tests of the basic behaviour. I _think_ they should probably mostly snapshot the generated sql, with a few actuallyy executing just to check that everything is plumbed together correctly.

R/verb-expand.R
```diff
@@ -0,0 +1,44 @@
+#' @importFrom tidyr expand
+#' @export
+expand.tbl_lazy <- function(data, ..., .name_repair = "check_unique") {
+  # TODO sort output?
+  # TODO wait for bugfix: distinct() ignores groups
+  # https://github.com/tidyverse/dbplyr/issues/535
+  dots <- quos_auto_name(quos(...))
+  dots <- purrr::discard(dots, rlang::quo_is_null)
+  names(dots) <- vctrs::vec_as_names(names(dots), repair = .name_repair)
```
hadley, line 9: I _think_ this should just be left to `distinct()` to handle

R/verb-expand.R
```diff
@@ -0,0 +1,44 @@
...
+  if (is_empty(dots)) {
+    # TODO what should be returned here?
+    abort("No columns selected")
+  }
+
+  distinct_tables <- purrr::imap(
+    dots,
+    ~ {
+      # ugly hack to deal with `nesting()`
+      if (rlang::quo_is_call(.x) && rlang::call_name(.x) == "nesting") {
```
hadley, line 20: Can simplify this by using the `name`  arg to `quo_is_call()`. Also note the rlang namespace is already imported, so you don't need `rlang::`.

I think the hack is fine; doing it 100% correctly is going to be a lot of work (and it's not even clear to me how you would do it)

R/verb-expand.R
```diff
@@ -0,0 +1,44 @@
...
+#' @importFrom tidyr complete
+#' @export
+complete.tbl_lazy <- function(data, ..., fill = list()) {
+  full <- expand(data, ...)
+
+  if (is_empty(full)) {
+    return(data)
+  }
+
+  full <- dplyr::full_join(full, data, by = colnames(full))
```
hadley, line 42: Similarly, don't need `dplyr::` here

R/verb-expand.R
```diff
@@ -0,0 +1,44 @@
+#' @importFrom tidyr expand
+#' @export
+expand.tbl_lazy <- function(data, ..., .name_repair = "check_unique") {
+  # TODO sort output?
```
hadley, line 4: I don't think so; I think the usual principle is to respect database behaviour unless the user specifically requests otherwise

R/verb-expand.R
```diff
@@ -0,0 +1,44 @@
...
+        x_expr <- rlang::quo_get_expr(.x)
+        args <- rlang::call_args(x_expr)
+        distinct(data, !!!args)
+      } else {
+        distinct(data, !!.y := !!.x)
+      }
+    }
+  )
+
+  purrr::reduce(distinct_tables, left_join, by = character())
```
hadley, line 30: Should this be a join or a union?

R/verb-expand.R
```diff
@@ -0,0 +1,44 @@
+#' @importFrom tidyr expand
```
hadley, line 1: These will need docs to help explain how they differ from the R versions.

R/verb-expand.R
```diff
@@ -0,0 +1,44 @@
...
+  # TODO sort output?
+  # TODO wait for bugfix: distinct() ignores groups
+  # https://github.com/tidyverse/dbplyr/issues/535
+  dots <- quos_auto_name(quos(...))
+  dots <- purrr::discard(dots, rlang::quo_is_null)
+  names(dots) <- vctrs::vec_as_names(names(dots), repair = .name_repair)
+
+  if (is_empty(dots)) {
+    # TODO what should be returned here?
+    abort("No columns selected")
```
hadley, line 13: I think it's fine for this to be an error; just needs to be a bit more explicit (and follow https://style.tidyverse.org/error-messages.html). Maybe something like "Must supply variables in `...`" ?

R/verb-expand.R
mgirlich: This doesn't work for two reasons:
1. `distinct()` doesn't rename with a repair strategy
2. `distinct()` cannot handle this as the code iterates over `dots` and therefore `distinct()` never sees all the variables at once.

hadley: In that case, I'd suggest filing an issue to remind us that distinct needs to duplicated variables, and we can come back to it later.

DESCRIPTION
mgirlich: I don't understand how to make this work properly without loading tidyr... If `tidyr` isn't loaded then it doesn't find the function `expand()` and `complete()`. And loading `tidyr` into the namespace doesn't seem like an option.

hadley: So the key idea is that we no longer use the NAMESPACE, and instead dynamically register using code in `.onLoad`. You can see the existing code in `.onLoad()` (although that's there for a different reason due to the problems with `dplyr::filter` being a generic and `stats::filter` not). You can get a few more details in `?vctrs::s3_register`, and otherwise let me know if you need more explanation.

mgirlich: That's what I did in this commit https://github.com/tidyverse/dbplyr/pull/539/commits/5befb0c613dcb00a0bd78f85f956e9a2ec1f926e. The issue is that the user needs to still load `tidyr`. As I understand only if `expand()` is loaded the S3 registration actually works as otherwise R doesn't find the generic `expand()` itself from which it should dispatch.

R/verb-expand.R
mgirlich: Hmm, seems like we misunderstand each other. `distinct()` does handle duplicated variables. It processes them step by step, similar to what `mutate()` does. See the following example.

    library(dplyr)
    library(dbplyr)

    memdb_frame(x = 1) %>% 
      distinct(x = x + 1, x = x + 1)
    #> # Database: sqlite 3.33.0 [:memory:]
    #>       x
    #>   <dbl>
    #> 1     3

    tibble(x = 1) %>% 
      distinct(x = x + 1, x = x + 1)
    #> # A tibble: 1 x 1
    #>       x
    #>   <dbl>
    #> 1     3

<sup>Created on 2020-11-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

In comparison `tidyr::expand()` renames

    tidyr::expand(
      dplyr::tibble(x = 1:2),
      x, x,
      .name_repair = "unique"
    )
    #> New names:
    #> * x -> x...1
    #> * x -> x...2
    #> # A tibble: 4 x 2
    #>   x...1 x...2
    #>   <int> <int>
    #> 1     1     1
    #> 2     1     2
    #> 3     2     1
    #> 4     2     2

<sup>Created on 2020-11-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

So, I don't see how we can use `distinct()` to handle the renaming. FYI I had to update my strategy for renaming to also handle variables in `nesting()`

R/verb-expand.R
mgirlich: This must be a cross join. This is the essential part to create all combinations between data. Basically the following `expand()`

    library(tidyr)
    library(dplyr)

    x <- c(1, 1, 2)
    y <- c("a", "b", "c")
    expand(
      tibble(x, y),
      x, y
    )
    #> # A tibble: 6 x 2
    #>       x y    
    #>   <dbl> <chr>
    #> 1     1 a    
    #> 2     1 b    
    #> 3     1 c    
    #> 4     2 a    
    #> 5     2 b    
    #> 6     2 c

translates to a cross join of two tibbles with unique values

    left_join(
      tibble(x) %>% distinct(),
      tibble(y) %>% distinct(),
      by = character()
    )
    #> # A tibble: 6 x 2
    #>       x y    
    #>   <dbl> <chr>
    #> 1     1 a    
    #> 2     1 b    
    #> 3     1 c    
    #> 4     2 a    
    #> 5     2 b    
    #> 6     2 c

<sup>Created on 2020-11-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

DESCRIPTION
hadley: Right, I think that's what we want?

R/verb-expand.R
hadley: Oooooook, now I get it!

R/verb-expand.R
hadley: I think I put my comment in the wrong place — can `complete()` use a union instead of a join?

mgirlich: No, this doesn't work. Union needs the same columns in both tables, so missing columns are created and filled with NULL. Then, the union doesn't work correctly anymore as you can see in the example

```r
# result with full join
#> # A tibble: 4 x 5
#>   group item_id item_name value1 value2
#>   <dbl>   <dbl> <chr>      <int>  <int>
#> 1     1       1 a              1      4
#> 2     1       2 b              3      6
#> 3     2       1 a             NA     NA
#> 4     2       2 b              2      5

# result with union
#> # A tibble: 7 x 5
#>   group item_id item_name value1 value2
#>   <dbl>   <dbl> <chr>      <int>  <int>
#> 1     1       1 a             NA     NA
#> 2     1       1 a              1      4
#> 3     1       2 b             NA     NA
#> 4     1       2 b              3      6
#> 5     2       1 a             NA     NA
#> 6     2       2 b             NA     NA
#> 7     2       2 b              2      5
```

Also this way the code is identical to `tidyr:::complete.data.frame()` up to a missing call to `reconstruct_tibble(data, full)`.

hadley: Ok, makes sense.

R/verb-expand.R
```diff
@@ -0,0 +1,144 @@
+#' Expand SQL tables to include all possible combinations of values
+#'
+#' @description
+#' This is a method for the [tidyr::expand] generics. It doesn't sort the
+#' result explicitly, so the order might be different to what `expand()`
+#' returns for data frames.
+#'
+#' @param data A pair of lazy data frame backed by database queries.
```
hadley, line 8: A pair?

R/verb-expand.R
```diff
@@ -0,0 +1,144 @@
...
+#'   # All possible combinations ---------------------------------------
+#'   fruits %>% expand(type)
+#'   fruits %>% expand(type, size)
+#'
+#'   # Only combinations that already appear in the data ---------------
+#'   fruits %>% expand(nesting(type, size))
+#' }
+expand.tbl_lazy <- function(data, ..., .name_repair = "check_unique") {
+  # TODO wait for bugfix: distinct() ignores groups
+  # https://github.com/tidyverse/dbplyr/issues/535
```
hadley, line 31: Can remove this now?

R/verb-expand.R
```diff
@@ -0,0 +1,144 @@
...
+  dots <- purrr::discard(quos(...), quo_is_null)
+
+  if (is_empty(dots)) {
+    abort("Must supply variables in `...`")
+  }
+
+  distinct_tbl_vars <- purrr::imap(
+    dots,
+    ~ {
```
hadley, line 40: Since these span multiple lines, can you please pull out into named function inside `expand`?

R/verb-expand.R
```diff
@@ -0,0 +1,144 @@
...
+    }
+  )
+
+  purrr::reduce(distinct_tables, left_join, by = group_vars(data))
+}
+
+#' Complete a SQL table with missing combinations of data
+#'
+#' Turns implicit missing values into explicit missing values. This is a method
+#' for the [tidyr::expand()] generic.
```
hadley, line 73:
```suggestion
#' for the [tidyr::complete()] generic.
```

R/verb-expand.R
```diff
@@ -0,0 +1,144 @@
...
+  replace_na(full, replace = fill)
+}
+
+#' Replace NAs with specified values
+#'
+#' This is a method for the [tidyr::replace_na()] generic.
+#'
+#' @param data A pair of lazy data frame backed by database queries.
+#' @param replace A list of values, with one value for each column that has NA
+#' values to be replaced.
```
hadley, line 113:
```suggestion
#'   values to be replaced.
```

R/verb-expand.R
```diff
@@ -0,0 +1,144 @@
...
+#' }
+replace_na.tbl_lazy <- function(data, replace = list(), ...) {
+  stopifnot(is_list(replace))
+  replace <- replace[names(replace) %in% colnames(data)]
+
+  if (is_empty(replace)) {
+    return(data)
+  }
+
+  coalesce_expr <- purrr::imap(
```
hadley, line 131: Need to check that `replace` has names?

R/verb-expand.R
```diff
@@ -0,0 +1,144 @@
...
+}
+
+#' Replace NAs with specified values
+#'
+#' This is a method for the [tidyr::replace_na()] generic.
+#'
+#' @param data A pair of lazy data frame backed by database queries.
+#' @param replace A list of values, with one value for each column that has NA
+#' values to be replaced.
+#' @param ... Additional arguments for methods. Currently unused.
```
hadley, line 114: In this context, I'd write: "Unused; included for compatibility with generic"

R/verb-expand.R
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  coalesce_expr <- purrr::imap(
+    replace,
+    function(value, col) {
+      quo(coalesce(!!sym(col), !!value))
+    }
+  )
+
+  mutate(
```
hadley, line 138: Put this on one line?

tests/testthat/test-verb-expand.R
```diff
@@ -0,0 +1,138 @@
+test_that("expand completes all values", {
+  skip_if_not_installed("tidyr")
```
hadley, line 2: In tests, you can just assume that suggested packages are installed, so you can remove all these skips. I'd also have a minor preference to avoid `local_package()`, just because it only saves a small amount of typing.

hadley (commented):
This is looking great — just a few last minor comments

tests/testthat/test-verb-expand.R
mgirlich: I am not sure whether I understood your point about `local_package()` completely. I replaced it with
`suppressPackageStartupMessages(library(tidyr))` and `detach("package:tidyr")`. Is this what you had in mind?

hadley: No, I mean just don't use it and use `::` instead

mgirlich: oh, oh, well, that would have been too obvious... Sorry about that!

hadley:
Thanks!!

PR #541, opened 2020-11-11
escape NA_real_ correctly

yitao-li:

### DIFF
--- a/R/escape.R
+++ b/R/escape.R
@@ -81,8 +81,8 @@ escape.character <- function(x, parens = NA, collapse = ", ", con = NULL) {
 escape.double <- function(x, parens = NA, collapse = ", ", con = NULL) {
   out <- ifelse(is.wholenumber(x), sprintf("%.1f", x), as.character(x))
 
-  # Special values
-  out[is.na(x)] <- "NULL"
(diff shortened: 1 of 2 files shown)


hadley:
This doesn't feel like the right fix to me.

yitao-li:
@hadley I wonder if the null-translation can be made customizable with generics similar to for example `sql_join_suffix()` of dbplyr 2.0 which was mentioned in https://dbplyr.tidyverse.org/articles/backend-2.html

Current state of this as I understood it:
- Translating `NA_real_` to `cast(null as double)` in SQL only works for DB backends that are SQL-92 compliant (and in particular won't work for MySQL)
- Translating it to `null * 2e-16` or some arbitrary arithmetic expression involving null and another double-precision number works for all DB backends I know of, but is quite ugly and impacts readability of the resulting SQL
- Translating `NA_real_` to just `null` in SQL will cause loss of type info when creating a new column of all `null`s
- Maybe there is some other better way of fixing this, but I haven't thought of such possibility yet

hadley:
Fixing this in dbplyr just feels like the wrong level of abstraction to me

PR #555, opened 2020-12-02
Use double cast for MS SQL as.integer() and as.integer64() translation

DavidPatShuiFong:
used by MSSQL (>=11 2012+) as.integer to cast first to numeric, and then to integer, issue #380 

Fix to error in code discovered by @krlmlr in line 163 of `R/backend-mssql.R`, conversation in https://github.com/tidyverse/dbplyr/pull/496#discussion_r523986061

Unfortunately `sql_try_cast("NUMERIC")` rounds the numbers to integer, but still returns a double. However `sql_try_cast("INTEGER")` will return `NA` for an entire column if there is an invalid value in a single row.

Re-define `sql_try_cast` to accept an (optional) second parameter for a 'double cast'.

In the case of `as.integer`, define as `sql_try_cast("NUMERIC", "INT")` to return an integer value.


R/translate-sql-helpers.R
```diff
@@ -245,11 +245,20 @@ sql_cast <- function(type) {
 
 #' @rdname sql_variant
 #' @export
-sql_try_cast <- function(type) {
-  type <- sql(type)
-  function(x) {
-    sql_expr(try_cast(!!x %as% !!type))
-    # try_cast available in MSSQL 2012+
+sql_try_cast <- function(type1, type2 = NULL) {
```
hadley, line 248: Rather than making this function more complicated, can you please just replace the translation for `as.integer()`, expanding the comment a little more to make it more clear why this is necessary.

hadley (commented):
Can you please also add a test (next to the existing test for `as.numeric()`) and a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

R/translate-sql-helpers.R
DavidPatShuiFong: (Thanks, yes, I can add a test for `as.numeric` and add a line to `NEWS.md`)

Just to clarify regarding 'just replace the translation for `as.integer()`, would you like me to leave the `sql_try_cast` function unchanged, and instead change the definition of `as.integer` in [R/backend-mssql.R](https://github.com/tidyverse/dbplyr/blob/master/R/backend-mssql.R) from using `sql_try_cast` and, say, define the function 'in-place' within [R/backend-mssql.R](https://github.com/tidyverse/dbplyr/blob/master/R/backend-mssql.R)?

e.g.

```r
as.integer = function(x) {
  sql_expr(try_cast(try_cast(!!x %as% "NUMERIC") %as% "INT"))
}
# 'try_cast' directly to INT results in returning NA for an entire column if a single row has an invalid value
# in MSSQL, NUMERIC returns a double, rounded to an integer, and
# does not return NA for an entire column if a single row has an invalid value
```

hadley: Yup!

DavidPatShuiFong:
removed changes to `sql_try_cast` (so, as before, `sql_try_cast` only accepts one parameter). added description to `NEWS.md` and added tests.

Turns out that `TRY_CAST("n" A BIGINT)` suffers the same problem as `TRY_CAST("n" AS INT)` i.e. if there is a single invalid row, then the entire column returns `NA`. Fix added in this pull request.

hadley:
Could you please also point me to the documentation that describes the difference in behaviour between `CAST(x as NUMERIC)` and `CAST(x as INTEGER)`?

DavidPatShuiFong:
> 
> 
> Could you please also point me to the documentation that describes the difference in behaviour between `CAST(x as NUMERIC)` and `CAST(x as INTEGER)`?

Sadly, I can't find documentation relating specifically to the difference, although there are some StackOverflow questions and answers showing people having difficulty in dealing with invalid character to integer conversions.

https://stackoverflow.com/questions/15977911/sql-server-converting-varchar-to-int
https://stackoverflow.com/questions/728833/convert-a-string-to-int-using-sql-query - the last answer by Steven de Salas seems to allude to the improved success in converting STRING -> NUMERIC -> INT.

I described the problem with converting character to integer in issue #380 and pull #496 , but here is a summary to reproduce the error, with Microsoft Server 12.0.2000 (Microsoft SQL Server 2014).

Creating a temporary table with `R`, including invalid rows for character to integer conversion.

```r
df <- data.frame(China = c(5.72, 3.86, 2.6, 2.65, 2.31, 1.66, 1.60, 1.61, 1.63),
                          Year = as.POSIXct(c(1970, 1975, 1980, 1985, 1990, 1995, 2000, 2005, 2010), origin = "1970-01-01"),
                          row.names = c("1970", "1975", "1980", "1985", "1990", "1995", "2000", "2005", "2010"))
 
df <- df %>%
  gather(key = "Country", value = "NchildPerWoman", -Year) %>%
  rename(n = NchildPerWoman) %>%
  mutate(n = as.character(n), Year = as.character(Year)) # convert two columns to character
 
df[1,3] <- "5.72x" # creates invalid numeric value
df[2,1] <- "xyz" # create invalid date
 
# where 'con' is a connection to a MSSQL database
DBI::dbWriteTable(con, "##nchild", df)
con %>% dplyr::tbl("##nchild")
```

produces the result

```
# Source:   table<##nchild> [?? x 3]
# Database: Microsoft SQL Server 12.00.2000[bpsrawdata@DESKTOP-BH8LGMO\BPSINSTANCE/bpssamples]
  Year                Country n    
  <chr>               <chr>   <chr>
1 1970-01-01 10:32:50 China   5.72x
2 xyz                 China   3.86 
3 1970-01-01 10:33:00 China   2.6  
4 1970-01-01 10:33:05 China   2.65 
5 1970-01-01 10:33:10 China   2.31 
6 1970-01-01 10:33:15 China   1.66 
7 1970-01-01 10:33:20 China   1.6  
8 1970-01-01 10:33:25 China   1.61 
9 1970-01-01 10:33:30 China   1.63 
```

Now, using Microsoft SQL Server Management Studio to do various SQL queries on the temporary table `##nchild`.
The problem of `TRY_CAST` with `INT`/`BIGINT` returning `NULL` for an entire column even if only a single row in invalid happens at the Microsoft SQL Server level.

```
SELECT "Year", "Country", TRY_CAST("n" AS FLOAT) AS "n" FROM "##nchild"

Year	Country	n
1970-01-01 10:32:50	China	NULL
xyz	China	3.86
1970-01-01 10:33:00	China	2.6
1970-01-01 10:33:05	China	2.65
1970-01-01 10:33:10	China	2.31
1970-01-01 10:33:15	China	1.66
1970-01-01 10:33:20	China	1.6
1970-01-01 10:33:25	China	1.61
1970-01-01 10:33:30	China	1.63
```

```
SELECT "Year", "Country", TRY_CAST("n" AS INTEGER) AS "n" FROM "##nchild"

Year	Country	n
1970-01-01 10:32:50	China	NULL
xyz	China	NULL
1970-01-01 10:33:00	China	NULL
1970-01-01 10:33:05	China	NULL
1970-01-01 10:33:10	China	NULL
1970-01-01 10:33:15	China	NULL
1970-01-01 10:33:20	China	NULL
1970-01-01 10:33:25	China	NULL
1970-01-01 10:33:30	China	NULL
```

```
SELECT "Year", "Country", TRY_CAST("n" AS BIGINT) AS "n" FROM "##nchild"

Year	Country	n
1970-01-01 10:32:50	China	NULL
xyz	China	NULL
1970-01-01 10:33:00	China	NULL
1970-01-01 10:33:05	China	NULL
1970-01-01 10:33:10	China	NULL
1970-01-01 10:33:15	China	NULL
1970-01-01 10:33:20	China	NULL
1970-01-01 10:33:25	China	NULL
1970-01-01 10:33:30	China	NULL
```

```
SELECT "Year", "Country", TRY_CAST("n" AS NUMERIC) AS "n" FROM "##nchild"

Year	Country	n
1970-01-01 10:32:50	China	NULL
xyz	China	4
1970-01-01 10:33:00	China	3
1970-01-01 10:33:05	China	3
1970-01-01 10:33:10	China	2
1970-01-01 10:33:15	China	2
1970-01-01 10:33:20	China	2
1970-01-01 10:33:25	China	2
1970-01-01 10:33:30	China	2
```

```
SELECT "Year", "Country", TRY_CAST(TRY_CAST("n" AS NUMERIC) AS INT) AS "n" FROM "##nchild"

Year	Country	n
1970-01-01 10:32:50	China	NULL
xyz	China	4
1970-01-01 10:33:00	China	3
1970-01-01 10:33:05	China	3
1970-01-01 10:33:10	China	2
1970-01-01 10:33:15	China	2
1970-01-01 10:33:20	China	2
1970-01-01 10:33:25	China	2
1970-01-01 10:33:30	China	2
```

hadley:
@DavidPatShuiFong it's odd that this isn't documented anywhere. Since this is such strange behaviour, would you mind making a simple reprex and hosting it on a gist? Something like this:

```R
library(dbplyr)

df <- data.frame(x = c("1", "2x"))
con <- DBI::dbConnect(...)
db <- copy_to(con, df, "##test")

db %>% mutate(
  float = TRY_CAST(x %as% sql("FLOAT")),
  integer = TRY_CAST(x %as% sql("INTEGER")),
  numeric = TRY_CAST(x %as% sql("NUMERIC")),
)
```

And I think it would be good to include a test of actual behaviour, maybe something like this:

```R
con <- src_test("mssql")
df <- data.frame(x = c("1", "2x"))
db <- copy_to(con, df, "##test")

out <- db %>% mutate(
  integer = as.integer(x),
  integer64 = as.integer64(x),
  numeric = as.numeric(x),
)
expect_identcal(out$integer, c(1L, NA))
expect_identcal(out$integer64, bit64::as.integer64(c(1L, NA)))
expect_identcal(out$numeric, c(1, NA))
```

(I didn't run this code because it's relatively hard for me to get a SQL server connection up and running but it should be close).

Thanks for all your work on this!

DavidPatShuiFong:
Thank you @hadley for the suggestions!

Here is a gist, which I have tested on MSSQL 2019 (15.0.4043 on Linux)

One peculiar thing I have found is that the failure to convert the entire column if there is a single invalid row *only* happens if the 'valid' rows contain non-integer values! Perhaps this is an edge case. But, in my case, if the data was dirty enough to contain garbage characters, it sure is dirty enough to contain non-integers.

https://gist.github.com/DavidPatShuiFong/7b47a9804a497b605e477f1bf6c38b37

```r
library(dbplyr)

df <- data.frame(x = c("1.3", "2x")) 
# both a 'float' (non-integer) and an 'invalid' number in the column
con <- DBI::dbConnect(...) # connection to a MSSQL database
db <- copy_to(con, df, "##test")

db %>% mutate(
  float = TRY_CAST(x %as% sql("FLOAT")),
  integer = TRY_CAST(x %as% sql("INTEGER")),
  bigint = TRY_CAST(x %as% sql("BIGINT")),
  numeric = TRY_CAST(x %as% sql("NUMERIC")),
  doublecast_integer = TRY_CAST(TRY_CAST(x %as% sql("NUMERIC")) %as% sql("INTEGER")),
  doublecast_bigint = TRY_CAST(TRY_CAST(x %as% sql("NUMERIC(38,0)")) %as% sql("BIGINT"))
)
```

results in

```
# Source:   lazy query [?? x 7]
# Database: Microsoft SQL Server 15.00.4043[dbo@localhost/master]
  x     float integer  bigint numeric doublecast_integer doublecast_bigint
  <chr> <dbl>   <int> <int64>   <dbl>              <int>           <int64>
1 1.3     1.3      NA      NA       1                  1                 1
2 2x     NA        NA      NA      NA                 NA                NA
```


> @DavidPatShuiFong it's odd that this isn't documented anywhere. Since this is such strange behaviour, would you mind making a simple reprex and hosting it on a gist? Something like this:
> 
> ```r
> library(dbplyr)
> library(dplyr, warn.conflicts = FALSE)
> 
> df <- data.frame(x = c("1", "2x"))
> con <- DBI::dbConnect(...)
> db <- copy_to(con, df, "##test")
> 
> db %>% mutate(
>   float = TRY_CAST(x %as% sql("FLOAT")),
>   integer = TRY_CAST(x %as% sql("INTEGER")),
>   numeric = TRY_CAST(x %as% sql("NUMERIC")),
> )
> ```

DavidPatShuiFong:
Here is a gist with the test of behaviour. Tested with MSSQL 2019 (on Linux) and with this 'pull request' revision of `dbplyr`.

Did you intend for these tests to be placed somewhere in the `testthat` section of `dbplyr`?

Thanks, David

https://gist.github.com/DavidPatShuiFong/4073a4fd50a7b10c97c1516f1a4d3e0e

```r
df <- data.frame(x = c("1.3", "2x"))
# dbplyr::test_register_src(
#   "mssql",
#   odbc::dbConnect(
#     odbc::odbc(), Driver = "ODBC Driver 17 for SQL Server",
#     Server = "<server>", UID = "<uid>", PWD = "<password>")
#   )
con <- src_test("mssql")
db <- copy_to(con, df, "##test")

out <- db %>% mutate(
  integer = as.integer(x),
  integer64 = as.integer64(x),
  numeric = as.numeric(x),
  ) %>%
  collect()

# testthat tests
expect_identical(out$integer, c(1L, NA))
expect_identical(out$integer64, bit64::as.integer64(c(1L, NA)))
expect_identical(out$numeric, c(1.3, NA))
```

> And I think it would be good to include a test of actual behaviour, maybe something like this:
> 
> ```r
> con <- src_test("mssql")
> df <- data.frame(x = c("1", "2x"))
> db <- copy_to(con, df, "##test")
> 
> out <- db %>% mutate(
>   integer = as.integer(x),
>   integer64 = as.integer64(x),
>   numeric = as.numeric(x),
> )
> expect_identcal(out$integer, c(1L, NA))
> expect_identcal(out$integer64, bit64::as.integer64(c(1L, NA)))
> expect_identcal(out$numeric, c(1, NA))
> ```

hadley:
I added the tests — thanks for working on this!

DavidPatShuiFong:
Results added to gist link, from both MSSQL 2019 and MSSQL 2014.

https://gist.github.com/DavidPatShuiFong/7b47a9804a497b605e477f1bf6c38b37

> Here is a gist, which I have tested on MSSQL 2019 (15.0.4043 on Linux)
> 
> One peculiar thing I have found is that the failure to convert the entire column if there is a single invalid row _only_ happens if the 'valid' rows contain non-integer values! Perhaps this is an edge case. But, in my case, if the data was dirty enough to contain garbage characters, it sure is dirty enough to contain non-integers.
> 
> https://gist.github.com/DavidPatShuiFong/7b47a9804a497b605e477f1bf6c38b37

PR #558, opened 2020-12-07
add verb `uncount()`

mgirlich:
As `uncount()` currently isn't a generic (also see https://github.com/tidyverse/tidyr/issues/1071) I named the function with `dbplyr_uncount()`.

I see two strategies for translating:

* use a recursive common table expression (see [this stackoverflow discussion](https://stackoverflow.com/questions/33327837/repeat-rows-n-times-according-to-column-value))
* or use an unequal join with a temporary table (the strategy I chose)

The recursive CTE would require much more code and also more testing across backends. Therefore, I simply chose the unequal join. I'm not sure how big the performance difference is.

There are also two variants to create the temporary table

* use the `generate_series()` function. Unfortunately, this is only available as an SQLite extension which is not included in  RSQLite.
* get the maximum value of the weights column and create a local tibble to upload.

The functionality is basically the same as the `tidyr` version with the small difference that the `weights` are not checked as strictly (i.e. no error if some weights are not non-negative integers).

Overview of the strategy:

1. get the max weight, `n_max`
2. create a temporary db table with ids from 1 to `n_max`
3. duplicate data by inner joining with this temporary table under the condition that `data$weight <= tmp$id`

The generated SQL looks like this

```SQL
SELECT `x`,
       `test`
FROM
  (SELECT `x`,
          `n`,
          `test`
   FROM `dbplyr_366` AS `LHS`
   INNER JOIN `dbplyr_367` AS `RHS` ON (`RHS`.`test` <= `LHS`.`n`))
```

One could simplify the SQL a little


R/verb-uncount.R
```diff
@@ -0,0 +1,73 @@
+#' "Uncount" a database table
+#'
+#' This is a method for the tidyr `uncount()` generic.
```
hadley, line 3: I think this needs to include a comment that it will uses a temporary table

R/verb-uncount.R
```diff
@@ -0,0 +1,73 @@
...
+    row_id_col <- .id
+  }
+  sql_on_expr <- expr(RHS[[!!row_id_col]] <= LHS[[!!weights_col]])
+
+  data_uncounted <- inner_join(
+    data,
+    tibble(!!row_id_col := seq2(1, n_max)),
+    by = character(),
+    sql_on = translate_sql(!!sql_on_expr, con = remote_con(data)),
+    copy = TRUE
```
hadley, line 53: I _think_ I'd like this a bit better if you explicitly created the temporary table, rather than relying on `copy = TRUE`

tests/testthat/test-verb-uncount.R
```diff
@@ -0,0 +1,48 @@
+test_that("symbols weights are dropped in output", {
```
hadley, line 1: Could you please include a snapshot test of the generated SQL?

hadley (commented):
Your summary of the approach in the PR body was super helpful; but I think it also needs to be included in the code as comments so that when looking at the code in the future it's easy to see the basic approach.

R/verb-uncount.R
mgirlich: Well spotted! I agree that it nicer to be explicit about copying

mgirlich:
The SQL snapshots don't work nicely together with the different backends. Depending on the active backend more or less temporary tables names are created, i.e. `dbplyr_table_name` has a different value. This is an issue for this snapshot test. For now I simply set the option manually as a workaround but it's not exactly the nicest workaround.
One could also force the backend tests to run later by changing their file name but that's not very nice either to not follow the normal conventions. Do you have a better idea?

hadley (approved):
Looks good! Just let me know how you want to merge these. If we wait and merge them all after they're all complete, I'm happy to handle the merge conflicts in the NEWS file on my end.

mgirlich:
I think prepared the other PRs to be ready to be merged. I guess it is easier and faster if you handle the merge conflicts on your side. Thanks!
By the way: great work with the new testthat. It is really nice and easy to work with the snapshots!

PR #560, opened 2020-12-07
`select()` informs when adding missing grouping variables

### LINKED ISSUE #559: select() should inform when adding grouping variables
The same way as it is done in dplyr

mgirlich:
Fixes #559 and #568


tests/testthat/test-verb-select.R
```diff
@@ -66,11 +66,16 @@ test_that("can rename with a function", {
 
 test_that("select preserves grouping vars", {
   mf <- memdb_frame(a = 1, b = 2) %>% group_by(b)
-  out <- mf %>% select(a) %>% collect()
+  expect_message(out <- mf %>% select(a) %>% collect())
```
hadley, line 69: Could you add a snapshot test (or possibly modify this test) to ensure that we have eyes on the generate message?

hadley:
Thanks! Just needs a news bullet now. It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

mgirlich:
You're welcome. Would be great if you find time to have a look at the other open pull requests.

NEWS.md
```diff
@@ -4,6 +4,10 @@
 
 * Add custom SQLite translation of `full_join()` and `right_join()` (#536, @mgirlich).
 
+* `select()` now does not relocate grouping variables to the front (#568, @mgirlich).
```
hadley, line 7: Sorry, I didn't spot this earlier, but our convention is name first, then issue number.

hadley:
I can try, but it's hard because I don't have any dbplyr context loaded in my brain right now.

PR #567, opened 2020-12-17
add verb fill()

### LINKED ISSUE #566: add support for `tidyr::fill()`

mgirlich:
closes #566.

Compared to `tidyr::fill()` I added an argument `order_by` as a deterministic order is required for the window functions used in the translation. With the extra argument and the use of dots in `tidyr::fill()` it also doesn't really fit as a generic (that's why it's called `dbplyr_fill`). It might be nicer to get rid of it by requiring a call to `arrange()` before. But this will then produce the warning `ORDER BY is ignored in subqueries without LIMIT`


R/verb-fill.R
```diff
@@ -0,0 +1,169 @@
...
+
+#' @export
+dbplyr_fill_int.MariaDBConnection <- dbplyr_fill_int.SQLiteConnection
+#' @export
+dbplyr_fill_int.MySQLConnection <- dbplyr_fill_int.SQLiteConnection
+#' @export
+dbplyr_fill_int.MySQL <- dbplyr_fill_int.SQLiteConnection
+
+#' @export
+dbplyr_fill_int.DBIConnection <- function(.con,
```
hadley, line 132: The default (DBIConnection) method should come first

R/verb-fill.R
```diff
@@ -0,0 +1,169 @@
...
+  dp %>%
+    transmute(
+      !!!syms(colnames(.data)),
+      !!!fill_sql
+    ) %>%
+    select(!!!colnames(.data))
+}
+
+#' @export
+dbplyr_fill_int.PostgreSQL <- dbplyr_fill_int.SQLiteConnection
```
hadley, line 114: I have mixed feelings about putting all these methods in this file. On one hand, it makes it easier to understand and review, but on the other hand, it makes it harder to see how to add a new backend, and doesn't follow the existing conventions. Maybe I should review like this, and then spread out after the fact?

R/verb-fill.R
```diff
@@ -0,0 +1,169 @@
+# ignore nulls
```
hadley, line 1: I think this would be easier for me to review if you could start with a high-level description of how the translation will work (for most databases). Here I think the key idea is using `LAST_VALUE` + `IGNORE NULLS` + the default frame (which goes from the first row in the group to the "current" row).

R/verb-fill.R
```diff
@@ -0,0 +1,169 @@
...
+  dbplyr_fill_int(
+    .con = remote_con(.data),
+    .data = .data,
+    cols_to_fill = cols_to_fill,
+    order_by_cols = order_by_cols,
+    .direction = .direction
+  )
+}
+
+dbplyr_fill_int <- function(.con, .data, cols_to_fill, order_by_cols, .direction) {
```
hadley, line 71: Why the `_int` prefix?

tests/testthat/_snaps/verb-fill.md
```diff
@@ -0,0 +1,57 @@
+# up-direction works
+
+    Code
+      dbplyr_fill(df_lazy_ns, n1, order_by = id, .direction = "up")
+    Output
+      <SQL>
+      SELECT `id`, `group`, MAX(`n1`) OVER (PARTITION BY `..dbplyr_partion_1`) AS `n1`
+      FROM (SELECT `id`, `group`, `n1`, SUM(CASE WHEN (((`n1`) IS NULL)) THEN (0) WHEN NOT(((`n1`) IS NULL)) THEN (1) END) OVER (ORDER BY -`id` ROWS UNBOUNDED PRECEDING) AS `..dbplyr_partion_1`
```
hadley, line 8: Why `-id`?

R/verb-fill.R
```diff
@@ -0,0 +1,169 @@
...
+#' )
+#' squirrels$id <- 1:12
+#'
+#' dbplyr_fill(
+#'   tbl_memdb(squirrels),
+#'   n_squirrels,
+#'   n_squirrels2,
+#'   order_by = c(id)
+#' )
+dbplyr_fill <- function(.data, ..., order_by, .direction = c("down", "up")) {
```
hadley, line 48: Could we use `window_order()` instead of `order_by`?

tests/testthat/_snaps/verb-fill.md
mgirlich: that's because of `.direction = "up"`

R/verb-fill.R
mgirlich: I explained both strategies in the respective function. I hope this helps in understanding it better.

R/verb-fill.R
mgirlich: yes, totally makes sense. The easier understanding was the reason I started it this way and then I forgot to split it into the different backend files...

R/verb-fill.R
mgirlich: There is the user facing function `dbplyr_fill()`. It does some basic work and then it calls a separate function `dbplyr_fill_int()` that dispatches on the type of connection. I thought of the `_int` suffix as standing for "internal". Does this make sense to you or do you prefer a different name?

R/verb-fill.R
mgirlich: Do you mean renaming the argument `order_by` to `window_order`? If so, yes, `window_order` might make more sense.

R/verb-fill.R
hadley: Ohhhhhh, `_int` standards for `integer` to me, and we definitely want to avoid that. We don't have a strong existing convention for an internal suffix, but we been using 0 later (e.g. `dplyr_fill0()`)

R/verb-fill.R
hadley: No I mean assume the user has used that function to set the ordering.

mgirlich: hmm that didn't work in my first approach (it produced a warning about `ORDER BY` being ignored in subqueries) but it actually works now. Nice :-)

R/verb-fill.R
hadley: Looks good — can you please move this comment down close to `fill0` so it's closer to the implementations?

R/verb-fill.R
```diff
@@ -0,0 +1,195 @@
...
+  cols_to_fill <- syms(names(tidyselect::eval_select(expr(c(...)), sim_data)))
+  order_by_cols <- op_sort(.data)
+  .direction <- arg_match0(.direction, c("down", "up"))
+
+  if (is_empty(order_by_cols)) {
+    abort(
+      c(
+        x = "`.data` does not have explicit order.",
+        i = "Please use arrange() or window_order() to make determinstic."
+      )
```
hadley, line 58: Can you please test this using a combination of `expect_error()` and `expect_snapshot()`? The pattern we're currently using looks something like this:

```R
expect_snapshot({
  (expect_error(df %>% fill(x))
})`

`expect_error()` ensures that an error is actually thrown, and `expect_snapshot()` verifies that it's the correct error, and gives a place to proof read the error as it will appear to the user.

File: R/verb-fill.R [hunk 8]
```diff
@@ -0,0 +1,195 @@
...
+    )
+  }
+
+  if (.direction == "up") {
+    # `desc()` cannot be used here because one gets invalid SQL if a variable
+    # is already wrapped in `desc()`
+    # i.e. `desc(desc(x))` produces `x DESC DESC`
```
hadley, line 65: I think the downside is that this won't work for non-numeric variables, but I don't see how to work around that currently. So I think leave the implementation as is, just mention the limitation in a comment.

R/verb-fill.R
```diff
@@ -0,0 +1,195 @@
...
+
+  .data %>%
+    transmute(
+      !!!syms(colnames(.data)),
+      !!!fill_sql
+    )
+}
+
+#' @export
+dbplyr_fill0.SQLiteConnection <- function(.con,
```
hadley, line 118: Just mention somewhere that this is a fall back for databases that don't have IGNORE NULLs support?

hadley (approved):

PR #571, opened 2020-12-30
fixing a typo ('dbQuoteIdentifer' -> 'dbQuoteIdentifier')

yitao-li:
I only found this typo after copy/pasting things from https://dbplyr.tidyverse.org/articles/backend-2.html to sparklyr source code (because sparklyr needs to be up-to-date with all the changes from `dbplyr` edition 2). Let's get this typo corrected so that other backend authors won't need to go through the same trouble.


hadley:
Thanks!

PR #580, opened 2021-01-25
Improve translation of substr() & str_sub()

### LINKED ISSUE #577: Poor substr() translation
Hi,
I was trying to create an SQL query which would grab the end of a string. In this example it was supposed to be the last five characters. However, the `show_query` function shows it returns `null` rather than calculated integers.
Is that a correct behaviour?

``` r
library(RSQLite)
library(DBI)
library(dplyr)
library(dbplyr)

# Create example SQL database and load with example table
conn <- DBI::dbConnect(RSQLite::SQLite(), "")
DBI::dbWriteTable(conn, 
                  "iris", 
                  iris %>% distinct(Species), 
                  overwrite = TRUE)

# Expected behaviour
DBI::dbGetQuery(conn,
                "SELECT `Species`
                        ,LENGTH(`Species`) AS `Length`
                        ,SUBSTR(`Species`, LENGTH(`Species`)-4, LENGTH(`Species`)) AS `Trimmed`
                FROM `iris`")
#>      Species Length Trimmed
#> 1     setosa      6   etosa
#> 2 versicolor     10   color
#> 3  virginica      9   inica

# Actual code example; start and stop substituted with NULL
tbl(conn, DBI::Id(Table = "iris")) %>%
    mutate(Length = length(Species),
           Trimmed = substr(Species, start = length(Species)-4, stop = length(Species))) %T>%
    show_query() %>%
    collect()
#> Warning in substr(Species, start = length(Species) - 4, stop = length(Species)):
#> NAs introduced by coercion
#> Warning in pmax(as.integer(stop) - start + 1L, 0L): NAs introduced by coercion
#> <SQL>
#> SELECT `Species`, length(`Species`) AS `Length`, SUBSTR(`Species`, NULL, NULL) AS `Trimmed`
#> FROM `iris`
#> Warning in substr(Species, start = length(Species) - 4, stop = length(Species)):
#> NAs introduced by coercion

#> Warning in substr(Species, start = length(Species) - 4, stop = length(Species)):
#> NAs introduced by coercion
#> # A tibble: 3 x 3
#>   Species    Length Trimmed
#>   <chr>       <int> <lgl>  
#> 1 setosa          6 NA     
#> 2 versicolor     10 NA     
#> 3 virginica       9 NA
```

<sup>Created on 2021-01-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I've tried using `nchar` and `len` instead of R `length` but with the same effect (My actual database uses `len` rather than `length`).

Thanks

Edit: Provided a better example showing that using native SQL code gets desired result

hadley:
The handling of negative start seems to be inconsistent from database to database, so I think it's safer to just always use LENGTH(). New optional_length parameter allows for MS SQL compatibility.

Fixes #577

@ianmcook would you mind checking my logic? It didn't change the existing unit tests that go through SQLite, so I think it's safe, but I am just terrible about reasoning through this sort of computation.


ianmcook:
@hadley setting aside `str_sub()` for the moment:

 Do you intend for `substr()` and `substring()` to behave like `SUBSTR()` and `SUBSTRING()` do in SQL, or like `substr()` and `substring()`do in base R? In this implementation, they behave more like their SQL variants:
- They accept negative start (like the SQL functions typically do, but the R functions do not)
- They are aliases of the same implementation (like is the norm in SQL, whereas the R functions differ with respect to the third argument)

However they also accept negative values for the third argument, which is typically not allowed in SQL.

If the goal is for these two functions to behave like SQL, perhaps it would be better to not attempt translation at all, and just pass the expressions through into the SQL. That way users can use column references and expressions in the second and third arguments if they like.

If the goal is for them to behave like base R, then I would recommend using a much simpler implementation that does not attempt to handle negative values in the second and third arguments.

I think implementing `str_sub()` separately from these will keep things much cleaner and simpler.

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # dbplyr (development version)
 
+* `str_sub()` and `substring()` get better translations when start is negative,
+  and work on MS SQL when `end = -1` (#577).
```
ianmcook, line 4: It might be good to change the title of #577 so folks don't get the mistaken idea that these functions now support dynamic start and stop

hadley:
The goal is to match the behaviour of the R functions (which I always forget). It sounds like I should set `start` to 1 if its negative (that appears to be what `substr()` does). I'm not too worried about the third argument since `substring()` (I think) is relatively rarely used, but I could copy the R default if you think it's worth it.

So maybe something more like this:

```R
sql_substr <- function(f = "SUBSTR") {
  function(x, start, stop = 1000000L) {
    start <- max(check_integer(start, "start"), 1)
    stop <- max(check_integer(stop, "stop"), 1)
    length <- stop - start + 1L

    sql_call2(f, x, start, length)
  }
}

check_integer <- function(x, arg) {
  if (length(x) != 1 || !is.numeric(x)) {
    abort(paste("`", arg, "` must be a single number"))
  }
  as.integer(x)
}
```

?

ianmcook:
Yep. Or you could let the SQL engine do the logic for you by doing translations like this:
- `substr(x, start)` --> `SUBSTR(x, GREATEST(1, start))`
- `substring(x, start)` --> `SUBSTR(x, GREATEST(1, start), LEAST(1000000, stop))`
- `substr(x, start, stop)` or `substring(x, start, stop)` --> `SUBSTR(x, GREATEST(1, start), GREATEST(1, stop))`

That would enable folks to use column references or expressions for `start` and `stop` (not just numeric literals).

However, then we need to handle SQL dialects differently:
- PostgreSQL, MySQL, Hive, Impala, Hana, Oracle, Redshift, Teradata use `LEAST()` and `GREATEST()`
- SQLite uses multi-argument `MIN()` and `MAX()` (not aggregate functions when used in that context!)
- SQL Server, Access don't have a function for this so you need something like `IIF(a < b, a, b)` and `IIF(a > b, a, b)`

For `str_sub()` it's probably not possible/reasonable to express the logic all in SQL expressions like that, but `substr()` and `substring()` are simple enough to do it that way.

hadley:
There's currently no way to rely on existing translations so this would require additional arguments, which seems like quite a lot of work for only moderate gain, so I think I'll just keep it simple for now.

PR #585, opened 2021-01-27
support argument `.groups` in `summarise()`

### LINKED ISSUE #584: Inconsistent behaviour between dplyr::summarise and dbplyr::summarise with .groups argument
In dplyr 1.0.0 the `.groups` argument was added to `summarise` to indicate what should happen with a grouping consisting of multiple variables. Unfortunately, this special argument does not seem to be recognized by the `dbplyr::summarise` implementation. Adding it results in a column `.groups` being added to the output.

This makes it difficult to write warning-free, generic code in the sense that it can take either a data.frame (or tibble) or an object returned from `tbl`.

<details>
The below code is a short illustration of the problem (using SQLite for testing purposes, the problem also exists for different databases/connectors)

```r
library(DBI)
library(RSQLite)
library(dplyr)

test <- tibble(a = c("a", "a", "b", "b", "c", "c", "c", "d"), b = c(1,2,1,1,3,3,4,1), y = runif(8))

con <- dbConnect(SQLite())
copy_to(con, name = "test", test)

my_summary <- function(df) {
  df %>%
    group_by(a, b) %>%
    summarise(y_ = min(y, na.rm = FALSE), .groups = "drop") %>%
    collect(n=Inf)
}

print(my_summary(test))
print(my_summary(tbl(con, "test")))
```

The first call to `my_summary` yields something like
```
# A tibble: 6 x 3
  a         b     y_
* <chr> <dbl>  <dbl>
```

while the second one returns

```
# A tibble: 6 x 4
# Groups:   a [4]
  a         b     y_ .groups
  <chr> <dbl>  <dbl> <chr>
```

with the `.groups` column being set to `"drop"`. Expected behaviour: both calls should return a `tibble` with the same schema.

In a way I suspect that there were always inconsistencies here (i.e. grouping dropping completely in dbplyr/SQL after a summarise vs. only the last variable being removed in dplyr) that are hard to fix, but recognizing `.groups` on the dbplyr side would be a nice consistency improvement.

</details>

mgirlich:
closes #584 

I am not sure about the environment check in `summarise_verbose()`.

### DIFF
--- a/R/verb-summarise.R
+++ b/R/verb-summarise.R
@@ -12,7 +12,6 @@
 #'   only supported option before version 1.0.0.
 #'   * "drop": All levels of grouping are dropped.
 #'   * "keep": Same grouping structure as `.data`.
-#'   * "rowwise": This is not possible for lazy data frames.
 #'
 #'   When `.groups` is not specified, it defaults to "drop_last".
 #'
@@ -38,16 +37,7 @@ summarise.tbl_lazy <- function(.data, ..., .groups = NULL) {
   dots <- quos(..., .named = TRUE)
   dots <- partial_eval_dots(dots, vars = op_vars(.data))
   check_summarise_vars(dots)
-
-  if (!is_null(.groups) && !.groups %in% c("drop_last", "drop", "keep")) {
-    abort(c(
-      paste0(
-        "`.groups` can't be ", as_label(.groups),
-        if (.groups == "rowwise") " for lazy tables"
-      ),
-      i = 'Possible values are NULL (default), "drop_last", "drop", and "keep"'
-    ))
-  }
+  check_groups(.groups)
 
   add_op_single("summarise", .data, dots = dots, args = list(.groups = .groups))
 }
@@ -69,18 +59,16 @@ check_summarise_vars <- function(dots) {
   }
 }
 
-summarise_verbose <- function(.groups, .env) {
-  is.null(.groups) &&
-    # TODO not sure whether this is correct
-    is_reference(asNamespace("dbplyr"), topenv(.env)) &&
-    # is_reference(topenv(.env), global_env()) &&
-    !identical(getOption("dplyr.summarise.inform"), FALSE)
-}
-
-summarise_inform <- function(..., .env = parent.frame()) {
-  inform(paste0(
-    "`summarise()` ", glue(..., .envir = .env), '. You can override using the `.groups` argument.'
-  ))
+check_groups <- function(.groups) {
+  if (!is_null(.groups) && !.groups %in% c("drop_last", "drop", "keep")) {
+    abort(c(
+      paste0(
+        "`.groups` can't be ", as_label(.groups),
+        if (.groups == "rowwise") " for lazy tables"
+      ),
+      i = 'Possible values are NULL (default), "drop_last", "drop", and "keep"'
+    ))
+  }
 }
 
 #' @export
@@ -93,18 +81,11 @@ op_grps.op_summarise <- function(op) {
   grps <- op_grps(op$x)
   .groups <- op$args$.groups %||% "drop_last"
 
-  if (identical(.groups, "drop_last")) {
-    n <- length(grps)
-    if (n == 1) {
-      character()
-    } else {
-      grps[-n]
-    }
-  } else if (identical(.groups, "keep")) {
-    grps
-  } else if(identical(.groups, "drop")) {
-    character()
-  }
+  switch(.groups,
+    drop_last = grps[-length(grps)],
+    keep = grps,
+    drop = character()
+  )
 }
 
 #' @export
@@ -122,7 +103,7 @@ sql_build.op_summarise <- function(op, con, ...) {
   # -> use `drop_last` as it was historically
   .groups <- .groups %||% "drop_last"
 
-  if (verbose && identical(.groups, "drop_last") && n > 1) {
+  if (verbose && n > 1) {
     new_groups <- glue::glue_collapse(paste0("'", group_vars[-n], "'"), sep = ", ")
     summarise_inform("has grouped output by {new_groups}")
   }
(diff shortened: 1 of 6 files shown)


R/verb-summarise.R
```diff
@@ -6,6 +6,19 @@
 #'
 #' @inheritParams arrange.tbl_lazy
 #' @inheritParams dplyr::summarise
+#' @param .groups \Sexpr[results=rd]{lifecycle::badge("experimental")} Grouping structure of the result.
+#'
+#'   * "drop_last": dropping the last level of grouping. This was the
+#'   only supported option before version 1.0.0.
+#'   * "drop": All levels of grouping are dropped.
+#'   * "keep": Same grouping structure as `.data`.
+#'   * "rowwise": This is not possible for lazy data frames.
```
hadley, line 15: I think you could just omit this row

R/verb-summarise.R
```diff
@@ -21,11 +34,22 @@
...
 #'   summarise(n()) %>%
 #'   show_query()
-summarise.tbl_lazy <- function(.data, ...) {
+summarise.tbl_lazy <- function(.data, ..., .groups = NULL) {
   dots <- quos(..., .named = TRUE)
   dots <- partial_eval_dots(dots, vars = op_vars(.data))
   check_summarise_vars(dots)
-  add_op_single("summarise", .data, dots = dots)
+
+  if (!is_null(.groups) && !.groups %in% c("drop_last", "drop", "keep")) {
```
hadley, line 42: Maybe move to `check_groups()`?

R/verb-summarise.R
```diff
@@ -66,8 +113,21 @@ op_sort.op_summarise <- function(op) NULL
 #' @export
 sql_build.op_summarise <- function(op, con, ...) {
   select_vars <- translate_sql_(op$dots, con, window = FALSE, context = list(clause = "SELECT"))
-  group_vars <- c.sql(ident(op_grps(op$x)), con = con)
+  group_vars <- op_grps(op$x)
+  n <- length(group_vars)
+
+  .groups <- op$args$.groups
+  verbose <- summarise_verbose(.groups, caller_env())
+  # we cannot automatically infer "keep" as for local frames
```
hadley, line 121: Why would we infer `keep`? I think I've forgotten something.

R/verb-summarise.R
```diff
@@ -45,6 +69,20 @@ check_summarise_vars <- function(dots) {
   }
 }
 
+summarise_verbose <- function(.groups, .env) {
```
hadley, line 72: Should live below `sql_build.op_summarise()`

R/verb-summarise.R
```diff
@@ -45,6 +69,20 @@ check_summarise_vars <- function(dots) {
...
+  is.null(.groups) &&
+    # TODO not sure whether this is correct
+    is_reference(asNamespace("dbplyr"), topenv(.env)) &&
```
hadley, line 75: I think it should be `is_reference(topenv(.env), global_env())` as in dplyr — the goal is to detect whether or not this is called directly by the user (in which case it will be somewhere beneath the global environment), or indirect via a package (in which is case it will be beneath a package namespace).

R/verb-summarise.R
```diff
@@ -53,10 +91,19 @@ op_vars.op_summarise <- function(op) {
 #' @export
 op_grps.op_summarise <- function(op) {
   grps <- op_grps(op$x)
-  if (length(grps) == 1) {
+  .groups <- op$args$.groups %||% "drop_last"
+
+  if (identical(.groups, "drop_last")) {
+    n <- length(grps)
+    if (n == 1) {
```
hadley, line 98: Do you need the if here? I think `grps[-n]` will work even if `grps` is length 1 (or even length 0).

I'd also suggest writing this as a switch:

```R
switch(.groups, 
  drop_last = grps[-length(grps)],
  keep = grps,
  drop = chacracter()
)
```

That makes it more clear that you're dealing with a small set of possible options.

R/verb-summarise.R
```diff
@@ -66,8 +113,21 @@ op_sort.op_summarise <- function(op) NULL
...
+  group_vars <- op_grps(op$x)
+  n <- length(group_vars)
+
+  .groups <- op$args$.groups
+  verbose <- summarise_verbose(.groups, caller_env())
+  # we cannot automatically infer "keep" as for local frames
+  # -> use `drop_last` as it was historically
+  .groups <- .groups %||% "drop_last"
+
+  if (verbose && identical(.groups, "drop_last") && n > 1) {
```
hadley, line 125: I think you could move the `identical(.groups, "drop_last")` check into `summarise_verbose()`?

hadley (commented):
Thanks for working on this!

R/verb-summarise.R
mgirlich: It looks like `grps[-n]` works I just didn't use it as currently it is done explicitly like that

```r
if (length(grps) == 1) {
  character()
} else {
  grps[-length(grps)]
}
```

But I don't see a reason why it shouldn't work as `op_grps()` should always return a character.

R/verb-summarise.R
mgirlich: From the documentation of `summarise`:

> When .groups is not specified, it is chosen based on the number of rows of the results:
> * If all the results have 1 row, you get "drop_last".
> * If the number of rows varies, you get "keep".

For local dataframes the check is easy but for db tables we cannot easily check that (I think it is not even possible that multiple rows are returned per group but).

R/verb-summarise.R
mgirlich: ah, that makes sense. I just had to catch the environment earlier on to make it work.

R/verb-summarise.R
hadley: Yeah, in databases `summarise()` always returns a single row.

R/verb-summarise.R
```diff
@@ -45,6 +64,24 @@ check_summarise_vars <- function(dots) {
   }
 }
 
+check_groups <- function(.groups) {
```
hadley, line 67: This was fine before, but it seemed like a good example for https://github.com/tidyverse/design/issues/124

Issue #574, opened 2021-01-15
COPY = TRUE is creating a global temporary table of some kind. Not a session specific temporary table.


ben1787:
This seems like a bug in the new version of dbplyr.

When I use copy = TRUE, or the underlying copy_to(..., temporary = TRUE) the table is not very temporary. Even if I create a new data connection to the db the table is still there. 

First I am posting the reprex using dbplyr 1.4.4. Below that I will paste the reprex with dbplyr 2.0.0:
``` r
library(DBI)
library(RPostgreSQL)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(dbplyr)
#> 
#> Attaching package: 'dbplyr'
#> The following objects are masked from 'package:dplyr':
#> 
#>     ident, sql
ch = dbConnect(drv = dbDriver("PostgreSQL"),
                dbname = db,
                host = host,
                port = port,
                user = username,
                password = pwd)

tickers_hedged = data.frame(factset_id='No Hedged Tickers', date = Sys.Date(), instrument_type = as.character(NA))
tickers_hedged_temp_table = paste('tickers_hedged',format(Sys.time(),format = '%Y%m%d%H%M%OS6'),sep='_')
copy_to(dest = ch, df = tickers_hedged, name = tickers_hedged_temp_table, temporary = TRUE)
tbl(ch, tickers_hedged_temp_table)
#> # Source:   table<tickers_hedged_20210120005010.614228> [?? x 3]
#> # Database: postgres 11.0.7
#>   factset_id        date       instrument_type
#>   <chr>             <date>     <chr>          
#> 1 No Hedged Tickers 2021-01-20 <NA>

dbDisconnect(ch)
#> [1] TRUE


ch2 = dbConnect(drv = dbDriver("PostgreSQL"),
                dbname = db,
                host = host,
                port = port,
                user = username,
                password = pwd)
tbl(ch2, tickers_hedged_temp_table)
#> Error in postgresqlExecStatement(conn, statement, ...): RS-DBI driver: (could not Retrieve the result : ERROR:  relation "tickers_hedged_20210120005010.614228" does not exist
#> LINE 1: SELECT * FROM "tickers_hedged_20210120005010.614228" AS "zzz...
#>                       ^
#> )
#expected behavior
```

[...]

PR #589, opened 2021-02-04
fix typo

dpprdan:
Well, I _think_ this is a typo from reading #574, in particular this https://github.com/tidyverse/dbplyr/issues/574#issuecomment-764955038


hadley:
Good catch, thanks!

PR #609, opened 2021-03-08
Add escapes for dates and datetimes for MS Access

### LINKED ISSUE #608: Dates in MS Access aren't escaped properly.
Escaping dates when using a connection to MS Access goes wrong

Access uses octothorpes to delimit dates, and requires dates to be formatted either YYYY-MM-DD or MM/DD/YYYY. Optionally with hh:mm:ss as time, timezones are not supported.

```r
escape(Sys.Date(), con = simulate_access())
# Expected: #2021-03-08#, or: #03/08/2021#
# Returned: '2021-03-08'
escape(Sys.time(), con = simulate_access())
# Expected: #2021-03-08 09:30:30#
# Returned: '2021-03-08T09:30:30Z'
```
 The relevant docs are [here](https://support.microsoft.com/en-us/office/examples-of-using-dates-as-criteria-in-access-queries-aea83b3b-46eb-43dd-8689-5fc961f21762#__toc358028377) concerning how to use dates in queries in MS Access.

To work around the issue, I currently use the following function. However, it doesn't handle collapse/parentheses, only single values.

```r
escape_access <- function(x){
  if(inherits(x, "Date")){
    return(sql(ifelse(is.na(x), "NULL", format(x, "#%F#"))))
  }
  else if(inherits(x, "POSIXt")){
    return(sql(ifelse(is.na(x), "NULL", format(x, "#%F %T#"))))
  }else{
    escape(x, con = simulate_access())
  }
}

```

erikvona:
Closes #608


R/backend-access.R
```diff
@@ -167,5 +167,22 @@ sql_escape_logical.ACCESS <- function(con, x) {
   y[is.na(x)] <- "NULL"
   y
 }
+                   
+#' @export                   
+sql_escape_date.ACCESS <-  function(con, x) {
+  # Access delimits dates using octothorpes, and uses YYYY-MM-DD
+  y <- format(x, "#%F#")
```
hadley, line 174: Can you please spell this out as `%Y-%m-%d`

R/backend-access.R
```diff
@@ -167,5 +167,22 @@ sql_escape_logical.ACCESS <- function(con, x) {
...
+  y[is.na(x)] <- "NULL"
+  y
+}
+
+#' @export
+sql_escape_datetime.ACCESS <-  function(con, x) {
+  # Access delimits datetimes using octothorpes, and uses YYYY-MM-DD HH:MM:SS
+  # Timezones are not supported in Access
+  y <- format(x, "#%F %T#")
```
hadley, line 183: What does not supporting time zones imply here? Should we always generate a time in the system time zone?

hadley (commented):

R/backend-access.R
erikvona: Of course, similar change made for `sql_escape_datetime`

R/backend-access.R
erikvona: It means timezone information is ignored when you escape a value that includes timezone information, e.g. `escape(as.POSIXct('2020-01-01 01:01:01', tz = "US/Alaska"), con = simulate_access())` will result in the same thing as `escape(as.POSIXct('2020-01-01 01:01:01', tz = "Europe/Berlin"), con = simulate_access())`. 

Some RDBMSes have a concept of timezones (e.g. SQL Server, Oracle) and when escaping a datetime field the timezone could be included, however, Access does not. 

I assume that's a non-issue, especially since escapes for those databases (e.g.  `escape(as.POSIXct('2020-01-01 01:01:01', tz = 'US/Alaska'), con = simulate_mssql())`) always provides the zulu timezone and also ignore timezone information.  But that's a separate issue, and one that's considerably harder to address and might break existing code when addressed.

hadley: Ah, got it. Thanks for the explanation!

PR #625, opened 2021-03-26
Adds support for Snowflake databases

edgararuiz:
- Adds a formal Snowflake "back-end" script 
- Customizes `log10()` translation 
- Adds simulate function
- Prevent ANALYZE TABLE to run as part of `copy_to()` by creating a method with an empty output
- Adds single test for `log10()` SQL output


edgararuiz:
@hadley PTAL

R/backend-snowflake.R
```diff
@@ -0,0 +1,36 @@
...
+      base_win
+  )
+}
+
+#' @export
+#' @rdname backend-snowflake
+simulate_snowflake <- function() simulate_dbi("Snowflake")
+
+#' @export
+sql_table_analyze.Snowflake <- function(con, table, ...) {}
```
hadley, line 36: Could you add a comment here explaining why, preferably with a link to snowflake docs?

edgararuiz: This was like trying to prove a negative.  Could not find an explicit acknowledgement in the docs that we don't need to run the table analysis.  I commented and included the link to *all* SQL commands, hopefully that helps.

hadley (approved):

PR #632, opened 2021-04-07
`quantile()` can use `na.rm`

### LINKED ISSUE #600: Error: Ignore na.rm = TRUE in the quantile function
While creating a shiny application that reads both local files and database queries, I found a problem with the quantile function and PostgreSQL.

If na.rm is not present in the quantile function for local data, the code would not work as intended. If na.rm is set to try it works perfectly. However, if na.rm is set to true and a database table is passed on then it errors out.

If na.rm could just be ignored for tbl_lazy then it would work perfectly.

```r
tbl_lazy(mtcars) %>%
  summarise(q = quantile(mpg, 0.5, na.rm = TRUE)
```

Output: 
```r
<SQL>
Error in quantile(mpg, 0.5, na.rm = TRUE) : 
  unused argument (na.rm = TRUE)
```

mgirlich:
Fixes #600.

@krlmlr As discussed I created a [milestone](https://github.com/tidyverse/dbplyr/milestone/6) with relatively simple PRs.


tests/testthat/test-translate-sql-quantile.R
```diff
@@ -1,14 +1,14 @@
...
+  expect_snapshot(translate_sql(quantile(x, 0.75, na.rm = TRUE), vars_group = "g"))
   expect_snapshot(translate_sql(median(x, na.rm = TRUE), window = FALSE))
   expect_snapshot(translate_sql(median(x, na.rm = TRUE), vars_group = "g"))
 })
 
-test_that("median functions warn once if na.rm = FALSE", {
-  expect_warning(translate_sql(median("x")), "Missing values")
-  expect_warning(translate_sql(median("x")), NA)
-  expect_warning(translate_sql(median("x", na.rm = TRUE)), NA)
+test_that("quantile functions warn once if na.rm = FALSE", {
```
krlmlr, line 8: Did you mean to remove the test for `median()` ?

krlmlr (commented):
Looks good, thanks!

tests/testthat/test-translate-sql-quantile.R
mgirlich: `median` is translated with `sql_median()` which is basically just a wrapper around `sql_quantile()`. So, replacing the tests for `median()` with tests for `quantile()` should be okay.

PR #633, opened 2021-04-07
fix `compute(temporary = TRUE)` for Oracle

### LINKED ISSUE #621: Compute() with Oracle - missing keyword GLOBAL
Hi,

First of all, thank you for your work. I think there's a bug with the command `compute()` when used with Oracle connection.

```r
library(ROracle)
library(dplyr)
drv <- dbDriver("Oracle")
conn <- dbConnect(drv, dbname = "DATABASE")
tbl(conn, "table") %>% compute
```
Returns this error 

```r
Error in .oci.SendQuery(conn, statement, data = data, prefetch = prefetch,  : 
ORA-14459: missing GLOBAL keyword
```

It's exactly the same that happens if you do `dbExecute(conn, 'CREATE TEMPORARY TABLE table_name(id NUMBER)')` instead of `dbExecute(conn, 'CREATE GLOBAL TEMPORARY TABLE table_name(id NUMBER)')` which works fine. [See other people concerned](https://stackoverflow.com/questions/49635646/how-to-use-dbplyr-to-store-a-query-in-an-oracle-temporary-table-without-ever-sto).

mgirlich:
Fixes #621


R/backend-oracle.R
```diff
@@ -114,6 +114,15 @@ sql_query_wrap.Oracle <- function(con, from, name = unique_subquery_name(), ...)
   }
 }
 
+#' @export
+sql_query_save.Oracle <- function(con, sql, name, temporary = TRUE, ...) {
+  build_sql(
+    "CREATE ", if (temporary) sql("GLOBAL TEMPORARY "), "TABLE \n",
+    as.sql(name, con), " AS ", sql,
```
krlmlr, line 121:
```suggestion
    "CREATE ", if (temporary) sql("GLOBAL TEMPORARY "), "TABLE\n",
    as.sql(name, con), " AS\n", sql,
```

krlmlr (commented):
Thanks, I wonder if we should be more generous with newlines.

R/backend-oracle.R
mgirlich: It improves readability a bit so I changed it here and for the other backends 👍

PR #637, opened 2021-04-13
Join custom alias

mgirlich:
* add `lhs_as` and `rhs_as` to join-verbs
* pass it along
* wrapping `x` and `y` in subquery moves from `sql_render()` to `sql_query_join()` resp. `sql_query_semi_join()` because they gain `lhs_as` as argument


krlmlr (commented):
I really love the idea of giving the LHS and RHS custom names. Two remarks:

- The patch is relatively large, because the new `_as` arguments are passed along everywhere. Can we smuggle `lhs_as` and `rhs_as` somewhere in the existing arguments, e.g. in `names(x)` or `by` ?
- Can we use a smart default for `lhs_as` and `rhs_as`, e.g. `deparse(substitute(x)))` ? This will be somewhat tricky with the generic but might be worthwhile.

mgirlich:
> * The patch is relatively large, because the new `_as` arguments are passed along everywhere. Can we smuggle `lhs_as` and `rhs_as` somewhere in the existing arguments, e.g. in `names(x)` or `by` ?

`names(x)` feels quite indirect and not that safe so I've gone via `by`.

> * Can we use a smart default for `lhs_as` and `rhs_as`, e.g. `deparse(substitute(x)))` ? This will be somewhat tricky with the generic but might be worthwhile.

Hmm, I think this only makes sense if `x` or `y` is a symbol.

mgirlich:
Yes, I think that is the only advantage. But I think that can help readability a lot.

R/backend-sqlite.R
```diff
@@ -141,14 +141,31 @@ sql_query_join.SQLiteConnection <- function(con, x, y, vars, type = "inner", by
   # workaround as SQLite doesn't support FULL OUTER JOIN and RIGHT JOIN
   # see: https://www.sqlite.org/omitted.html
 
-  # Careful: in the right join resp. the second join in the full join do not
-  # name `y` LHS and `x` RHS as it messes up the select query!
+  if (!type %in% c("full", "right")) {
```
hadley, line 144: I think it's better to be explicit about which joins it applies to, not which joins it doesn't.

R/verb-joins.R
```diff
@@ -68,7 +70,8 @@ NULL
 inner_join.tbl_lazy <- function(x, y, by = NULL, copy = FALSE,
                                 suffix = NULL,
                                 auto_index = FALSE, ...,
-                                sql_on = NULL, na_matches = c("never", "na")) {
+                                sql_on = NULL, na_matches = c("never", "na"),
+                               lhs_as = "LHS", rhs_as = "RHS") {
```
hadley, line 74: Use `x_as` and `y_as`  for consistency with first arguments? Or would it be better to have `as = c("LHS", "RHS")`? Or `as_x` and `as_y`?

tests/testthat/_snaps/backend-sqlite.md
```diff
@@ -22,10 +22,10 @@
       LEFT JOIN `df` AS `RHS`
       ON (`LHS`.`x` = `RHS`.`x`)
       UNION
-      SELECT COALESCE(`LHS`.`x`, `RHS`.`x`) AS `x`, `LHS`.`y` AS `y.x`, `RHS`.`y` AS `y.y`, `z`
+      SELECT COALESCE(`RHS`.`x`, `LHS`.`x`) AS `x`, `LHS`.`y` AS `y.x`, `RHS`.`y` AS `y.y`, `z`
```
hadley, line 25: Should this have changed?

mgirlich: This does not really matter because a) they are equal (see the `ON`) and b) the `COALESCE()` is not really needed here but rather an artifact from simulating the `full_join()`. As the result should be correct it should not matter that the query is a bit more complicated than necessary.

R/verb-joins.R
mgirlich: Having only one argument saves a bit of typing but I prefer having two arguments. To me `x_as` and `y_as` read the most natural.

hadley: I have mild preference towards making the common part of a name the prefix (since it helps with autocomplete), but that doesn't make much difference here. If you prefer `x_as` and `y_as` then we can stick with that.

tests/testthat/_snaps/backend-sqlite.md
hadley: I'm just slightly worried that this indicates we're now preferring the RHS of the join over the LHS in full joins?

mgirlich: Hmm, I understand your concern but this was also the case before but just hidden because lazy frames are always named `df`. Before the second join query was actually

```sql
  SELECT
    COALESCE(`LHS`.`x`, `RHS`.`x`) AS `x`,
    `LHS`.`y` AS `y.x`,
    `z1`,
    `RHS`.`y` AS `y.y`,
    `z2`
  FROM `df2` AS `LHS`
  LEFT JOIN `df1` AS `RHS`
    ON (`LHS`.`x` = `RHS`.`x`)
```

So, `df2` was `LHS` instead of `RHS`. Now this is

```sql
SELECT
    COALESCE(`RHS`.`x`, `LHS`.`x`) AS `x`,
    `LHS`.`y` AS `y.x`,
    `z1`,
    `RHS`.`y` AS `y.y`,
    `z2`
  FROM `df2` AS `RHS`
  LEFT JOIN `df1` AS `LHS`
    ON (`RHS`.`x` = `LHS`.`x`)
```

Changing the `COALESCE` would basically require an extra argument to `sql_join_vars()` or writing a custom left join query method for SQLite. I don't think this is worth it...

hadley (approved):

PR #642, opened 2021-04-14
Indentation

### LINKED ISSUE #636: Create nicer SQL
The SQL queries generated by `dbplyr` are not very readable. Of course one can use one of the many available online SQL formatters but it would be still be very nice if the query itself would be more readable.

For example turn this

```SQL
SELECT `x.x`, `x.y`, `y`
FROM (SELECT `x.x`, `x.y`, `y`, AVG(`x.x`) OVER (PARTITION BY `y`) AS `q01`
FROM (SELECT `LHS`.`x` AS `x.x`, `RHS`.`x` AS `x.y`, `y`
FROM `dbplyr_001` AS `LHS`
LEFT JOIN `dbplyr_002` AS `RHS`
ON (LHS.x < RHS.x)
))
WHERE (`q01` > 1.0)
```

into something like this

```SQL
SELECT `x.x`,
       `x.y`,
       `y`
FROM
  (SELECT `x.x`,
          `x.y`,
          `y`,
          AVG(`x.x`) OVER (PARTITION BY `y`) AS `q01`
   FROM
     (SELECT `LHS`.`x` AS `x.x`,
             `RHS`.`x` AS `x.y`,
             `y`
      FROM `dbplyr_001` AS `LHS`
      LEFT JOIN `dbplyr_002` AS `RHS` ON (LHS.x < RHS.x)))
WHERE (`q01` > 1.0)
```

I looked around a little for existing R packages providing an SQL formatter but there seems to hardly any. I found [sqlparseR](https://cran.r-project.org/web/packages/sqlparseR/index.html) which has quite a lot of extra Python dependencies.

`dbplyr` could directly generate pretty SQL code but this would probably require quite a bit of work and maintenance. So, it might be easier to create a wrapper around an existing SQL formatter.

mgirlich:
Fixes #636.

I am not sure what should be controllable via an option and what should be the new default. I think the following options make sense

* **base indent:** the base indent used; defaults to two spaces "&nbsp;&nbsp;" but this could be any number of spaces or tabs
* **indent_strategy:** I added four strategies
  * `"legacy"`: produces the same output as `dbplyr` does currently
  * `"minimal"`: indent subqueries
  * `"indent"`: each field on its own line. Line break after keyword and fields are one level more intended than the keyword
```sql
SELECT
  `mpg`,
  `cyl`,
  `disp`
```
  * `"align"`: each field on its own line. No line break and fields are all aligned
```sql
SELECT `mpg`,
       `cyl`,
       `disp`
```
* **align on AS:**

maybe
* **"AND" position:**
  * `"line_end"`
  * `"line_start"`: it is a bit easier to remove (or comment out) the last condition
```sql
WHERE
  `mpg` > 1 AND
  `cyl` = 6

-- vs

WHERE
  `mpg` > 1
  AND `cyl` = 6
```
* **"comma" position:** same like for "AND"
* **"keyword alignment:**
  * `"left"`
  * `"right"`
```sql
SELECT *
FROM df

-- vs

SELECT *
  FROM df
```
* **truncate strings**: truncate long strings?
* **truncate lists/vectors**

## Select query
``` r
tbl_lazy(mtcars) %>% 
  mutate(mpg = mpg + 1,
         mpg = mpg - 1) %>% 
  filter(
    mpg > 1,
    cyl == 6
  ) %>% 
  group_by(mpg, cyl) %>% 
  summarise(mpg = mean(mpg, na.rm = TRUE))
```

before
```sql
SELECT `mpg`, `cyl`, AVG(`mpg`) AS `mpg`
FROM (SELECT `mpg` - 1.0 AS `mpg`, `cyl`, `disp`, `hp`, `drat`, `wt`, `qsec`, `vs`, `am`, `gear`, `carb`
FROM (SELECT `mpg` + 1.0 AS `mpg`, `cyl`, `disp`, `hp`, `drat`, `wt`, `qsec`, `vs`, `am`, `gear`, `carb`
FROM `df`) `q01`) `q02`
WHERE ((`mpg` > 1.0) AND (`cyl` = 6.0))
GROUP BY `mpg`, `cyl`
```

after
```sql
SELECT
  `mpg`,
  `cyl`,
  AVG(`mpg`) AS `mpg`
FROM (
  SELECT
    `mpg` - 1.0 AS `mpg`,
    `cyl`,
    `disp`,
    `hp`,
    `drat`,
    `wt`,
    `qsec`,
    `vs`,
    `am`,
    `gear`,
    `carb`
  FROM (
    SELECT
      `mpg` + 1.0 AS `mpg`,
      `cyl`,
      `disp`,
      `hp`,
      `drat`,
      `wt`,
      `qsec`,
      `vs`,
      `am`,
      `gear`,
      `carb`
    FROM `df`
  ) `q01`
) `q02`
WHERE
  (`mpg` > 1.0) AND
  (`cyl` = 6.0)
GROUP BY
  `mpg`,
  `cyl`
```

## Union Query
```r
union(
  union(
    tbl_lazy(mtcars) %>% 
      select(mpg, cyl),
    tbl_lazy(mtcars) %>% 
      select(mpg, cyl)
  ),
  tbl_lazy(mtcars) %>% 
    filter(mpg > 1) %>% 
    transmute(mpg = mpg + 1, cyl)
)
```

before

```sql
((SELECT `mpg`, `cyl`
FROM `df`)
UNION
(SELECT `mpg`, `cyl`
FROM `df`))
UNION
(SELECT `mpg` + 1.0 AS `mpg`, `cyl`
FROM `df`
WHERE (`mpg` > 1.0))
```

after

```sql
(
  (
    SELECT
      `mpg`,
      `cyl`
    FROM `df`
  )
  UNION
  (
    SELECT
      `mpg`,
      `cyl`
    FROM `df`
  )
)
UNION
(
  SELECT
    `mpg` + 1.0 AS `mpg`,
    `cyl`
  FROM `df`
  WHERE (`mpg` > 1.0)
)
```

### DIFF
--- a/tests/testthat/_snaps/backend-mssql.md
+++ b/tests/testthat/_snaps/backend-mssql.md
@@ -102,22 +102,24 @@
 # handles ORDER BY in subqueries
 
     Code
-      sql_query_select(simulate_mssql(), "x", "y", order_by = "z", subquery = TRUE)
+      sql_query_select(simulate_mssql(), ident("x"), ident("y"), order_by = "z",
+      subquery = TRUE)
     Warning <warning>
       ORDER BY is ignored in subqueries without LIMIT
       i Do you need to move arrange() later in the pipeline or use window_order() instead?
     Output
-      <SQL> SELECT 'x'
-      FROM 'y'
+      <SQL> SELECT `x`
+      FROM `y`
 
 # custom limit translation
 
     Code
-      sql_query_select(simulate_mssql(), "x", "y", order_by = "z", limit = 10)
+      sql_query_select(simulate_mssql(), ident("x"), ident("y"), order_by = ident("z"),
+      limit = 10)
     Output
-      <SQL> SELECT TOP 10 'x'
-      FROM 'y'
-      ORDER BY 'z'
+      <SQL> SELECT TOP 10 `x`
+      FROM `y`
+      ORDER BY `z`
 
 # custom escapes translated correctly
 
--- a/tests/testthat/_snaps/backend-mssql.md
+++ b/tests/testthat/_snaps/backend-mssql.md
@@ -34,7 +34,9 @@
       mf %>% mutate(z = ifelse(x == 1L, 1L, 2L))
     Output
       <SQL>
-      SELECT `x`, IIF(`x` = 1, 1, 2) AS `z`
+      SELECT
+        `x`,
+        IIF(`x` = 1, 1, 2) AS `z`
       FROM `df`
 
 ---
@@ -43,7 +45,9 @@
       mf %>% mutate(z = case_when(x == 1L ~ 1L))
     Output
       <SQL>
-      SELECT `x`, CASE
+      SELECT
+        `x`,
+        CASE
       WHEN (`x` = 1) THEN (1)
       END AS `z`
       FROM `df`
@@ -54,7 +58,9 @@
       mf %>% mutate(z = !is.na(x))
     Output
       <SQL>
-      SELECT `x`, CAST(IIF(~((`x`) IS NULL), 1, 0) AS BIT) AS `z`
+      SELECT
+        `x`,
+        CAST(IIF(~((`x`) IS NULL), 1, 0) AS BIT) AS `z`
       FROM `df`
 
 ---
--- a/R/backend-access.R
+++ b/R/backend-access.R
@@ -41,15 +41,16 @@ sql_query_select.ACCESS <- function(con, select, from,
                               where = NULL,  group_by = NULL,
                               having = NULL, order_by = NULL,
                               limit = NULL,  distinct = FALSE, ...,
-                              subquery = FALSE) {
+                              subquery = FALSE,
+                              level = 0) {
 
   sql_select_clauses(con,
-    select    = sql_clause_select(con, select, distinct, top = limit),
-    from      = sql_clause_from(con, from),
-    where     = sql_clause_where(con, where),
-    group_by  = sql_clause_group_by(con, group_by),
-    having    = sql_clause_having(con, having),
-    order_by  = sql_clause_order_by(con, order_by, subquery, limit)
+    select    = sql_clause_select(con, select, distinct, top = limit, level = level),
+    from      = sql_clause_from(con, from, level = level),
+    where     = sql_clause_where(con, where, level = level),
+    group_by  = sql_clause_group_by(con, group_by, level = level),
+    having    = sql_clause_having(con, having, level = level),
+    order_by  = sql_clause_order_by(con, order_by, subquery, limit, level = level)
   )
 }
 
--- a/R/backend-access.R
+++ b/R/backend-access.R
@@ -42,15 +42,15 @@ sql_query_select.ACCESS <- function(con, select, from,
                               having = NULL, order_by = NULL,
                               limit = NULL,  distinct = FALSE, ...,
                               subquery = FALSE,
-                              level = 0) {
+                              lvl = 0) {
 
   sql_select_clauses(con,
-    select    = sql_clause_select(con, select, distinct, top = limit, level = level),
-    from      = sql_clause_from(con, from, level = level),
-    where     = sql_clause_where(con, where, level = level),
-    group_by  = sql_clause_group_by(con, group_by, level = level),
-    having    = sql_clause_having(con, having, level = level),
-    order_by  = sql_clause_order_by(con, order_by, subquery, limit, level = level)
+    select    = sql_clause_select(con, select, distinct, top = limit, lvl = lvl),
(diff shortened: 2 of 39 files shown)


tests/testthat/_snaps/backend-mssql.md
```diff
@@ -96,22 +102,24 @@
...
-      sql_query_select(simulate_mssql(), "x", "y", order_by = "z", subquery = TRUE)
+      sql_query_select(simulate_mssql(), ident("x"), ident("y"), order_by = "z",
+      subquery = TRUE)
     Warning <warning>
       ORDER BY is ignored in subqueries without LIMIT
       i Do you need to move arrange() later in the pipeline or use window_order() instead?
     Output
-      <SQL> SELECT 'x'
-      FROM 'y'
+      <SQL> SELECT `x`
```
hadley, line 111: Should the quoting have changed? Did you fix a bug?

hadley (commented):
This feels like a big improvement to readability, and I don't think we need to provide any options to control the display (if you want that, you can use an external SQL formatter).

To keep things more compact, how hard would it be to only put each SELECTed variable on its own line if the total width is greater than (say) 80 characters? That way:

```sql
SELECT
  `x`,
  `y`
FROM `df`
GROUP BY
  `x`,
  `y`
```

could become

```sql
SELECT `x`, `y`
FROM `df`
GROUP BY `x`, `y`
```

That's much more compact and IMO readability doesn't suffer.

tests/testthat/_snaps/backend-mssql.md
mgirlich: Ah, yes, there was a bug in the tests, passing bare characters instead of identifiers/sql, see

https://github.com/tidyverse/dbplyr/pull/642/files#diff-0a78a645632601539fd36938d193efcde6f89ef87f569165ee090508e4fe0474

mgirlich:
Basically, that should be relatively easy. But this raises some questions regarding mixing styles:

* Check the length of each clause separately? That is put each SELECTed variable on its own line but all ORDER and GROUP BY on one line?

Compact
```sql
SELECT
  `mpg` - 1.0 AS `mpg`,
  `cyl`,
  `disp`,
  `hp`,
  `drat`,
  `wt`,
  `qsec`,
  `vs`,
  `am`,
  `gear`,
  `carb`
FROM `df`
WHERE (`mpg` > 1.0) AND (`cyl` = 6.0)
ORDER BY `mpg`, `cyl`
```

Long
```sql
SELECT
  `mpg` - 1.0 AS `mpg`,
  `cyl`,
  `disp`,
  `hp`,
  `drat`,
  `wt`,
  `qsec`,
  `vs`,
  `am`,
  `gear`,
  `carb`
FROM `df`
WHERE
  (`mpg` > 1.0)
  AND (`cyl` = 6.0)
ORDER BY
  `mpg`,
  `cyl`
```

* Similarly for queries. If the `SELECT` clause in the inner query has each variable on its own line what should the outer query do?

Compact
```sql
SELECT `mpg`, `cyl`, AVG(`mpg`) AS `mpg`
FROM (
  SELECT
    `mpg` - 1.0 AS `mpg`,
    `cyl`,
    `disp`,
    `hp`,
    `drat`,
    `wt`,
    `qsec`,
    `vs`,
    `am`,
    `gear`,
    `carb`
  FROM `df`
) `q01`
WHERE (`mpg` > 1.0) AND (`cyl` = 6.0)
GROUP BY `mpg`, `cyl`
```

Long
```sql
SELECT
  `mpg`,
  `cyl`,
  AVG(`mpg`) AS `mpg`
FROM (
  SELECT
    `mpg` - 1.0 AS `mpg`,
    `cyl`,
    `disp`,
    `hp`,
    `drat`,
    `wt`,
    `qsec`,
    `vs`,
    `am`,
    `gear`,
    `carb`
  FROM `df`
) `q01`
WHERE
  (`mpg` > 1.0)
  AND (`cyl` = 6.0)
GROUP BY
  `mpg`,
  `cyl`
```

The compact version is probably a bit nicer but mixing styles can also be a bit confusing.

mgirlich:
And which options in the starting post do you prefer regarding readability?

hadley:
I think you'd just apply the formatting locally to each SELECT/GROUP BY/ORDER clause; it's fine for there to be some inconsistency as using global reasoning is out of scope IMO (since it's so much more work).

I don't have any strong feelings for the other options except that I think all the clauses for a single query should be indented by the same amount, i.e.:

```sql
SELECT *
FROM df
```

mgirlich:
If the line length is at most 80 characters then everything is printed on the same line. Should we make an option out of this?

hadley:
I think it's fine just to set a fixed width for now; I'm worried that adding any options leads to a slippery slope of user requests to make this is fully fledged formatting solution.

mgirlich:
Replaced by #732.

PR #645, opened 2021-04-15
support `missing` argument in `if_else()`

### LINKED ISSUE #641: if_else missing keyword errors
The following code errors in dbplyr but shouldn't:

``` r
library(tidyverse)
library(dbplyr)

lf <- lazy_frame(a = c(1,2,3,NA))

lf %>% mutate(if_else(condition = a > 1, true = a, false = 1, missing = 1))
#> Error in if_else(condition = a > 1, true = a, false = 1, missing = 1): unused argument (missing = 1)
```

A simple `coalesce()` does the job for now, but putting this on the radar! 

<sup>Created on 2021-04-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

<sup>Created on 2021-04-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

mgirlich:
Fixes #641

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dbplyr (development version)
 
+* `if_else()` now supports the `missing` argument (@mgirlich, #641).
+
 # dbplyr 2.1.1
 
 * New support for Snowflake (@edgararuiz)
--- a/R/backend-.R
+++ b/R/backend-.R
@@ -135,7 +135,7 @@ base_scalar <- sql_translator(
   },
 
   `if` = sql_if,
-  if_else = function(condition, true, false) sql_if(condition, true, false),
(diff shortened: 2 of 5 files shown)


krlmlr (commented):
Thanks. What's the impact on backends that might have implemented a custom translations for `if_else()`? (I don't know if this is a thing already.)

mgirlich:
If a backend already has a custom translation for `if_else()` then the translation from the default backend should simply be overwritten. This is also what happens in this PR for MS SQL.
So, I don't think this has any consequences for other backends.

krlmlr:
So, as long as nobody calls `if_else()` with four arguments, everything remains the same in this case?

mgirlich:
Yes, nothing changes if you use less than four arguments resp. pass nothing to `missing`.

PR #646, opened 2021-04-16
Minor vignette enhancements

mgirlich:
The generated SQL is now minimal, so the note doesn't really fit there anymore. I am not sure whether you want to put the note somewhere else.

### DIFF
--- a/vignettes/translation-verb.Rmd
+++ b/vignettes/translation-verb.Rmd
@@ -35,9 +35,6 @@ airports <- copy_to(con, nycflights13::airports)
       show_query()
     ```
     
-    (As you can see here, the generated SQL isn't always as minimal as you
-    might generate by hand.)
-    
 *  `filter()` generates a `WHERE` clause:
     
     ```{r}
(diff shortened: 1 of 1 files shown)


krlmlr (commented):
Should we come up with an example that generates SQL that has room for improvement?

mgirlich:
I think if you have a simple example it makes sense.

mgirlich:
A simple example which could be simplified

``` r

flights <- copy_to(con, nycflights13::flights)

flights %>%
  mutate(speed = distance / (air_time / 60)) %>%
  filter(year > 2014) %>%  
  show_query()
#> <SQL>
#> SELECT *
#> FROM (SELECT `year`, `month`, `day`, `dep_time`, `sched_dep_time`, `dep_delay`, `arr_time`, `sched_arr_time`, `arr_delay`, `carrier`, `flight`, `tailnum`, `origin`, `dest`, `air_time`, `distance`, `hour`, `minute`, `time_hour`, `distance` / (`air_time` / 60.0) AS `speed`
#> FROM `nycflights13::flights`)
#> WHERE (`year` > 2014.0)
```

<sup>Created on 2021-11-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #647, opened 2021-04-16
Add `add_count()` method

### LINKED ISSUE #614: add_count returns grouped dataframe
According to the [dplyr docs](https://dplyr.tidyverse.org/reference/count.html), `add_count()` is supposed to:

> group transiently, so the output has the same groups as the input.

Although this is true for the dplyr version, I discovered that the dbplyr version does change the groups.

``` r
library(tidyverse)
library(dbplyr)

df <- tribble(
  ~name,    ~gender,   ~runs,
  "Max",    "male",       10,
  "Sandra", "female",      1,
  "Susan",  "female",      4
)

# ungrouped output
df %>% add_count(gender, wt = runs)
#> # A tibble: 3 x 4
#>   name   gender  runs     n
#>   <chr>  <chr>  <dbl> <dbl>
#> 1 Max    male      10    10
#> 2 Sandra female     1     5
#> 3 Susan  female     4     5

# grouped output
memdb_frame(df) %>% add_count(gender, wt = runs)
#> # Source:   lazy query [?? x 4]
#> # Database: sqlite 3.34.1 [:memory:]
#> # Groups:   gender
#>   name   gender  runs     n
#>   <chr>  <chr>  <dbl> <dbl>
#> 1 Sandra female     1     5
#> 2 Susan  female     4     5
#> 3 Max    male      10    10
```

<sup>Created on 2021-03-11 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

mgirlich:
Fixes #614


davidchall:
@mgirlich - I think you need to change the DESCRIPTION so that dbplyr temporarily depends on the dev version of dplyr. Currently, it looks like you export a function instead of an S3 method, because roxygen doesn't find the generic.

davidchall:
... and now it's on CRAN 😂

R/verb-count.R
```diff
@@ -26,6 +27,24 @@ count.tbl_lazy <- function(x, ..., wt = NULL, sort = FALSE, name = NULL) {
   out
 }
 
+#' @rdname count.tbl_lazy
+#' @importFrom dplyr add_count
+#' @importFrom lifecycle deprecated
+#' @export
+add_count.tbl_lazy <- function (x, ..., wt = NULL, sort = FALSE, name = NULL, .drop = deprecated()) {
```
krlmlr, line 34: I've seen `lifecycle::deprecated()` in other places, not sure what the common practice is. Also, if it's unsupported, should this be simply `NULL` instead?

R/verb-count.R
```diff
@@ -26,6 +27,24 @@ count.tbl_lazy <- function(x, ..., wt = NULL, sort = FALSE, name = NULL) {
...
+    abort("`.drop` argument not supported for lazy tables.")
+  }
+  if (!missing(...)) {
+    out <- group_by(x, ..., .add = TRUE)
+  }
+  else {
+    out <- x
+  }
+  out <- dplyr::add_tally(out, wt = !!enquo(wt), sort = sort, name = name)
+  group_by(out, !!!syms(group_vars(x)))
```
krlmlr, line 45: Should we regroup only if we grouped before?

tests/testthat/test-verb-count.R
```diff
@@ -4,13 +4,17 @@ test_that("generates expected SQL for common situations", {
   expect_snapshot(db %>% count(g))
   expect_snapshot(db %>% count(g, wt = x))
   expect_snapshot(db %>% count(g, sort = TRUE))
+
+  expect_snapshot(db %>% add_count(g, sort = TRUE))
```
krlmlr, line 8: Out of scope here, just wondering:

```suggestion
  expect_snapshot({
    db %>% count(g)
    db %>% count(g, wt = x)
    db %>% count(g, sort = TRUE)
    db %>% add_count(g, sort = TRUE)
  })
```

?

krlmlr (commented):
Looks good, thanks!

R/verb-count.R
mgirlich: * grouped before -> grouped again
* not grouped before -> this boils down to `group_by(out)` which adds an unnecessary ungroup operation but still produces the correct result.

One could add an extra check before it the tbl was grouped but I think that `group_by()` resp. `ungroup()` should avoid adding the unnecessary ungroup operation.

tests/testthat/test-verb-count.R
mgirlich: Neat, I didn't know this works resp. just didn't think about this pattern!

krlmlr: It's not necessarily better: the existing code allows adding new snippets without having to accept the snapshot manually.

R/verb-count.R
mgirlich: I'm afraid I don't really know either.

eutwt: From `?deprecated`: "We recommend importing lifecycle::deprecated() in your namespace and use it without the namespace qualifier.

In general, we advise against such magical defaults, i.e. defaults that cannot be evaluated by the user. In the case of deprecated(), the trade-off is worth it because the meaning of this default is obvious and there is no reason for the user to call deprecated() themselves."

mgirlich: I think this should simply be `NULL` as it is unsupported anyway.

PR #616, opened 2021-03-12
Group by ungroups for empty dots

mgirlich:
Fixes #615

PR #649, opened 2021-04-20
[bug-fix] remote_name() should ignore group_by() and ungroup() ops correctly

yitao-li:
because group_by and ungroup have no effect on the remote name (or lack thereof)


krlmlr:
Do we also need to implement `remote_name.op_arrange()`, because `compute()` unconditionally includes an `arrange()` call?

hadley:
What problem does this solve?

yitao-li:
@hadley Before this change, if something, say `x`, has a remote name to begin with, an ungroup op `x %>% ungroup()` would cause the remote name to become `NULL`, even though `ungroup()` should not have that effect. Same for `group_by()`.

krlmlr:
This is especially relevant after `compute()`, which does unconditional `group_by()` and `arrange()` . I propose to implement a method for the "arrange" operation too.

yitao-li:
@krlmlr I'm not entirely sure about `op_arrange()`: I thought it would generate an 'order by' clause? (whereas group_by() and ungroup() won't generate any clause by themselves).

krlmlr:
I think `x %>% arrange(...) %>% remote_name()` should be the same as `x %>% remote_name()`, because the sort order is irrelevant for most databases. This ensures that `x %>% arrange(...) %>% compute() %>% remote_name()` is the same as `x %>% compute() %>% remote_name()` .

yitao-li:
@krlmlr I'm still not completely convinced to be honest:

consider
`x %>% arrange(...) %>% remote_query()` vs `x %>% remote_query()` -- both queries will be (textually) different and whether both would end up with identical materialized results after `compute()` is implementation-dependent.

whereas for `x %>% ungroup() %>% remote_query()` vs `x %>% remote_query()` -- both queries are guaranteed to be word-by-word identical so if you replace `remote_query()` with `compute()` the materialized results would have been identical.

krlmlr:
I forgot a reprex.

``` r
library(tidyverse)
library(dbplyr)


memdb_frame(x = 1) %>%
  compute() %>%
  remote_name()
#> NULL

memdb_frame(x = 1) %>%
  group_by(x) %>%
  compute() %>%
  remote_name()
#> NULL

memdb_frame(x = 1) %>%
  arrange(x) %>%
  compute() %>%
  remote_name()
#> NULL

memdb_frame(x = 1) %>%
  group_by(x) %>%
  arrange(x) %>%
  compute() %>%
  remote_name()
#> NULL

memdb_frame(x = 1) %>%
  arrange(x) %>%
  arrange() %>%
  compute() %>%
  remote_name()
#> NULL
```

<sup>Created on 2021-04-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

The first two cases should really return a `remote_name()` . I'd argue strongly that the fifth also should return a `remote_name()` -- `arrange()` should effectively pop the last sort operation from the stack. As for the third and fourth cases, I'd leave this to your consideration.

hadley:
Note the documentation for `remote_name()`: `remote_name()` gives the name remote table, or NULL if it's a query. So you're proposing a considerable change here.

yitao-li:
@hadley The main goal I have is to restore the behavior of dbplyr 2.0. where if `x` has a remote name (i.e., not a query to begin with), then the result of `x %>% compute()` will also have a remote name.

This was the case up to dbplyr 2.0 but stopped being the case in dbplyr 2.1.x. So, if there is some reasonable way to restore everything to the dbplyr 2.0.0 behavior that would be great.

krlmlr:
It looks like this was introduced in 43559165cb051b3a2f1a9a1f4abfc05c32f8d020 (#616).

hadley:
I no longer have a strong sense of what's correct here. @mgirlich can you make the call?

mgirlich:
Replaced by #850.
I think it makes sense that after grouping and `arrange(x) %>% arrange()` a name is returned. The (extended) example from above now looks as following:

``` r
mf <- memdb_frame(x = 1)

# name after group_by()
mf %>% 
  group_by(x) %>% 
  remote_name()
#> <IDENT> dbplyr_001

# name after unarrange
mf %>% 
  arrange(x) %>% 
  arrange() %>% 
  remote_name()
#> <IDENT> dbplyr_001

# always a name after `compute()`
mf %>% 
  compute() %>%
  remote_name()
#> <IDENT> dbplyr_002

mf %>% 
  group_by(x) %>%
  compute() %>%
  remote_name()
#> <IDENT> dbplyr_003

mf %>% 
  arrange(x) %>%
  compute() %>%
  remote_name()
#> <IDENT> dbplyr_004

mf %>% 
  group_by(x) %>%
  arrange(x) %>%
  compute() %>%
  remote_name()
#> <IDENT> dbplyr_005

mf %>% 
  arrange(x) %>%
  arrange() %>%
  compute() %>%
  remote_name()
#> <IDENT> dbplyr_006
```

<sup>Created on 2022-04-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #655, opened 2021-05-17
support `.name_repair` in `nesting()`

### LINKED ISSUE #654: `nesting()` doesn't respect `.name_repair`
Instead of repairing names it adds a column `.name_repair`

``` r
library(tidyr)
library(dbplyr)

memdb_frame(x = 1) %>% expand(nesting(x, .name_repair = "unique"))
#> # Source:   lazy query [?? x 2]
#>       x .name_repair
#>   <dbl> <chr>       
#> 1     1 unique
```

<sup>Created on 2021-05-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

mgirlich:
Fixes #654


R/verb-expand.R
```diff
@@ -64,6 +54,23 @@ expand.tbl_lazy <- function(data, ..., .name_repair = "check_unique") {
   purrr::reduce(distinct_tables, left_join, by = group_vars(data))
 }
 
+extract_dot_vars <- function(dot) {
```
krlmlr, line 57: Is the name still applicable, now that this is a top-level function?

R/verb-expand.R
```diff
@@ -64,6 +54,23 @@ expand.tbl_lazy <- function(data, ..., .name_repair = "check_unique") {
...
+  # ugly hack to deal with `nesting()`
+  if (quo_is_call(dot, name = "nesting")) {
```
krlmlr, line 59: This hasn't worked before, but as we're touching this code:

Would that work if the user calls `tidyr::nesting()` in `expand()` ?

Also, do we want to implement an early return instead?

```suggestion
  if (!quo_is_call(dot, name = "nesting")) {
    return(...)
```

krlmlr (commented):
Nice hack!

R/verb-expand.R
mgirlich: 👍  Renamed to `extract_expand_dot_vars()`.

R/verb-expand.R
mgirlich: Yes, it does work with `tidyr::nesting()`. By default `quo_is_call()` ignores the namespace. But it can be made more rigorous as follows

``` r
library(rlang)

ns <- c("tidyr", "")
quo_is_call(quo(nesting()), name = "nesting", ns = ns)
#> [1] TRUE
quo_is_call(quo(tidyr::nesting()), name = "nesting", ns = ns)
#> [1] TRUE
quo_is_call(quo(base::nesting()), name = "nesting", ns = ns)
#> [1] FALSE
```

<sup>Created on 2021-08-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #668, opened 2021-06-29
Use known column names in copy_to()

krlmlr:
This avoids an extra roundtrip to the database.


hadley:
Add a test since it'd be easy to forget to do this in the future?

hadley:
@krlmlr do you want to finish this off? We're aiming for a release shortly.

krlmlr:
On it.

krlmlr:
I've noted that `tbl()` is called internally quite a few more times, and I suspect that the column names are always known. Happy to fix it as part of this PR.

For testing, I propose to implement an internal helper that requires `vars` to be non-NULL and forwards to `tbl_sql()`, and a test that checks all namespace functions for the presence of calls to `tbl()`.

krlmlr:
Quick draft for an expectation that asserts that a symbol isn't used inside the package. It's not pretty, in particular the expectation message isn't very self-evident yet. Happy to take feedback: perhaps a similar thing exists elsewhere?

```r
expect_function_unused <- function(name) {
  start <- as.list(asNamespace("dbplyr"), all.names = TRUE)
  is_function <- purrr::map_lgl(start, is.function)
  purrr::walk(start[is_function], expect_function_unused_call, name)
}

expect_function_unused_call <- function(x, name) {
  if (is.function(x)) {
    expect_function_unused_call(body(x), name)
  } else if (is.call(x)) {
    if (identical(as.character(x[[1]]), name)) {
      expect_true(FALSE, label = as_label(x))
    }

    purrr::walk(as.list(x)[-1], expect_function_unused_call, name)
  } else if (is.symbol(x)) {
    if (identical(as.character(x), name)) {
      expect_true(FALSE, label = as_label(x))
    }
  }
}
```

tests/testthat/_snaps/verb-uncount.md
```diff
@@ -8,7 +8,15 @@
...
+        INNER JOIN (
+      SELECT CAST(`..dbplyr_row_id` AS INTEGER) AS `..dbplyr_row_id`
+      FROM (
+        SELECT NULL AS `..dbplyr_row_id`
+        WHERE (0 = 1)
+        UNION ALL
+        VALUES (1)
+      ) AS `values_table`
+        ) AS `RHS`
           ON (`RHS`.`..dbplyr_row_id` <= `LHS`.`w`)
```
krlmlr, line 20: @mgirlich: The indentation is off here. I wonder if we want an `op_df()` that we can insert into the SQL expression tree that would, when rendered, be turned into a CTE?

krlmlr:
Still not sure how to test it. Should I add a test that verifies that `tbl()` and `tbl_sql()` are not used internally?

tests/testthat/_snaps/verb-uncount.md
mgirlich: When I implemented `copy_inline()` I knew that in these situations the indentation is off but I thought it would take a least a little bit longer until someone complains 😄
The issue is that the SQL for `copy_inline()` is created directly, so it cannot know about the level where it needs to be rendered.

Even with CTE the indentation is off

```r
dbplyr_uncount(df, w) %>% show_query(cte = TRUE)
```

```sql
WITH `q01` AS (
  SELECT `x`, `w`, `..dbplyr_row_id`
  FROM `test` AS `LHS`
  INNER JOIN (
SELECT CAST(`..dbplyr_row_id` AS INTEGER) AS `..dbplyr_row_id`
FROM (
  SELECT NULL AS `..dbplyr_row_id`
  WHERE (0 = 1)
  UNION ALL
  VALUES (1)
) AS `values_table`
  ) AS `RHS`
    ON (`RHS`.`..dbplyr_row_id` <= `LHS`.`w`)
)
SELECT `x`
FROM `q01`
```

I'll take a look how much effort it is to change this.

I guess in this case we want, if CTEs are activated, the values clause to be rendered in a CTE. I'm not sure whether there are other cases like this in the dbplyr code. But I already thought it might be interesting to have a user facing function for this.

mgirlich: With explicitly turning on CTEs the query looks much better after https://github.com/tidyverse/dbplyr/pull/810.
Do you still see the need for an `op_vars()` to turn parts into a CTE?

hadley:
I'd suggest just testing that the names are present where `tbl()` is used. It's less important to worry about future usage.

tests/testthat/_snaps/verb-uncount.md
krlmlr: With `op_vars` we would get `WITH q01 AS (SELECT ...) WITH q02 AS SELECT ... INNER JOIN q01 ...` ? That would be a tad more readable.

Does this issue arise whenever a `copy_inline()` is used as a joined or unioned table?

krlmlr:
`tbl()` is a generic in dplyr. This PR changes all calls to `tbl()` into calls to a new private function that requires that the `vars` argument is provided, so that `tbl()` is no longer used internally. What else do we need to change here?

hadley:
Just test the places it's used.

tests/testthat/_snaps/verb-uncount.md
mgirlich: Isn't the first example in #810 exactly what you want?

```sql
WITH `q01` AS (
  SELECT CAST(`..dbplyr_row_id` AS INTEGER) AS `..dbplyr_row_id`
  FROM (
    SELECT NULL AS `..dbplyr_row_id`
    WHERE (0 = 1)
    UNION ALL
    VALUES (1)
  ) AS `values_table`
),
`q02` AS (
  SELECT `x`, `w`, `..dbplyr_row_id`
  FROM `test` AS `LHS`
  INNER JOIN `q01` AS `RHS`
    ON (`RHS`.`..dbplyr_row_id` <= `LHS`.`w`)
)
SELECT `x`
FROM `q02`
```

krlmlr:
I changed the codecov settings, now codecov reports that 100% of the patch are tested. If we still need new tests, can you please specify what they should look like? The behavior or the internal data structures don't change with this PR.

R/src_dbi.R
```diff
@@ -85,6 +85,16 @@ tbl.src_dbi <- function(src, from, ...) {
   tbl_sql(c(subclass, "dbi"), src = src, from = from, ...)
 }
 
+# Internal calls to `tbl()` should be avoided in favor of tbl_src_dbi().
+# The former may query the database for column names if `vars` is omitted,
+# the latter always requires `vars`.
+tbl_src_dbi <- function(src, from, vars) {
```
hadley, line 91: How about just adding something like:

```R
tbl <- function(...) {
  abort("Use `tbl_src_dbi()` instead")
}
```

I forgot that we were already automatically pulling in the variable names if they weren't supplied.

krlmlr: Unfortunately this doesn't work because we are importing `tbl()` from dplyr and also declaring methods for it. As soon as we add the override, the dbplyr vignette fails rendering because the `tbl.sqlite()` method can't be found. I tried various hacks, including very weird ones.

We still could inspect our source code in a test.

hadley: Ok, let's just leave it as is then.

hadley (approved):

PR #671, opened 2021-07-06
Keep grouping variables in `transmute()`

### LINKED ISSUE #670: transmute() discards groups
but shouldn't? The resulting lazy frame still refers to the now discarded columns.

``` r
library(tidyverse)

dbplyr::memdb_frame(a = 1, b = 2) %>%
  group_by(a) %>%
  transmute(b)
#> # Groups:   a
#>       b
#>   <dbl>
#> 1     2

tibble(a = 1, b = 2) %>%
  group_by(a) %>%
  transmute(b)
#> # A tibble: 1 x 2
#> # Groups:   a [1]
#>       a     b
#>   <dbl> <dbl>
#> 1     1     2
```

<sup>Created on 2021-07-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

mgirlich:
Fixes #670.

`mutate.tbl_lazy()` now uses the same logic as for data frames. This made it trivial to add support for the `.keep` argument. See https://github.com/tidyverse/dplyr/pull/6035 for an explanation of the new logic.

Note that `keep = "used"` and `keep = "unused"` do not work when sql is used. Maybe this should be detected and throw an error.

# Outdated
~Note that in `dplyr` the order of grouping variables can change. For simplicity and consistency with `dplyr` the same thing happens with this PR~

``` r

# column `b` gets pushed to the front
data.frame(a = 1, b = 2) %>%
  group_by(a, b) %>%
  transmute(a = a + 1)
#> # A tibble: 1 x 2
#> # Groups:   a, b [1]
#>       b     a
#>   <dbl> <dbl>
#> 1     2     2
```

<sup>Created on 2021-07-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>


krlmlr (commented):
I'd need to take a closer look here.

- The missing group variables seem to be appended before everything else, do we want this? Can we make them stay put as close to the original location as possible?
- What happens (here and in dplyr) if the `transmute()` changes grouping variables?

mgirlich:
> * The missing group variables seem to be appended before everything else, do we want this? Can we make them stay put as close to the original location as possible?

`dplyr` does that and and I rather wanted to be consistent. Personally, I'm not sure it's the best behaviour.

> * What happens (here and in dplyr) if the `transmute()` changes grouping variables?

The grouping variable is mutated and the grouping variables are kept the same (and in `dplyr` the groups are also correctly recalculated

``` r

data.frame(a = c(0, 10, 1, 11), b = c(1, 1, 2, 2)) %>%
  group_by(b) %>%
  transmute(b = b - min(a))
#> # A tibble: 4 × 1
#> # Groups:   b [1]
#>       b
#>   <dbl>
#> 1     1
#> 2     1
#> 3     1
#> 4     1

dbplyr::memdb_frame(a = c(0, 10, 1, 11), b = c(1, 1, 2, 2)) %>%
  group_by(b) %>%
  transmute(b = b - min(a))
#> Warning: Missing values are always removed in SQL.
#> Use `MIN(x, na.rm = TRUE)` to silence this warning
#> This warning is displayed only once per session.
#> # Groups:   b
#>       b
#>   <dbl>
#> 1     1
#> 2     1
#> 3     1
#> 4     1
```

<sup>Created on 2021-08-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

mgirlich:
I changed to a different strategy to detect new layers. This has the advantage to produce shorter SQL in some cases for a single `mutate()` and can in a different PR also be used to produce shorter SQL for multiple `mutate()` in a row.

Example of shorter SQL

```r

lf <- lazy_frame(x = 1, y = 2)
lf %>% 
  mutate(
    x = x + 1,
    x = x + 1,
    y = y + 1,
    y = y + 1
  )
```

Before

```sql
SELECT `x`, `y` + 1.0 AS `y`
FROM (SELECT `x` + 1.0 AS `x`, `y` + 1.0 AS `y`
FROM (SELECT `x` + 1.0 AS `x`, `y`
FROM `df`) `q01`) `q02`
```

After

```sql
SELECT `x` + 1.0 AS `x`, `y` + 1.0 AS `y`
FROM (SELECT `x` + 1.0 AS `x`, `y` + 1.0 AS `y`
FROM `df`) `q01`
```

hadley:
I'm worried that changing the order the variables are evaluated might introduce subtle bugs. But I'm not 100% sure this is a real worry given that these expressions can't have side-effects.

mgirlich:
Closed in favour of #802.

PR #672, opened 2021-07-06
Fix `op_sort.op_arrange()` for multiple `arrange()`

### LINKED ISSUE #665: confusing output after multiple `arrange()` calls
For example,

```
library(dplyr)
library(dbplyr)

df <- memdb_frame(a = 1, b = 2)

df %>% arrange(a) %>% arrange(b) %>% print()
```

will output

```
# Source:     table<dbplyr_001> [?? x 2]
# Database:   sqlite 3.35.5 [:memory:]
# Ordered by: a, b
      a     b
  <dbl> <dbl>
1     1     2
```

even though its generated query is

```
df %>% arrange(a) %>% arrange(b) %>% remote_query() %>% print()
```

```
<SQL> SELECT *
FROM `dbplyr_001`
ORDER BY `b`
```
(as expected)

So, the generated query is what I expected, but the "Ordered by: a, b" from the first print() output was confusing.

Similarly,

```
df %>% arrange(a, b) %>% print()
```

will also output "Ordered by: a, b", even though it will generate "ORDER BY `a`, `b`" instead of "ORDER BY `b`" as the SQL query.

Similar confusion happens for

```
df %>% arrange(a) %>% arrange(b) %>% arrange(a) %>% print()
```

as well (generated query will say "ORDER BY `b`" as expected but the print() output will say "Ordered by: a, b, a").

Anyhow, this seems to have caused some confusion and panicking for folks who probably mostly relied on REPL for debugging their SQL queries (e.g., see https://github.com/sparklyr/sparklyr/issues/1847) but hopefully it's something that's easy to fix.

mgirlich:
Fixes #665.

It seems this was intended in the past as there was even a test for this but as multiple arrange in a row now produce a warning and the previous ones are ignored probably this is not desired anymore. I'm not sure whether/how you want to test it as multiple `arrange()` give a warning anyways.

I'm not sure what a good news entry would look like in this case...

``` r

df <- memdb_frame(a = 1, b = 2)

df %>% 
  arrange(a) %>% 
  arrange(b)

# --- Before
#> # Source:     table<dbplyr_001> [?? x 2]
#> # Database:   sqlite 3.35.5 [:memory:]
#> # Ordered by: a, b
#>       a     b
#>   <dbl> <dbl>
#> 1     1     2

# --- After
#> # Source:     table<dbplyr_001> [?? x 2]
#> # Database:   sqlite 3.35.5 [:memory:]
#> # Ordered by: b
#>       a     b
#>   <dbl> <dbl>
#> 1     1     2
```

<sup>Created on 2021-07-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # dbplyr (development version)
 
+* `slice_sample()` now works for MySQL/MariaDB (@mgirlich, #617).
```
hadley, line 3: Not like this 😛  I think you can just skip a news entry in this case

hadley (approved):

PR #674, opened 2021-07-06
Translate evaluated expression

### LINKED ISSUE #634: R expression to SQL expression translation bug
the expression ``((1/(1 + abs(x))) - 1) * sign(x)`` does not seem to translate accurately into SQL under
certain conditions

in the below example, the expressions ``bug1`` and ``bug2`` are the same, the equality test returns TRUE.
expression ``bug1`` produces different results in ``R`` and ``SQL``
expression ``bug2`` produces the same results in ``R`` and ``SQL``

``` r
require(dplyr)
#> Loading required package: dplyr
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
require(dbplyr)
#> Loading required package: dbplyr
#> 
#> 
#>     ident, sql

bug_expr<-function(v)
{
  x <- enexpr(v)
  k  <- expr((1 / ( 1 + abs(!!x)))-1)
  expr(!!k*sign(!!x))
}

bug1<- bug_expr(x)
bug2<- expr(((1/(1 + abs(x))) - 1) * sign(x))

bug1==bug2
#> [1] TRUE

res <-
  memdb_frame(x=0:-3) %>%
  mutate(v_db=!!bug_expr(x)) %>%
  mutate(v1_db=!!bug1) %>%
  mutate(v2_db=!!bug2) %>%
  collect() %>% 
  mutate(v_r=!!bug_expr(x)) %>%
  mutate(v1_r=!!bug1) %>%
  mutate(v2_r=!!bug2)
  
all(res$v_db==res$v_r)
#> [1] FALSE
all(res$v1_db==res$v1_r)
#> [1] FALSE
all(res$v2_db==res$v2_r)
#> [1] TRUE
```

mgirlich:
Fixes #634.

I think the problem of partially evaluated expressions should only be an issue for infix expressions. I wasn't sure about adding operators like `|`, `&&`, or `==` to the list but it might make sense as well.


R/translate-sql-helpers.R
```diff
@@ -132,19 +132,62 @@ copy_env <- function(from, to = NULL, parent = parent.env(from)) {
 #' @param pad If `TRUE`, the default, pad the infix operator with spaces.
 #' @export
 sql_infix <- function(f, pad = TRUE) {
+  # (x1 <- expr((2 - 1) * x))
+  # #> (2 - 1) * x
+  #
+  # (x2 <- expr(!!expr(2 - 1) * x))
+  # #> (2 - 1) * x
+  #
+  # `x1` and `x2` look the same when printed but their AST (abstract syntax tree)
```
hadley, line 141: You could point to https://adv-r.hadley.nz/quasiquotation.html#non-standard-ast to keep this comment a bit shorter.

R/translate-sql-helpers.R
```diff
@@ -132,19 +132,62 @@ copy_env <- function(from, to = NULL, parent = parent.env(from)) {
...
+      y <- escape_infix_expr(enquo(y))
+
       build_sql(x, sql(f), y)
     }
   }
 }
 
+escape_infix_expr <- function(x) {
+  infix_calls <- c("+", "-", "*", "/", "%%", "^")
+  if (quo_is_call(x, n = 2) && call_name(x) %in% infix_calls) {
```
hadley, line 184: I think `quo_is_call(x, name, n = 2)` will work?

And I'm not sure `n = 2` is correct? Doesn't the same problem arise with unary `+` and `-`?

mgirlich: Hmm, when do we need parantheses for unary `+` and `-`? I came up with something like

```r
translate_sql(-!!expr(-x))
#> <SQL> --`x`
```

Do you see other cases?

Thanks to your question I just noticed that unary `+` isn't supported at all currently:

``` r

# works
translate_sql(-2)
#> <SQL> -2.0
# error
translate_sql(+2)
#> Error in eval_bare(x, .env): argument "y" is missing, with no default
```

<sup>Created on 2021-12-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mgirlich: Not sure how/whether we want to handle `translate_sql(-!!expr(-x))`. First the inner part, `-x`, translated to SQL ``-`x` ``. Then the outer minus is translated. Here one would have to analyse the SQL to decide whether parantheses are needed.
Probably it should be sufficient to check if the SQL starts with "-". What do you think?

hadley: I was thinking of this case:

``` r
dbplyr::translate_sql(-!!rlang::expr(1+2))
#> <SQL> -1.0 + 2.0
```

<sup>Created on 2021-12-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

OTOH is this very common? Should we just suggest that if you're going to interpolate in expressions involving infix operators, you should wrap the whole thing in parentheses to avoid operator precedence issues?

mgirlich:
I think this slightly changed approach should handle binary and unary minus.

tests/testthat/test-translate-sql-helpers.R
```diff
@@ -42,3 +42,16 @@ test_that("win_rank() is accepted by the sql_translator", {
     )
   )
 })
+
+test_that("can translate infix expression without parantheses", {
+  expect_equal(translate_sql(!!expr(2 - 1) * x), sql("(2.0 - 1.0) * `x`"))
+  expect_equal(translate_sql(!!expr(2 / 1) * x), sql("(2.0 / 1.0) * `x`"))
+
```
hadley, line 49:
```suggestion
```

tests/testthat/test-translate-sql-helpers.R
```diff
@@ -42,3 +42,16 @@ test_that("win_rank() is accepted by the sql_translator", {
...
+  expect_equal(translate_sql(!!expr(2 * 1) - x), sql("(2.0 * 1.0) - `x`"))
+})
+
+test_that("unary minus works with expressions", {
+  expect_equal(translate_sql(-!!rlang::expr(x+2)), sql("-(`x` + 2.0)"))
+
+  expect_equal(translate_sql(--x), sql("-(-`x`)"))
```
hadley, line 56:
```suggestion
  expect_equal(translate_sql(-!!expr(x+2)), sql("-(`x` + 2.0)"))
  expect_equal(translate_sql(--x), sql("-(-`x`)"))
```

hadley (approved):

hadley:
I forgot to say this earlier, but can you please do squash-merges? I think that gives a cleaner history, and in most cases we don't need to preserve the history within in a PR.

PR #677, opened 2021-07-07
Fix `fill()` for SQL Server

### LINKED ISSUE #651: fill does not work with SQL server
fill does not work with SQL server
```r
> tibble(x=c(1,2,3,NA,5,6,NA,NA),i=seq_along(x)) %>%
+ copy_to(name="test",dest=db_con,temporary=FALSE,overwrite=TRUE,append=FALSE) %>%
+ window_order(i) %>%
+ fill(x)  
Error: nanodbc/nanodbc.cpp:1655: 00000: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Incorrect syntax near 'IGNORE'.  [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Statement(s) could not be prepared. 
<SQL> 'SELECT TOP 11 "i", LAST_VALUE("x" IGNORE NULLS) OVER (ORDER BY "i") AS "x"
FROM "test"'
```

mgirlich:
Fixes #651


R/verb-fill.R
```diff
@@ -200,4 +203,9 @@ last_value_sql.Hive <- function(con, x) {
   translate_sql(last_value(!!x, TRUE), con = con)
 }
 
+#' @export
+`last_value_sql.Microsoft SQL Server` <- function(con, x) {
+  build_sql("LAST_VALUE(", ident(as.character(x)), ") IGNORE NULLS", con = con)
```
krlmlr, line 208: Does this need parentheses just to be sure?

```suggestion
  build_sql("(LAST_VALUE(", ident(as.character(x)), ") IGNORE NULLS)", con = con)
```

krlmlr (commented):
Do we need a test here? Do we have snapshot tests for MSSQL already?

R/verb-fill.R
mgirlich: I don't think so but we can also add some. `last_value_sql()` is only used in `fill()` and then the queries are simple select statements, e.g.

```SQL
SELECT `id`, `group`, LAST_VALUE(`n1` IGNORE NULLS) OVER (ORDER BY -`id`) AS `n1`
      FROM `df`
```

If you think it is still better to add parentheses please add them.

mgirlich:
> Do we need a test here? Do we have snapshot tests for MSSQL already?

👍  I just added a test.

PR #678, opened 2021-07-07
Add `copy_inline()`

### LINKED ISSUE #628: Create copy_to equivalent with read-only access
I use a database server where normal users do not have `TEMPORARY` privileges. This means that connections to the database are read-only and functions such as `compute()` or `dbWriteTable()` are not available. I suspect that this is not uncommon in the real world.

One downside of this is that if I have data locally that I want to join with data on the server, the only real option is to `collect` or otherwise download data from the server. But in a typical use case, the data I have locally are small, while the data on the server are large. So this is very inefficient.

The function below is a quick fix for this issue. It basically turns a data frame into SQL and creates a remote data frame using that SQL. Its use is illustrated [here](https://gist.github.com/iangow/aa7a631cae80b999313b1a8d777c0d3c). Performance is (unsurprisingly) worse than `dbWriteTable`, but it's not too bad.

The function below is obviously incomplete and PostgreSQL-focused (it works for my current use case), but would a function like this belong somewhere in `dbplyr`?

```r
library(DBI)

df_to_pg <- function(df, conn) {
  
  convert_vec <- function(vec) {
    if (class(vec)=="character") {
      convert_char(vec)
    } else if (class(vec)=="Date") {
      convert_date(vec)
    } else if (class(vec)=="integer") {
      convert_int(vec)
    } else if (class(vec)=="numeric") {
      convert_num(vec)
    }
  }
  
  make_string <- function(vec) {
    paste0("'{", paste(paste0('"', vec, '"'), collapse=","), "}'")
  }
  
  convert_char <- function(vec) {
    paste0(make_string(vec), "::text[]")
  }
  
  convert_int <- function(vec) {
    paste0(make_string(vec), "::integer[]")
  }
  
  convert_date <- function(vec) {
    paste0(make_string(vec), "::date[]")
  }
  
  convert_num <- function(vec) {
    paste0(make_string(vec), "::float8[]")
  }
  
  temp_starter_sql <- list()
  for (i in 1:length(df)) {
    temp_starter_sql[[i]] = paste0("UNNEST (", 
                                   convert_vec(df[[i]]), ") AS ", 
                                   names(df)[[i]])
  }
  
  temp_sql <- paste0("SELECT ", paste0(temp_starter_sql, collapse = ",\n"))
  
  temp_df_sql <- tbl(conn, sql(temp_sql))
  return(temp_df_sql)
}
```

mgirlich:
Closes #628.

* The columns are cast explicitly because otherwise joins do not work (e.g. in Postgres joins do not work for date columns without explicit casts)

## Notes
* The `VALUES` syntax is similar but not quite the same across databases. I haven't checked the documentation for all DB but I think the only difference is the additional `ROW`
* [MySQL](https://dev.mysql.com/doc/refman/8.0/en/values.html) only supports `VALUES` since 8.0.19
* Would be great to have a collection of websites where to try out different databases...

``` r
devtools::load_all("~/GitHub/dbplyr/")

squirrels <- tibble::tribble(
  ~group,    ~name,     ~role,     ~n_squirrels, ~ n_squirrels2,
  1,      "Sam",    "Observer",   NA,                 1,
  1,     "Mara", "Scorekeeper",    8,                NA,
  2,     "Mike",    "Observer",   NA,                NA,
  2,  "Rachael",    "Observer",   NA,                 6,
  3,     "Kara", "Scorekeeper",    9,                 10,
  3,    "Emily",    "Observer",   NA,                NA,
  3, "Danielle",    "Observer",   NA,                NA
)

squirrels_db <- copy_inline(con, squirrels)
squirrels_db
#> # Source:   SQL [?? x 5]
#>   group name     role        n_squirrels n_squirrels2
#>   <dbl> <chr>    <chr>             <dbl>        <dbl>
#> 1     1 Sam      Observer             NA            1
#> 2     1 Mara     Scorekeeper           8           NA
#> 3     2 Mike     Observer             NA           NA
#> 4     2 Rachael  Observer             NA            6
#> 5     3 Kara     Scorekeeper           9           10
#> 6     3 Emily    Observer             NA           NA
#> 7     3 Danielle Observer             NA           NA

squirrels_db %>% 
  show_query()
```

```sql
SELECT
  CAST(`group` AS REAL) AS `group`,
  CAST(`name` AS TEXT) AS `name`,
  CAST(`role` AS TEXT) AS `role`,
  CAST(`n_squirrels` AS REAL) AS `n_squirrels`,
  CAST(`n_squirrels2` AS REAL) AS `n_squirrels2`
FROM (
  SELECT
    NULL AS `group`,
    NULL AS `name`,
    NULL AS `role`,
    NULL AS `n_squirrels`,
    NULL AS `n_squirrels2`
  WHERE (0 = 1)
  UNION ALL
  VALUES
    (1.0, 'Sam', 'Observer', NULL, 1.0),
    (1.0, 'Mara', 'Scorekeeper', 8.0, NULL),
    (2.0, 'Mike', 'Observer', NULL, NULL),
    (2.0, 'Rachael', 'Observer', NULL, 6.0),
    (3.0, 'Kara', 'Scorekeeper', 9.0, 10.0),
    (3.0, 'Emily', 'Observer', NULL, NULL),
    (3.0, 'Danielle', 'Observer', NULL, NULL)
) AS `values_table
```

<sup>Created on 2021-07-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>


R/verb-copy-to.R
```diff
@@ -87,6 +87,64 @@ copy_to.src_sql <- function(dest, df, name = deparse(substitute(df)),
...
+
+#' @export
+sql_values.DBIConnection <- function(con, df) {
+  sql_values_clause(con, df)
+}
+
+sql_values_clause <- function(con, df, row = FALSE) {
+  sim_data <- rep_named(colnames(df), list(NULL))
+  cols_clause <- escape(sim_data, con = con, parens = FALSE)
+  null_row_clause <- build_sql("SELECT ", cols_clause, " WHERE false", con = con)
```
krlmlr, line 119:
```suggestion
  null_row_clause <- build_sql("SELECT ", cols_clause, " WHERE 0 = 1", con = con)
```

Not all databases know about logical literals.

R/verb-copy-to.R
```diff
@@ -87,6 +87,64 @@ copy_to.src_sql <- function(dest, df, name = deparse(substitute(df)),
...
+  }
+
+  values <- purrr::map(df, escape, con = con, collapse = NULL, parens = FALSE)
+  rows <- purrr::pmap(
+    unname(values),
+    function(...) {
+      dots <- list(...)
+      out <- sql_vector(dots, con = con, collapse = ", ", parens = TRUE)
+      if (row) {
+        out <- sql(paste0("ROW", out))
```
krlmlr, line 132: Is this missing a space?

```suggestion
        out <- sql(paste0("ROW ", out))
```

R/verb-copy-to.R
```diff
@@ -87,6 +87,64 @@ copy_to.src_sql <- function(dest, df, name = deparse(substitute(df)),
...
+  sim_data <- rep_named(colnames(df), list(NULL))
+  cols_clause <- escape(sim_data, con = con, parens = FALSE)
+  null_row_clause <- build_sql("SELECT ", cols_clause, " WHERE false", con = con)
+
+  if (nrow(df) == 0) {
+    return(null_row_clause)
+  }
+
+  values <- purrr::map(df, escape, con = con, collapse = NULL, parens = FALSE)
+  rows <- purrr::pmap(
```
krlmlr, line 126: Can we use a variant of `rlang::exec(paste, values_with_comma)` here? The new `values_with_comma` would contain the first data item followed by a comma followed by the second data item ... . This would make use of vectorization.

R/verb-copy-to.R
```diff
@@ -87,6 +87,64 @@ copy_to.src_sql <- function(dest, df, name = deparse(substitute(df)),
...
+  sql_values_clause(con, df)
+}
+
+sql_values_clause <- function(con, df, row = FALSE) {
+  sim_data <- rep_named(colnames(df), list(NULL))
+  cols_clause <- escape(sim_data, con = con, parens = FALSE)
+  null_row_clause <- build_sql("SELECT ", cols_clause, " WHERE false", con = con)
+
+  if (nrow(df) == 0) {
+    return(null_row_clause)
```
krlmlr, line 122: How can we achieve type stability in the 0-row case?

krlmlr (commented):
Thanks, this would save quite a few roundtrips to the database!

R/verb-copy-to.R
mgirlich: No, it is intended this way: https://dev.mysql.com/doc/refman/8.0/en/values.html

```sql
VALUES ROW(1,-2,3), ROW(5,7,9), ROW(4,6,8);
```

R/verb-copy-to.R
mgirlich: I used the idea in the new approach, thanks!

R/verb-copy-to.R
mgirlich: Solved with the new approach

mgirlich:
I switched to a new approach which basically generates the following SQL:

``` sql
SELECT CAST("int" AS INTEGER) AS "int", "chr", CAST("date" AS DATE) AS "date", CAST("datetime" AS TIMESTAMP) AS "datetime"
FROM (
  (SELECT NULL AS "int", NULL AS "chr", NULL AS "date", NULL AS "datetime" WHERE 0 = 1)
  UNION ALL
  (VALUES
    (1, 'a', '2021-11-22', '2021-11-22T08:38:21Z'),
    (2, 'b', '2021-11-23', '2021-11-23T08:38:21Z'))
) "values_table"
```

* The inner `SELECT NULL as ...` clause is needed to name the values (e.g. needed for SQLite)
* To avoid having a cast in every single value element, the cast is done in the outer select

The cast is done with `translate_sql(as.<type>(!!col))`. Unfortunately, this doesn't work for Datetimes in SQL :-( So this would need a double dispatch... Do you have a better idea?

The alternative would have been to `DBI::dbQuoteLiteral()` but after you said that it isn't exactly type stable I rather wanted to use `translate_sql()`.

R/verb-copy-to.R
```diff
@@ -87,6 +87,64 @@ copy_to.src_sql <- function(dest, df, name = deparse(substitute(df)),
   invisible(out)
 }
 
+#' @export
+db_values <- function(con, df, name = deparse(substitute(df))) {
```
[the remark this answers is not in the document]
mgirlich: No, so I removed it for now. But it might be nice to have when we have CTEs.

R/verb-copy-to.R
```diff
@@ -87,6 +87,120 @@ copy_to.src_sql <- function(dest, df, name = deparse(substitute(df)),
   invisible(out)
 }
 
+#' @export
+db_values <- function(con, df) {
```
hadley, line 91: `db_values()` doesn't seem to be a terribly evocative name to me. It feels like the right name should somehow be related to `copy_to()`. Or maybe you should be able to request values from with an argument to `copy_to()`? (This could would stay as is, we'd just keep `db_values()` internal)

R/verb-copy-to.R
```diff
@@ -87,6 +87,120 @@ copy_to.src_sql <- function(dest, df, name = deparse(substitute(df)),
...
+  if (ncol(df) == 0) {
+    abort("`df` needs at least one column.")
+  }
+
+  if (!inherits(df, "data.frame")) {
```
hadley, line 96: I'd do this test before testing `ncol`

R/verb-copy-to.R
```diff
@@ -87,6 +87,120 @@ copy_to.src_sql <- function(dest, df, name = deparse(substitute(df)),
...
+  )
+}
+
+select_typed_col <- function(type, col, con) {
+  UseMethod("select_typed_col")
+}
+
+#' @export
+select_typed_col.default <- function(type, col, con) {
+  translate_sql(!!col, con = con)
```
hadley, line 161: This feels weird to me — maybe it would be better to create a new translation function that converts from the R type to the database type? And then all the `as.*` function would use that instead of managing their own SQL casts.

Something like:

```R
sql_type <- function(con, type) {
  UseMethod("sql_type")
}

sql_type.DBIConnection <- function(con, type) {
  switch(type, 
    numeric = "NUMERIC",
    double = "NUMERIC",
    integer = "INTEGER",
    character = "TEXT",
    logical = "BOOLEAN",
    Date = "DATE",
    POSIXct = "TIMESTAMP",
    integer64  = "BIGINT",
    stop("Unsupported type")
  )
}
sql_type.odbc <- function(con, type) {
  switch(type, 
    number = "DOUBLE",
    double = "DOUBLE",
    NextMethod()
  )
}
```

Does that make sense?

R/verb-copy-to.R
mgirlich: Imho the following speaks against adding an argument to `copy_to()`:
* Many arguments do not make sense: `name` (would only make sense in a CTE), `overwrite`, `temporary`, `unique_indexes`, `indexes`, `analyze`, `in_transaction`
* The naming is a bit weird to me. On the one hand it allows using the local table in the database but on the other hand it does not really copy it to the database.
* I always find it more difficult to use method specific arguments because it is more difficult to look them up and there is no autocomplete

Maybe something like `copy_to_inline()` or `copy_to_sql()`?

R/verb-copy-to.R
mgirlich: So, `sql_cast()` would basically become

```r
sql_cast <- function(con, r_type) {
  sql_type <- sql_type(con, r_type)
  function(x) {
    sql_expr(cast(!!x %as% !!sql_type))
  }
}
```

Basically, `sql_type()` would be a reimplementation of `DBI:::dbDataType()`, right? E.g. look at the [`RSQLite::dbDataType()`](https://github.com/r-dbi/RSQLite/blob/main/R/dbDataType_SQLiteDriver.R) or [`RPostgres:::get_data_type()`](https://github.com/r-dbi/RPostgres/blob/e9f0366742217b9337979e914215419000970629/R/PqConnection.R).

mgirlich: What should `sql_type(con, type)` look like?

1. `type` a type string, e.g. `"Date"` or `"integer"` (like you proposed).
2. `type` a prototype. e.g. `1L` or `Sys.Date()` (like `DBI::dbDataType()`)

The second approach is easier to use in `db_values()`, the first a bit easier for the translations e.g. `as.numeric = sql_cast("Date")` vs `as.numeric = sql_cast(sql_type(vctrs::new_date()))`.

R/verb-copy-to.R
hadley: Those arguments make sense to me.

Maybe `copy_inline()`? Or `bundle_to()`?

R/verb-copy-to.R
hadley: Maybe have a mutually exclusive `type` and `ptype` arguments so you could call `sql_type(con, type = "numeric")` or `sql_type(con, ptype = Sys.Date())`?

mgirlich: Looking at the different backends I think that it is necessary to use the translation of the `as.*()` functions. For example for MySQL the translation of `as.logical()` is defined as

```r
as.logical = function(x) sql_expr(IF(x, TRUE, FALSE))
```

or MS SQL

```r
as.integer = function(x) {
  sql_expr(try_cast(try_cast(!!x %as% NUMERIC) %as% INT))
}
```

mgirlich: So, for MySQL we want this R code

```r
copy_inline(
  simulate_mysql(),
  tibble(
    lgl = TRUE,
    date = Sys.Date()
  )
) %>% 
  show_query()
```

to result in this SQL code

```SQL
SELECT IF(`lgl`, TRUE, FALSE) AS `lgl`, CAST(`date` AS DATE) AS `date`
FROM (
  (
    SELECT NULL AS `lgl`, NULL AS `date`
    WHERE (0 = 1)
  )
  UNION ALL
  (  VALUES (TRUE, '2022-03-18'))
) `values_table`
```

This does not work with `sql_type()` but only by translating `as.logical()` etc. This looks ugly but I don't really see a way around it.

hadley: Ah good point. The S3 dispatch still feels a bit weird to me, but I don't see a simpler way of doing it, so I think it'd just be worth noting why this design is necessary in a comment.

mgirlich:
Should the first argument be a connection, a lazy table or - like in `copy_toI()` - a remote data source (though they are superseded)?

hadley (approved):

hadley:
A connection feels right to me.

PR #682, opened 2021-07-23
Fix typo in shiny object name

fh-kpikhart:
I hit this error while building a shiny app running on a lazy_tbl:
```
ERROR: Cannot translate shiny inputs to SQL.
* Force evaluation in R with (e.g.) `!!inputs$x` or `local(inputs$x)`
```

When I tried the proposed solution, it did not work. This is because the object is called `input$x` not `inputs$x`


hadley:
Thanks!

PR #691, opened 2021-08-16
Early return `select()`, `filter()`, and `arrange()`

mgirlich:
Return early if the operation doesn't change the data.


R/verb-select.R
```diff
@@ -101,6 +101,10 @@ simulate_vars <- function(x) {
 # op_select ---------------------------------------------------------------
 
 op_select <- function(x, vars) {
+  if (can_select_return_early(x, vars)) {
```
hadley, line 104: Again, I think I'd rather give this a more descriptive name, like `selects_same_variables()` maybe?

tests/testthat/test-verb-arrange.R
```diff
@@ -76,6 +76,12 @@ test_that("can combine arrange with dual table verbs", {
   })
 })
 
+test_that("only add step if necessary", {
+  lf <- lazy_frame(x = 1:3, y = 1:3)
+  expect_equal(lf %>% arrange(), lf)
+  expect_equal(lf %>% filter(!!!list()), lf)
```
hadley, line 82: Why test `filter` here?

tests/testthat/test-verb-filter.R
```diff
@@ -29,6 +29,12 @@ test_that("each argument gets implicit parens", {
   expect_equal_tbl(mf1, mf2)
 })
 
+test_that("only add step if necessary", {
+  lf <- lazy_frame(x = 1:3, y = 1:3)
+  expect_equal(lf %>% filter(), lf)
+  expect_equal(lf %>% filter(!!!list()), lf)
```
hadley, line 35: Why can't this be `lf %>% filter()`?

hadley (approved):
I've approved this so you're free to merge, but do take a look at the comment and make changes if needed (but I don't need to review again, just use your best judgement)

tests/testthat/test-verb-filter.R
mgirlich: Hmm, there was a case where `some_verb()` (I don't know which verb it was) and `some_verb(!!!list())` was different. But I don't see where this should have been the case so I remove these tests.

PR #705, opened 2021-09-17
Fix `sql_expr_matches.DBIConnection()`

### LINKED ISSUE #704: `na_matches = "na"` doesn't work for DuckDB
``` r
library(dbplyr)

con <- DBI::dbConnect(duckdb::duckdb())
x <- copy_to(
  src_dbi(con),
  tibble(x = 1),
  name = "dummy"
)
left_join(x, x, by = "x", na_matches = "na") %>% 
  show_query()
#> <SQL>
#> SELECT "LHS"."x" AS "x"
#> FROM "dummy" AS "LHS"
#> LEFT JOIN "dummy" AS "RHS"
#> ON (CASE WHEN ("LHS"."x" = "RHS"."x") OR ("LHS"."x" IS NULL AND "RHS"."x" IS NULL) THEN 0 ELSE 1 = 0)
left_join(x, x, by = "x", na_matches = "na")
#> Error in .local(conn, statement, ...): duckdb_prepare_R: Failed to prepare query SELECT *
#> FROM (SELECT "LHS"."x" AS "x"
#> FROM "dummy" AS "LHS"
#> LEFT JOIN "dummy" AS "RHS"
#> ON (CASE WHEN ("LHS"."x" = "RHS"."x") OR ("LHS"."x" IS NULL AND "RHS"."x" IS NULL) THEN 0 ELSE 1 = 0)
#> ) "q01"
#> LIMIT 11
#> Error: Parser Error: syntax error at or near ")"
#> LINE 5: ...LEFT JOIN "dummy" AS "RHS"
#> ON (CASE WHEN ("LHS"."x" = "RHS"."x") OR ("LHS"."x" IS NULL AND "RHS"."x" IS NULL) THEN 0 ELSE 1 = 0)
#>                                                   ^
```

<sup>Created on 2021-09-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mgirlich:
Fixes #704.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dbplyr (development version)
 
+* Joins with `na_matches = "na"` now work for DuckDB (@mgirlich, #704).
+
 * `nesting()` now supports the `.name_repair` argument (@mgirlich, #654).
 
 * The SQL translation of `quantile()` now doesn't error anymore when using
--- a/R/db-sql.R
+++ b/R/db-sql.R
@@ -82,7 +82,8 @@ sql_expr_matches.DBIConnection <- function(con, x, y) {
   build_sql(
     "CASE WHEN (", x, " = ", y, ") OR (", x, " IS NULL AND ", y, " IS NULL) ",
     "THEN 0 ",
-    "ELSE 1 = 0",
+    "ELSE 1 = 0 ",
+    "END",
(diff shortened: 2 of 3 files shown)


krlmlr (approved):
Thanks, good catch!

PR #715, opened 2021-10-06
Better error message in `summarise()`

### LINKED ISSUE #710: summarize fails with two `across` constructs
This breaks with master:

```r
library(dbplyr)
df <- memdb_frame(a = 1:3, b = 10:12)
vars <- c("a", "b")
summarise(df,
          across(!!vars, mean, .names = "{.col}"),
          across(!!vars, length, .names = "{.col}.N")) %>%
  show_query()
## Error: `a.N` refers to a variable created earlier in this summarise().
## Do you need an extra mutate() step?
## Backtrace:
##     ▆
##  1. ├─... %>% show_query()
##  2. ├─dplyr::show_query(.) at magrittr/R/pipe.R:136:2
##  3. ├─dplyr::summarise(...)
##  4. └─dbplyr:::summarise.tbl_lazy(...)
##  5.   └─dbplyr:::check_summarise_vars(dots)

```

Replacing `"{.col}"` with `"{.col}.mean"` does work, but from the error it is far from clear what the problem might be and how to fix it.

mgirlich:
Closes #710. I left out a NEWS bullet as this seemed pretty minor but if you want I'll add one.

### DIFF
--- a/tests/testthat/test-verb-summarise.R
+++ b/tests/testthat/test-verb-summarise.R
@@ -18,10 +18,7 @@ test_that("summarise performs partial evaluation", {
 
 test_that("can't refer to freshly created variables", {
   mf1 <- lazy_frame(x = 1)
-  expect_error(
-    summarise(mf1, y = sum(x), z = sum(y)),
-    "cannot use a variable created"
-  )
+  expect_snapshot_error(summarise(mf1, y = sum(x), z = sum(y)))
 })
 
 test_that("summarise(.groups=)", {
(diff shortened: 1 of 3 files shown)


tests/testthat/test-verb-summarise.R
```diff
@@ -20,7 +20,7 @@ test_that("can't refer to freshly created variables", {
   mf1 <- lazy_frame(x = 1)
   expect_error(
     summarise(mf1, y = sum(x), z = sum(y)),
```
hadley, line 22: Can you switch this to a snapshot test? I think reviewing the full generated error is useful in these cases.

mgirlich: 👍

hadley (approved):

Issue #725, opened 2021-11-22
Only select necessary columns in nested queries


ben1787:
Currently when you instantiate a table with tbl(conn,'table_name') it creates a select a, b, c from table_name with all columns referenced.

With a columnar database pulling fewer columns make for a faster query. It would be great both from that perspective and from a clean query perspective, for dbplyr to be able to drop unused columns from the query.

For example: 
table = tbl(conn,'poes')
table %>% 
group_by(id) %>% 
summarise(a = max(username, na.rm=T)) %>% 
sql_render()

<SQL> SELECT "id", MAX("username") AS "a"
FROM (SELECT "tickerexchange", "tickervalue", "confidence", "createddate", "updateddate", "deleted", "username", "fund", "tickernoteid", "currencycode", "createdby", "tickervalueloc", "systemcreateddate", "systemupdateddate", "subfund", "confidenceoverride", "lastfileupdateddate", "islatestrecord", "id", "overridestatus", "convictionoverridereason", "initial_ticker", "system_created_datetime", "client_created_datetime"
FROM quant.dbo.poes) "dbplyr_002"
GROUP BY "id"

Ideally this query should instead look like:
<SQL> SELECT "id", MAX("username") AS "a"
FROM (SELECT "username", "id"
FROM quant.dbo.poes) "dbplyr_002"
GROUP BY "id"

mgirlich:
With the dev version of dbplyr your query already is way shorter. Further optimisations are likely to come this year.

Issue #722, opened 2021-11-05
Produce less nested SQL for joins


mgirlich:
1. Use `SELECT`, `WHERE`, ... directly in the join query

``` r
library(dplyr, warn.conflicts = FALSE)
library(dbplyr, warn.conflicts = FALSE)

lf <- lazy_frame(x = 1, a = 1)
lf2 <- lazy_frame(x = 1, b = 2)
lf3 <- lazy_frame(x = 1, c = 3)

lf2 <- lazy_frame(x = 1, z = 3)
left_join(lf, lf2, by = "x") %>% 
  mutate(y = y + 1, z = z + 1)
#> <SQL>
#> SELECT `x`, `y` + 1.0 AS `y`, `z` + 1.0 AS `z`
#> FROM (SELECT `LHS`.`x` AS `x`, `y`, `z`
#> FROM `df` AS `LHS`
#> LEFT JOIN `df` AS `RHS`
#> ON (`LHS`.`x` = `RHS`.`x`)
#> ) `q01`

# could be
# SELECT `LHS`.`x` AS `x`, `y` + 1.0 AS `y`, `z` + 1.0 AS `z`
# FROM `df` AS `LHS`
# LEFT JOIN `df` AS `RHS`
# ON (`LHS`.`x` = `RHS`.`x`)
```

This is a bit tricky because we in `sql_build()` we have to translate the symbol `x` to `"`LHS`.`x`" and this has to be done everywhere. For example in `sql_build.op_filter()` we need to pass this alias to `translate_sql_()`

```r
where_sql <- translate_sql_(op$dots, con, context = list(clause = "WHERE"))
```

2. Allow multiple joins without nesting

``` r
left_join(lf, lf2, by = "x") %>% 
  left_join(lf3, by = "x")
#> <SQL>
#> SELECT `LHS`.`x` AS `x`, `a`, `b`, `c`
#> FROM (SELECT `LHS`.`x` AS `x`, `a`, `b`
#> FROM `df` AS `LHS`
#> LEFT JOIN `df` AS `RHS`
#> ON (`LHS`.`x` = `RHS`.`x`)
#> ) `LHS`
#> LEFT JOIN `df` AS `RHS`
#> ON (`LHS`.`x` = `RHS`.`x`)

# coul be
#> SELECT `LHS`.`x` AS `x`, `a`, `b`, `c`
#> FROM `df` AS `LHS`
#> LEFT JOIN `df` AS `RHS`
[...]

Issue #721, opened 2021-11-04
Produce shorter SQL


mgirlich:
The generated SQL is way more nested than it would have to be. It would be great if we could shorten the queries 

## Nested Mutates
This is probably not very common but maybe there is an easy fix.

``` r
library(dplyr, warn.conflicts = FALSE)
library(dbplyr, warn.conflicts = FALSE)

lf <- lazy_frame(x = 1, y = 2)
lf %>% 
  mutate(
    x = x + 1,
    x = x + 1,
    y = y + 1,
    y = y + 1
  )
#> <SQL>
#> SELECT `x`, `y` + 1.0 AS `y`
#> FROM (SELECT `x` + 1.0 AS `x`, `y` + 1.0 AS `y`
#> FROM (SELECT `x` + 1.0 AS `x`, `y`
#> FROM `df`) `q01`) `q02`

# could be
# SELECT `x` + 1.0 AS `x`, `y` + 1.0 AS `y`
# FROM (SELECT `x` + 1.0 AS `x`, `y` + 1.0 AS `y`
# FROM `df`) `q01`
```

## Group Followed by Multiple Mutates

```r
lf %>% 
  group_by(x) %>% 
  mutate(x = x + 1) %>% 
  mutate(y = y + 1)
#> <SQL>
#> SELECT `x`, `y` + 1.0 AS `y`
#> FROM (SELECT `x` + 1.0 AS `x`, `y`
#> FROM `df`) `q01`

# could be
# SELECT `x` + 1.0 AS `x`, `y` + 1.0 AS `y`
# FROM `df`
```

## Mutate Followed by Filter
This feels pretty common to me

``` r
lf %>% 
  mutate(x = x + 1) %>% 
  filter(y == 1)
#> <SQL>
#> SELECT *
#> FROM (SELECT `x` + 1.0 AS `x`, `y`
#> FROM `df`) `q01`
#> WHERE (`y` = 1.0)

# could be
# SELECT `x` + 1.0 AS `x`, `y`
# FROM `df`
# WHERE (`y` = 1.0)
```

## Join Followed by Anything
This feels pretty common to me

``` r
lf2 <- lazy_frame(x = 1, z = 3)
left_join(lf, lf2, by = "x") %>% 
  mutate(y = y + 1, z = z + 1)
#> <SQL>
#> SELECT `x`, `y` + 1.0 AS `y`, `z` + 1.0 AS `z`
[...]

Issue #719, opened 2021-10-22
C stack usage is too close to the limit


p-schaefer:
I'm having an issue that might be related to #482. I'm developing a shiny app to interface with a a mySQL database with 120 tables. For some queries I have to join upwards of 20 tables from which I then filter the data. But at some point, as I stack filters on, I get an error about C Stack Usage too close to the limit, and I believe it has to do with the length of the sql query.  

I'm using dbplyr's ```left_joins()``` and  ```filter()``` in for loops to accomplish this, is there perhaps a better way?

Sorry I don't know the best way of generating an example.

tomjemmett:
I've just had a very similar error occur, I'm query against a Microsoft Sql Server database on Windows, but here are steps I can use to reproduce the error (at least, pseudo steps, actual database/column names are redacted):

``` r
# this fails
my_tb <- tbl(con, "table") |> filter(stuff == "thing")
# but this succeeds, using %>% instead of |>
my_tb <- tbl(con, "table") %>% filter(stuff == "thing")
# strangely, this works
tbl(con, "table") |> filter(stuff == "thing")
```

hadley:
Can you please provide a minimal reprex (*repr*oducible *ex*ample)? The goal of a reprex is to make it as easy as possible for me to recreate your problem so that I can fix it: please help me help you! If you've never heard of a reprex before, start by reading [about the reprex package](https://reprex.tidyverse.org/index.html), including the advice further down the page. Please make sure your reprex is created with the reprex package as it gives nicely formatted output and avoids a number of common pitfalls.

mgirlich:
I created a reprex

```r
library(dbplyr)
library(dplyr)
lf <- lazy_frame(x = 1)

gen_nested_lf <- function(n) {
  out <- lf
  for (i in 1:n) {
    out <- left_join(out, lf, by = "x") %>% 
      filter(x == 1)
  }
  
  out
}

query <- remote_query(gen_nested_lf(72))
```

With CRAN dbplyr the maximum that works for me is `n = 72`. The dev version (probably due to the new render pipeline) gives a slightly different error and works until `n = 141`. I think this is reasonable and I might come back to this later.
Another issue is that at some point the query parser of the database produces errors because the query gets too long. So, to fix this properly is quite a bit of work I'm afraid.

As a workaround you can use `compute()`, e.g. something like

```r
out <- lf
for (i in 1:n) {
  out <- left_join(out, lf, by = "x") %>% 
    filter(x == 1)
[...]

PR #726, opened 2021-11-24
New render pipeline

mgirlich:
## Current Translation Approach

The translation currently consists of two stages:

1. The "operation" stage: Whenever a "basic" `dplyr` verb is applied to a lazy table it adds an operation describing the verb and its argument to the `ops` field of the lazy table. This quickly results in a deeply nested list.
2. `sql_build()` creates a query object (`select_query()`, `join_query`, `semi_join_query`, or `set_op_query`) which is a container containing SQL for e.g. `SELECT` or `WHERE`
3. `sql_optimise()` tries to combine the query objects generated in the previous step to generate less nested SQL
4. `sql_render()` converts the query objects to actual SQL for the corresponding backend

Issues with the approach:
* the operations list is quickly deeply nested resulting in a big call stack in `sql_build()`
* many optimisations cannot be applied anymore as it would require re-rendering the SQL generated in the previous step
* Things like using CTEs is a bit hacky

## New Approach

Replace the `ops` field by a `query_list` field. This `query_list` consists of "lazy" query objects which are similar to the current query objects but they are containers for quosures instead of containers for SQL.

The new pipe basically looks as follows:

1. Each "basic" `dplyr` verb adds a or modifies the current "lazy" query object to the `query_list` field of the lazy tables
2. `sql_optimise()` removes unnecessary variables, tries to merge query objects, and removes trivial ones
3. `sql_build()` iterates over `query_list` and translates the quosures to SQL. It produces the current query objects
4. `sql_render()` can stay unchanged.

This should allow to tackle the following issues
* #725 
* #722 
* #721 
* #719

And help in these PRs
* #656 
* #637

Overall, this should allow to produce much more natural SQL.

### DIFF
--- a/R/query2.R
+++ b/R/query2.R
@@ -1,0 +1,45 @@
+select_query2 <- function(from,
+                          select = list(),
+                          where = list(),
+                          group_by = list(),
+                          order_by = list(),
+                          limit = NULL,
+                          distinct = FALSE) {
+  vctrs::vec_assert(from, character(), 1L)
+  assert_list_of_sql_parts(select)
+  assert_list_of_sql_parts(where)
+  assert_list_of_sql_parts(group_by)
+  assert_list_of_sql_parts(order_by)
+  stopifnot(is.null(limit) || (is.numeric(limit) && length(limit) == 1L))
+  stopifnot(is.logical(distinct), length(distinct) == 1L)
+
+  structure(
+    list(
+      from = from,
+      select = select,
+      where = where,
+      group_by = group_by,
+      order_by = order_by,
+      distinct = distinct,
+      limit = limit
+    ),
+    class = c("select_query2", "query2")
+  )
+}
+
+assert_list_of_sql_parts <- function(x) {
+  vctrs::vec_assert(x, list())
+
+  all_valid <- purrr::every(x, is_sql_part)
+  if (!all_valid) {
+    abort("Every part of `x` must be a quosure, a symbol, or sql.")
+  }
+}
+
+is_sql_part <- function(x) {
+  if (rlang::is_quosure(x)) return(TRUE)
+  if (rlang::is_symbol(x)) return(TRUE)
+  if (is.sql(x)) return(TRUE)
+
+  FALSE
+}


ben1787:
This would be great

mgirlich:
* `from`: one of the following
  * reference to an umodified lazy table e.g. from `tbl_lazy()`
  * reference to a database table e.g. from `tbl()` or `copy_to()`
  * reference to another query in the query list
* `select`: a tibble with the following columns
  * name: output name (string) of the column
  * expr: a quosure or an expression describing the expression for the output column
  * group_vars: a character string describing the grouping variables in case `expr` has a window function
  * order_vars: a character string describing the ordering variables in case `expr` has a window function
  * frame: two integerish numbers describing the frame in case `expr` has a window function
* `where`:
  * unnamed list of quosures
* `group_by`
  * unnamed list of symbols
  * this could use quosures in the future to allow inline some code
* `order_by`
  * unnamed list of symbols
  * this could use quosures in the future to allow inline some code
* `having`
  * currently not used in `dbplyr` but could be used in a future step
* `distinct`: if `TRUE` use `SELECT DISTINCT`
* `limit`: number of rows to return
* `group_vars`: current grouping vars
* `group_order`: current ordering vars
* `frame`: current frame


The query fields `group_vars`, `group_order`, and `frame` refer to the currently set grouping variables, ordering variables and window frame. The field `group_by` is only filled when it is actually required, that is after:
* `summarise()`
* `filter()` with a window function

The columns `group_vars`, `group_order`, and `frame` in `select` are filled after:
* `mutate()` with a window function

mgirlich:
# Effect of Basic Verbs

## `relocate()`
* reorder tibble in `$select`
* does not affect order of grouping variables (also it doesn't matter)

## `select()` to Drop Columns
* slice tibble in `$select`
* cannot drop grouping columns, so this does not affect `$group_vars`

## `rename()`
* modify column `name` in `$select`
* do not modify grouping variables
* TODO: a potential optimisation for the future would be to recognise that a variable is only renamed, modify the dots in `mutate()`, `filter()`, or `summarise()` using the renamed variable and then try to inline them

## `filter()`
* if `...` use a window function
  * with `mutate()` create a temporary column for all window functions
  * replace all window functions in `...` with the newly created columns
  * add a new lazy query to the query list with the new dots
  * TODO: this could use `HAVING`
* new query if `$select` uses a window function because `WHERE` is evaluated before `SELECT`
* new query if `...` use a column that is mutated in `$select`. TODO: if the column is only renamed it would be possible to instead modify `...` and not use a new query this way.
* TODO `$order_by` resp. `$order_vars` might matter, e.g. if they use a window function
* an existing `$where` can be merged
* `$group_by` does not matter because window functions cannot be used in `WHERE` and the case where `...` use window functions is handled separately
* new query if `$distinct` is `TRUE` and `...` use a window function

## `summarise()`
* `$group_by` is filled with `$group_vars`
* `$where` doesn't matter as `WHERE` is evaluated before `SELECT`
* `$group_by` doesn't matter as `GROUP BY` is evaluated before `SELECT`
* needs a new query if `$select` is not trivial
* new query if `$distinct` is `TRUE`

## `mutate()`/`transmute()`
* `$where` doesn't matter as `WHERE` is evaluated before `SELECT`
* `$group_by` doesn't matter as `GROUP BY` is evaluated before `SELECT`
* `$select`
  * new query if variable mutated in `...` is not trivial and not renamed
  * new query if `...` uses mutated variables
  * new query if `...` use window function and a group var is mutated
* for each expr in `...` that uses a window function
  * fill columns `group_vars`, `order_vars` and `frame` with the current value of the respective fields
* new query if `$distinct` is `TRUE`

## `distinct()`
* never needs a new query?

## `group_by()`
* overwrite current `$group_vars`

## `arrange()`
* overwrite current `$order_vars`

mgirlich:
For two table verbs there are two options

1. Merge the two query lists into a single, flat query list.
2. Nest the query list, basically like it is currently done.

Option 2 is definitely much easier to begin with and should make it easier to convert the render pipe. So, let's start with option 2.

mgirlich:
## Sorting
There are two ways to "sort": `window_order()` which does not generate an `ORDER BY` clause and `arrange()` which generates an `ORDER BY` clause.

Get current ordering for `SELECT` clause via `op_sort()`. It is called by
* `slice_by()` -> with `window_order()`
* `fill()`
* `collapse()`,`compute()` -> restore sort
* `sql_build.op_select()`
* `tbl_sum.tbl_sql()`
* `op_sort.op_single()`, `op_sort.op_double()`, `op_sort.tbl_lazy()`

defined order
* op_arrange -> c(op_sort(op$x), op$dots)
* op_order -> op$dots

no order
* op_base
* op_summarise

pass on
* op_single
* op_double
* tbl_lazy

Notes:
* current implementation is buggy and doesn't handle renaming

``` r
library(dplyr)
library(dbplyr)

lazy_frame(x = 1) %>% 
  window_order(x) %>% 
  rename(ax = x) %>% 
  mutate(xcum = cumsum(x))
#> <SQL>
#> SELECT `ax`, SUM(`x`) OVER (ORDER BY `x` ROWS UNBOUNDED PRECEDING) AS `xcum`
#> FROM (SELECT `x` AS `ax`
#> FROM `df`) `q01`
```

<sup>Created on 2021-12-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

## Grouping
Get current grouping via `op_grps()`. Called in

* `op_vars.op_summarise()` -> summarise keeps grouping variables
* `op_grps.op_summarise()`
* `sql_build.op_summarise()` -> `GROUP BY` clause
* `mutate.tbl_lazy()` -> used for `all_vars` argument?
* `distinct.tbl_lazy()` -> distinct on grouping variables + specified variables
* `fill()`
* `compute()`, `collapse()`, `collect()` -> restore groups

defined grouping
* op_group_by
```r
if (isTRUE(op$args$add)) {
  union(op_grps(op$x), names(op$dots))
} else {
  names(op$dots)
}
```

* op_select: rename
* op_summarise: drop last, keep, or keep all

no group
* op_base
* op_ungroup

pass on
* op_single
* op_double
* tbl_lazy

Notes:
* `collect()` is confusing: `intersect(op_grps(x), names(out))`
* `mutate.tbl_lazy()` is confusing: `nest_vars(all_vars = union(op_vars(.data), op_grps(.data)))`. Maybe to keep group var `x` even when doing `mutate(x = NULL)`?

## Frame
Get current frame via `op_frame()`

defined frame
* op_frame -> op$args$range

no frame
* op_base

pass on
* op_single
* op_double
* tbl_lazy

mgirlich:
Closed in favour of #729

PR #728, opened 2021-12-03
Lazy render pipeline

mgirlich:
This is the first PR of optimising the render pipe line and producing shorter SQL.
The goals of this PR are:

* do not change the generated SQL
* use "lazy queries" instead of `op` objects for rendering
* only convert R code to SQL when it is actually necessary, i.e. when the user `collect()`s or uses `show_query()`

The PR is quite big as basically every verb has to be touched. I remove all the code around `op` object. We could also keep it for a while so that the user can switch back to the "old" pipeline via an option (this can easily be done in `sql_render.tbl_lazy()`). Also this might break some packages (e.g. `sparklyr` currently uses it in some places).

## Overview
A lazy table now stores a nested list of "lazy queries" in its field "lazy_query". The lazy queries do not store SQL code but all the information necessary to generate the correct SQL. There are six lazy queries:

* `lazy_query_local` and `lazy_query_remote` are the equivalent of `op_base_*`. They are used to store the information about a local resp. database table.
* `lazy_select_query`, `lazy_join_query`, `lazy_semi_join_query`, and `lazy_set_op_query` are the lazy versions of `select_query` etc. The last three are actually (currently) identical to their non-lazy counterparts because the non-lazy versions already do not store SQL.

The `lazy_select_query` has the following fields similar to `select_query`

* `from`: a lazy query
* `select`: a tibble with the following columns
  * `name`: output name (string) of the column
  * `expr`: a quosure or an expression describing the expression for the output column
  * `group_vars`: a character string describing the grouping variables in case `expr` has a window function
  * `order_vars`: a quosure or an expressions describing the ordering expressions in case `expr` has a window function
  * `frame`: two integerish numbers describing the frame in case `expr` has a window function
* `where`: an unnamed list of quosures
* `group_by`
  * an unnamed list of symbols
  * this could use quosures in the future to allow inline some code
* `order_by`
  * an unnamed list of quosures
* `distinct`: if `TRUE` use `SELECT DISTINCT`
* `limit`: number of rows to return

and the following additional fields
* `last_op`: the most recent applied operation; this is (currently) needed for optimisations in `mutate()`, `select()`, and `head()`
* `group_vars`: current grouping vars; an unnamed list of symbols
* `group_order`: current ordering vars; an unnamed list of quosures
* `frame`: current frame; two integerish numbers
* `select_operation`: either `"mutate"` or `"summarise"`. This is needed to know how to generate the SQL code in the `SELECT` clause
* `message_summarise`: needed to inform about new groups after `summarise()`

## Next Steps
* `sql_optimise()` removes unnecessary variables from subqueries
* `filter()` checks if it can avoid adding a new query
* `mutate()`/`transmute()` check if it can avoid adding a new query. This should use an improved `get_mutate_layers()` from #671.
* Use `HAVING` clause
* Applying a verb on a join does not result in a new query (i.e. actually use the `SELECT` clause etc after a join)

## Notes
* In `test-verb-arrange.R` there is the following test

```r
"warn if arrange before join"
lf %>% arrange(a) %>% left_join(rf)
lf %>% arrange(a) %>% semi_join(rf)
lf %>% arrange(a) %>% union(rf)
```

The last test with `union()` did not produce a warning before and does not do it now

* In some cases multiple `arrange()`s seem to combine, in others not. I think they should not combine but this should be clarified. Also see https://github.com/tidyverse/dbplyr/pull/672

* Fixed a bug where columns dropped with `NULL` where still appeared in `op_vars`

``` r
library(dplyr)
library(dbplyr)

lazy_frame(x = 1, y = 1) %>%
  mutate(y = NULL) %>% 
  op_vars()
#> [1] "x" "y"
```

<sup>Created on 2021-12-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


tests/testthat/test-verb-arrange.R
```diff
@@ -1,4 +1,5 @@
 test_that("two arranges equivalent to one", {
+  skip("not clear whether test makes sense")
```
mgirlich, line 2: I don't think a warning is required here.

tests/testthat/test-verb-arrange.R
```diff
@@ -122,6 +123,7 @@ test_that("arranges captures DESC", {
 })
 
 test_that("multiple arranges combine", {
+  skip("not clear whether test makes sense")
```
mgirlich, line 126: `out` is defined twice and I don't think that multiple `arrange()` should really combine.

mgirlich:
@krlmlr If you have some spare time, I'd be happy if you could look over this PR. This would also allow a simpler implementation of CTEs #656.

R/partial-eval.R
```diff
@@ -70,7 +70,7 @@ partial_eval <- function(call, vars = character(), env = caller_env()) {
   }
 }
 
-partial_eval_dots <- function(dots, vars) {
+partial_eval_dots <- function(dots, vars, named = FALSE) {
```
mgirlich, line 73: Currently, `partial_eval_dots()` does not name `across()` calls. The names are needed for `mutate()` and `transmute()` but e.g. for `filter()` the dots must not be named.

R/verb-compute.R
```diff
@@ -18,7 +18,7 @@ collapse.tbl_sql <- function(x, ...) {
 
   tbl(x$src, sql) %>%
     group_by(!!! syms(op_grps(x))) %>%
-    add_op_order(op_sort(x))
+    arrange.tbl_lazy(!!!op_sort(x))
```
mgirlich, line 21: I'm not sure why this used `add_op_order()` instead of `arrange()`. If there is a good reason one can change this to use `add_arrange()`.

R/verb-head.R
```diff
@@ -31,21 +31,38 @@ head.tbl_lazy <- function(x, n = 6L, ...) {
...
+  x
+}
+
+add_head <- function(x, n) {
+  lazy_query <- x$lazy_query
+  if (!inherits(lazy_query, "lazy_select_query")) {
+    lazy_query <- lazy_select_query(
+      from = lazy_query,
+      last_op = "head",
+      select = syms(set_names(op_vars(lazy_query))),
```
mgirlich, line 44: It might be worth using `syms(set_names(op_vars(lazy_query)))` as the default value. I avoided it first to be more explicit and avoid some bugs but this seems to be the standard case.

R/verb-select.R
```diff
@@ -83,89 +83,76 @@ rename_with.tbl_lazy <- function(.data, .fn, .cols = everything(), ...) {
...
+
+  .data$lazy_query <- add_select(.data, syms(unlist(new_vars)))
   .data
 }
 
 simulate_vars <- function(x) {
   as_tibble(rep_named(op_vars(x), list(logical())))
 }
 
+simulate_vars2 <- function(x) {
```
mgirlich, line 101: This is needed because `relocate()` can also rename variables. This didn't work before

tests/testthat/_snaps/query-select.md
```diff
@@ -13,6 +13,6 @@
     Code
       lf_render
     Output
-      <SQL> SELECT `x`
+      <SQL> SELECT *
```
mgirlich, line 16: This comes from a minor optimisation that was necessary to produce mostly the same SQL as before

tests/testthat/_snaps/verb-arrange.md
```diff
@@ -22,9 +22,6 @@
     Code
       # double arrange
       lf %>% arrange(a) %>% arrange(b)
-    Warning <warning>
```
mgirlich, line 25: I don't think this warning was necessary

tests/testthat/_snaps/verb-arrange.md
```diff
@@ -55,7 +52,6 @@
       <SQL>
       SELECT *
       FROM `df`
-      ORDER BY `a`
```
mgirlich, line 58: Not sure whether `arrange()` should actually "unarrange" or not. At least there was a PR for this #471

tests/testthat/test-translate-sql.R
```diff
@@ -66,7 +66,7 @@ test_that("na_if is translated to NULLIF (#211)", {
 })
 
 test_that("connection affects quoting character", {
-  lf <- lazy_frame(field1 = 1, con = simulate_sqlite())
+  lf <- lazy_frame(field1 = 1, field2 = 2, con = simulate_sqlite())
```
mgirlich, line 69: Needed to add `field2` to not get `SELECT *`

tests/testthat/test-verb-filter.R
```diff
@@ -68,3 +68,48 @@ test_that("filter generates simple expressions", {
 
   expect_equal(out$where, sql('`x` > 1'))
 })
+
+
+# lazy_select_query -------------------------------------------------------
```
mgirlich, line 73: These tests were mostly for developing. Probably they can be removed again

tests/testthat/test-verb-mutate.R
```diff
@@ -165,10 +165,10 @@ test_that("mutate generates simple expressions", {
 
 test_that("mutate can drop variables with NULL", {
   out <- lazy_frame(x = 1, y = 1) %>%
-    mutate(y = NULL) %>%
-    sql_build()
+    mutate(y = NULL)
 
-  expect_named(out$select, "x")
+  expect_named(sql_build(out)$select, "x")
+  expect_equal(op_vars(out), "x")
```
mgirlich, line 171: Before `op_vars()` still included `y`. This is fixed.

tests/testthat/test-verb-select.R
```diff
@@ -122,7 +122,7 @@ test_that("select renames variables", {
 
 test_that("select can refer to variables in local env", {
   vars <- c("x", "y")
-  out <- lazy_frame(x = 1, y = 1) %>%
+  out <- lazy_frame(x = 1, y = 1, z = 1) %>%
```
mgirlich, line 125: Needed to add `z` to not get `SELECT *`

mgirlich:
* Read the notes in the Overview
* Glance over `lazy_select_query()`
* Look at `sql_build.lazy_select_query()`. This converts the `lazy_select_query` in a `select_query`. The translation of the `select` field in `get_select_sql()` is the most relevant part.
* Look at `verb-select.R`, `verb-mutate.R`, `verb-summarise.R`
* Look at the other verb files
* Look at the remaining files (not much happens in them)

ben1787:
@mgirlich Can this be tested out yet? Would love to try using it.

R/backend-postgres.R
```diff
@@ -19,7 +19,7 @@
 #'
 #' lf <- lazy_frame(a = TRUE, b = 1, c = 2, d = "z", con = simulate_postgres())
 #' lf %>% summarise(x = sd(b, na.rm = TRUE))
-#' lf %>% summarise(y = cor(b, c), y = cov(b, c))
+#' lf %>% summarise(y = cor(b, c), z = cov(b, c))
```
mgirlich, line 22: `summarise(df, y = ..., y = ...)` currently does not work correctly (also before this PR). I created an issue for this #785

tests/testthat/_snaps/verb-arrange.md
```diff
@@ -22,10 +22,6 @@
     Code
       # double arrange
       lf %>% arrange(a) %>% arrange(b)
-    Condition
```
mgirlich, line 25: I don't think this warning was actually necessary and it just vanished.

mgirlich:
@ben1787 Yes, you can try it out. Use `remotes::install_github("tidyverse/dbplyr#728")` to install it.
I'm happy about feedback regarding any bugs. Note that it doesn't **yet** provide benefits (apart from the many improvements in the dev version of `dbplyr`) but it is the foundation for improved SQL.

R/lazy-select-query.R
```diff
@@ -0,0 +1,226 @@
...
+  tibble(
+    name = var_names %||% character(),
+    expr = vars %||% list(),
+    group_vars = rep_along(vars, list(group_vars)),
+    order_vars = rep_along(vars, list(order_vars)),
+    frame = rep_along(vars, list(frame))
+  )
+}
+
+update_lazy_select <- function(lazy_query, vars) {
```
hadley, line 92: I think I'd find this easier to understand if it took a `select` object, and modifying the `lazy_query` happened at the next level up.

R/lazy-select-query.R
```diff
@@ -0,0 +1,226 @@
...
+  idx <- vctrs::vec_match(sel_vars, select$name)
+  select <- vctrs::vec_slice(select, idx)
+  select$name <- names(vars)
+  lazy_query$select <- select
+
+  lazy_query
+}
+
+#' @export
+print.lazy_select_query <- function(x, ...) {
```
hadley, line 106: Would be good to add a snapshot test for the print method.

R/verb-compute.R
hadley: It doesn't make much sense to me to use either here, since most RDBMs won't order subqueries anyway. I'd recommend leaving it as is ([Chesterton's fence](https://en.wikipedia.org/wiki/Wikipedia:Chesterton's_fence)) and filing an issue to explore in a separate PR.

R/verb-filter.R
```diff
@@ -23,35 +23,40 @@ filter.tbl_lazy <- function(.data, ..., .preserve = FALSE) {
...
+      last_op = "filter",
+      select = syms(set_names(vars)),
+      where = dots
     )
   } else {
     # Do partial evaluation, then extract out window functions
-    where <- translate_window_where_all(op$dots, ls(dbplyr_sql_translation(con)$window))
+    where <- translate_window_where_all(dots, ls(dbplyr_sql_translation(con)$window))
 
     # Convert where$expr back to a lazy dots object, and then
```
hadley, line 50: Comment needs update?

R/verb-joins.R
```diff
@@ -69,8 +69,7 @@ inner_join.tbl_lazy <- function(x, y, by = NULL, copy = FALSE,
                                 suffix = NULL,
                                 auto_index = FALSE, ...,
                                 sql_on = NULL, na_matches = c("never", "na")) {
-
-  add_op_join(
+  lazy_query <- add_join(
```
hadley, line 72:
```suggestion
  x$lazy_query <- add_join(
```

?

tests/testthat/_snaps/verb-arrange.md
hadley: I think this is important, because in dplyr this would yield the equivalent of `lf %>% arrange(b, a)` because the sorts are stable.

tests/testthat/_snaps/verb-arrange.md
hadley: Agreed that this behaviour seems more consistent. Allow it maybe seems like it should still generate a warning?

hadley (commented):
First round of review. I'll aim to actually start playing with the code on Monday. I'm starting to get a sense of what this change means, it seems like it makes things more consistent.

R/verb-compute.R
mgirlich: Ah, actually `add_op_order()` comes from `window_order()` not from `arrange()`. I think keeping the `window_order()` makes more sense.

mgirlich: If this also makes sense to you, then I change it to `window_order()` and add a test for this.

tests/testthat/_snaps/verb-arrange.md
mgirlich: So should `lf %>% arrange(a) %>% arrange(b)` be equivalent to
a) `lf %>% arrange(b)` or
b) `lf %>% arrange(b, a)`?

And should it generate a warning?

The current behaviour of `dbplyr` is inconsistent

``` r

lf <- dbplyr::lazy_frame(a = 1:3, b = 3:1)
lf %>% arrange(a) %>% arrange(b)
#> <SQL>
#> SELECT *
#> FROM `df`
#> ORDER BY `b`

out <- dbplyr::lazy_frame(x = 1:3, y = 3:1) %>% arrange(x) %>% arrange(y)
out %>% dbplyr::op_sort()
#> [[1]]
#> <quosure>
#> expr: ^x
#> env:  0x7fb0e54abb90
#> 
#> [[2]]
#> <quosure>
#> expr: ^y
#> env:  0x7fb0e54abae8
out %>% show_query()
#> <SQL>
#> SELECT *
#> FROM `df`
#> ORDER BY `y`
```

<sup>Created on 2022-03-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/verb-select.R
```diff
@@ -104,37 +104,66 @@ simulate_vars <- function(x, drop_groups = FALSE) {
...
+  lazy_query <- .data$lazy_query
+
+  # drop NULLs
+  vars <- purrr::discard(vars, ~ is_quosure(.x) && quo_is_null(.x))
+  if (selects_same_variables(.data, vars)) {
+    return(lazy_query)
   }
 
-  if (inherits(x, "op_select")) {
+  if (identical(lazy_query$last_op, "select") || identical(lazy_query$last_op, "mutate")) {
```
hadley, line 125: `lazy_query$last_op %in% c("select", "mutate")`, since `last_op` is always a string?

R/verb-select.R
hadley: I'm not sure I understand the difference here. Doesn't this produce a 1-row tibble with the names of the variables in the cell values? Why doesn't it need `.name_repair` like `simulate_vars()`?

tests/testthat/_snaps/verb-arrange.md
hadley: I think if it returns `df %>% arrange(b, a)` then it wouldn't need a warning? But maybe it's not that important?

hadley (approved):
I'm happy for you to finish this off and merge it. It seems mostly a neutral refactoring to me, but I trust you that it'll make things easier down the road. I can't say that I fully understand it, but it doesn't seem any more complex that the existing implementation, which I also don't fully understand :smile: It seems straightforward enough that I'm confident that I can puzzle it out if needed.

mgirlich:
> It seems mostly a neutral refactoring to me, but I trust you that it'll make things easier down the road.

At least for now it is the purpose to be very neutral and produce the same SQL. The previous (resp. still current) approach with `ops` has these issues:

* the operations list is quickly deeply nested resulting in a big call stack in `sql_build()`.
* pieces of SQL are generated quite early in the pipeline. This makes it more difficult (or impossible) to apply some optimisations for shorter SQL.
* Things like using CTEs is a bit hacky

With the new approach

* the lazy query object can contain the information of multiple ops, which should result in less nested lists and therefore reduce recursion. It should also be easier to avoid deep recursions if it turns out this is still necessary. So this should reduce the issue of #719.
* produces SQL much later in the pipeline which makes it easier to optimise. This should allow to tackle #725, #722, #721.
* The implementation of CTE support got much easier.

R/verb-select.R
mgirlich: The `last_op` field only exists for `lazy_select_query`, so `lazy_query$last_op` can also be `NULL`.

ben1787:
@mgirlich thank you for all the recent contributions!

PR #732, opened 2021-12-14
Indent SQL

mgirlich:
This replace the old PR for indentation. This way the commit history is much cleaner and it was a bit easier to compare the snapshots locally.

The SQL parts after `SELECT`, `GROUP BY`, ... are on the same line if it is only a single part or the total number of characters on the line is at most 80. Otherwise each SQL part is put on its own line with an indentation of two spaces.

## Select query
``` r
tbl_lazy(mtcars) %>% 
  mutate(mpg = mpg + 1,
         mpg = mpg - 1) %>% 
  filter(
    mpg > 1,
    cyl == 6
  ) %>% 
  group_by(mpg, cyl) %>% 
  summarise(mpg = mean(mpg, na.rm = TRUE))
```

before
```sql
SELECT `mpg`, `cyl`, AVG(`mpg`) AS `mpg`
FROM (SELECT `mpg` - 1.0 AS `mpg`, `cyl`, `disp`, `hp`, `drat`, `wt`, `qsec`, `vs`, `am`, `gear`, `carb`
FROM (SELECT `mpg` + 1.0 AS `mpg`, `cyl`, `disp`, `hp`, `drat`, `wt`, `qsec`, `vs`, `am`, `gear`, `carb`
FROM `df`) `q01`) `q02`
WHERE ((`mpg` > 1.0) AND (`cyl` = 6.0))
GROUP BY `mpg`, `cyl`
```

after
```sql
SELECT `mpg`, `cyl`, AVG(`mpg`) AS `mpg`
FROM (
  SELECT
    `mpg` - 1.0 AS `mpg`,
    `cyl`,
    `disp`,
    `hp`,
    `drat`,
    `wt`,
    `qsec`,
    `vs`,
    `am`,
    `gear`,
    `carb`
  FROM (
    SELECT
      `mpg` + 1.0 AS `mpg`,
      `cyl`,
      `disp`,
      `hp`,
      `drat`,
      `wt`,
      `qsec`,
      `vs`,
      `am`,
      `gear`,
      `carb`
    FROM `df`
  ) `q01`
) `q02`
WHERE (`mpg` > 1.0) AND (`cyl` = 6.0)
GROUP BY `mpg`, `cyl`
```

## Union Query
```r
union(
  union(
    tbl_lazy(mtcars) %>% 
      select(mpg, cyl),
    tbl_lazy(mtcars) %>% 
      select(mpg, cyl)
  ),
  tbl_lazy(mtcars) %>% 
    filter(mpg > 1) %>% 
    transmute(mpg = mpg + 1, cyl)
)
```

before

```sql
((SELECT `mpg`, `cyl`
FROM `df`)
UNION
(SELECT `mpg`, `cyl`
FROM `df`))
UNION
(SELECT `mpg` + 1.0 AS `mpg`, `cyl`
FROM `df`
WHERE (`mpg` > 1.0))
```

after

```sql
(
  (
    SELECT `mpg`, `cyl`
    FROM `df`
  )
  UNION
  (
    SELECT `mpg`, `cyl`
    FROM `df`
  )
)
UNION
(
  SELECT `mpg` + 1.0 AS `mpg`, `cyl`
  FROM `df`
  WHERE (`mpg` > 1.0)
)
```


tests/testthat/_snaps/backend-mssql.md
```diff
@@ -96,22 +96,24 @@
...
-      sql_query_select(simulate_mssql(), "x", "y", order_by = "z", subquery = TRUE)
+      sql_query_select(simulate_mssql(), ident("x"), ident("y"), order_by = "z",
+      subquery = TRUE)
     Warning <warning>
       ORDER BY is ignored in subqueries without LIMIT
       i Do you need to move arrange() later in the pipeline or use window_order() instead?
     Output
-      <SQL> SELECT 'x'
-      FROM 'y'
+      <SQL> SELECT `x`
```
mgirlich, line 105: The test passed character "x" and "y" (which does not really make sense) instead of `ident("x")` and `ident("y")`

R/backend-sqlite.R
```diff
@@ -43,11 +43,12 @@ sql_query_explain.SQLiteConnection <- function(con, sql, ...) {
...
 
 #' @export
-sql_query_set_op.SQLiteConnection <- function(con, x, y, method, ..., all = FALSE) {
+sql_query_set_op.SQLiteConnection <- function(con, x, y, method, ..., all = FALSE, lvl = 0) {
   # SQLite does not allow parentheses
+  method <- paste0(method, if (all) " ALL")
   build_sql(
     x,
-    "\n", sql(method), if (all) sql(" ALL"), "\n",
+    "\n", indent_lvl(sql_kw(method), lvl = lvl), "\n",
```
mgirlich, line 51: `sql_kw()` is not really necessary but it might useful to wrap SQL keywords in a method, e.g. for coloured output.
Currently, this is not yet consistent in this PR so depending on what you think I either remove `sql_kw()` or wrap things consistently with `sql_kw()`

R/db-sql.R
```diff
@@ -219,15 +231,16 @@ sql_query_select.DBIConnection <- function(con, select, from, where = NULL,
                                limit = NULL,
                                distinct = FALSE,
                                ...,
-                               subquery = FALSE) {
+                               subquery = FALSE,
+                               lvl = 0) {
```
mgirlich, line 235: It might be nicer to apply apply the level indent in `sql_select_clauses()`. Consequences:

* no need to pass `lvl` to `sql_clause_*()`
* `sql_clause_*()` would need to return a character vector instead of a string
* `sql_select_clauses()` would need to get a `lvl` argument

tests/testthat/_snaps/query-semi-join.md
```diff
@@ -24,7 +24,7 @@
       SELECT * FROM `df` AS `LHS`
       WHERE EXISTS (
         SELECT 1 FROM `df` AS `RHS`
-        WHERE (`LHS`.`x` = `RHS`.`x` AND `LHS`.`y` = `RHS`.`y`)
+        WHERE (`LHS`.`x` = `RHS`.`x`) AND (`LHS`.`y` = `RHS`.`y`)
```
mgirlich, line 27: The new parantheses are because `sql_query_semi_join.DBIConnection()` now uses `sql_clause_where()` instead of directly using `WHERE` in its `build_sql()` clause.

tests/testthat/_snaps/verb-fill.md
```diff
@@ -5,16 +5,25 @@
...
       <SQL>
       SELECT `id`, `group`, MAX(`n1`) OVER (PARTITION BY `..dbplyr_partion_1`) AS `n1`
-      FROM (SELECT `id`, `group`, `n1`, SUM(CASE WHEN (((`n1`) IS NULL)) THEN (0) WHEN NOT(((`n1`) IS NULL)) THEN (1) END) OVER (ORDER BY -`id` ROWS UNBOUNDED PRECEDING) AS `..dbplyr_partion_1`
-      FROM `df`)
+      FROM (
+        SELECT
+          `id`,
+          `group`,
+          `n1`,
+          SUM(CASE WHEN (((`n1`) IS NULL)) THEN (0) WHEN NOT(((`n1`) IS NULL)) THEN (1) END) OVER (ORDER BY -`id` ROWS UNBOUNDED PRECEDING) AS `..dbplyr_partion_1`
```
mgirlich, line 13: This looks even worse now 😬

hadley:
The example makes me wonder if it's worth figuring out how to use `*`, when we're selecting all the variables.

mgirlich:
I actually implemented using `*` to select all variables when working on the lazy render pipeline #728. It's definitely nice to have in some cases.

R/query-set-op.R
```diff
@@ -24,16 +24,18 @@ print.set_op_query <- function(x, ...) {
 }
 
 #' @export
-sql_render.set_op_query <- function(query, con = NULL, ..., subquery = FALSE) {
-  from_x <- sql_render(query$x, con, ..., subquery = FALSE)
-  from_y <- sql_render(query$y, con, ..., subquery = FALSE)
+sql_render.set_op_query <- function(query, con = NULL, ..., subquery = FALSE, lvl = 0) {
+  # TODO improve this hack a little
+  sub_lvl <- lvl + !inherits(con, "SQLiteConnection")
```
mgirlich, line 29: This is necessary because `sql_query_set_op.DBIConnection()` wraps the queries in parantheses whereas `sql_query_set_op.SQLiteConnection()` does not.

R/sql-clause.R
```diff
@@ -15,79 +16,124 @@ sql_select_clauses <- function(con,
...
 
-  build_sql(
-    "SELECT ",
-    if (distinct) sql("DISTINCT "),
-    if (!is.null(top)) build_sql("TOP ", as.integer(top), " ", con = con),
-    escape(select, collapse = ", ", con = con),
+  clause <- build_sql(
+    sql_kw("SELECT"),
+    if (distinct) sql_kw(" DISTINCT"),
+    if (!is.null(top)) build_sql(sql_kw(" TOP "), as.integer(top), con = con),
```
mgirlich, line 41: `con` is only needed for `build_sql()`. But apart from escaping `top` the call to `build_sql()` is not necessary. Integers don't really need to be escaped so I think it is safe to remove `build_sql()` and therefore the `con` argument. What do you think?

R/backend-access.R
```diff
@@ -41,15 +41,17 @@ sql_query_select.ACCESS <- function(con, select, from,
...
+                              lvl = 0) {
 
   sql_select_clauses(con,
     select    = sql_clause_select(con, select, distinct, top = limit),
-    from      = sql_clause_from(con, from),
-    where     = sql_clause_where(con, where),
-    group_by  = sql_clause_group_by(con, group_by),
-    having    = sql_clause_having(con, having),
-    order_by  = sql_clause_order_by(con, order_by, subquery, limit)
+    from      = sql_clause_from(from),
```
hadley, line 49: Why remove `con` here?

R/db-sql.R
```diff
@@ -257,11 +275,11 @@ sql_select.DBIConnection <- function(con, select, from, where = NULL,
 
 #' @rdname db-sql
 #' @export
-sql_query_join <- function(con, x, y, vars, type = "inner", by = NULL, na_matches = FALSE, ...) {
+sql_query_join <- function(con, x, y, vars, type = "inner", by = NULL, na_matches = FALSE, ..., lvl = 0) {
```
hadley, line 278: If a package has provided their own `sql_query_join()` method, this will still work, just the indenting will be incorrect, right?

R/db-sql.R
```diff
@@ -172,17 +172,32 @@ sql_query_save.DBIConnection <- function(con, sql, name, temporary = TRUE, ...)
...
   } else if (is.schema(from)) {
     setNames(as.sql(from, con), name)
   } else {
-    build_sql("(", from, ") ", ident(name %||% unique_subquery_name()), con = con)
+    ident_name <- ident(name %||% unique_subquery_name())
+    build_sql(ident_subquery(from, con, lvl), " ", ident_name, con = con)
+  }
+}
+
+ident_subquery <- function(from, con, lvl) {
```
hadley, line 190: I think we'll need to export this for other packages to use?

R/sql-clause.R
hadley: That seems reasonable. My only concern is that we might eventually hit a SQL variant that does something particularly weird, and we end up needing to re-thread `con` all the way down to the right place. Maybe that's unlikely because the overall form of the `SELECT` query is so highly conserved?

R/sql-clause.R
```diff
@@ -15,79 +16,124 @@ sql_select_clauses <- function(con,
...
     limit = limit
   )
-  out <- purrr::compact(out)
-  escape(unname(out), collapse = "\n", parens = FALSE, con = con)
+  sql_format_clauses(out, lvl, con)
 }
 
-sql_clause_select <- function(con, select, distinct = FALSE, top = NULL) {
+sql_clause <- function(kw, parts, sep = ",", parens = FALSE, lvl = 0) {
+  list(
```
hadley, line 23: Is it worth giving this a class? If only to make debugging/exploration a little easier?

tests/testthat/_snaps/backend-sqlite.md
```diff
@@ -17,15 +16,24 @@
...
-      ON (`LHS`.`x` = `RHS`.`x`)
+      FROM (
+        SELECT
+          COALESCE(`LHS`.`x`, `RHS`.`x`) AS `x`,
+          `LHS`.`y` AS `y.x`,
+          `RHS`.`y` AS `y.y`,
+          `z`
+        FROM `df` AS `LHS`
+        LEFT JOIN `df` AS `RHS`
+        ON (`LHS`.`x` = `RHS`.`x`)
```
hadley, line 27: I think `ON` should either go on the same line as `LEFT JOIN` or be indented?

tests/testthat/_snaps/query-semi-join.md
hadley: LGTM

tests/testthat/_snaps/verb-expand.md
```diff
@@ -109,14 +115,23 @@
...
+            SELECT DISTINCT `x`
+            FROM `df`
+          ) `LHS`
+          CROSS JOIN (
+            SELECT DISTINCT `y`
+            FROM `df`
+          ) `RHS`
+        ) `LHS`
+        FULL JOIN `df` AS `RHS`
+        ON (`LHS`.`x` = `RHS`.`x` AND `LHS`.`y` = `RHS`.`y`)
```
hadley, line 135: Why don't the parens change here?

tests/testthat/_snaps/verb-fill.md
hadley: Yeah, no changes at this level is going to make this any pretty. But it seems like we could reduce by using an `ELSE`? e.g. `SUM(CASE WHEN `n1` IS NULL THEN 0 ELSE 1 END)` ?

hadley: Or should we also indent the clauses of a window statement?

```sql
SUM(CASE WHEN (((`n1`) IS NULL)) THEN (0) WHEN NOT(((`n1`) IS NULL)) THEN (1) END) 
  OVER (ORDER BY -`id` ROWS UNBOUNDED PRECEDING) 
  AS `..dbplyr_partion_1`
```

hadley (commented):
Looks good! A few comments/questions below.

R/db-sql.R
mgirlich: Yes, there should be no problem. The `lvl` argument is just for indenting and the output is the same as before apart from indent.

R/sql-clause.R
mgirlich: It feels quite unlikely that we need `con` but with SQL you never know... So, we might as well just keep it.

R/backend-access.R
mgirlich: `con` is not needed anymore (see discussion below) so we could remove it but we can also keep it just to be safe.

tests/testthat/_snaps/verb-expand.md
mgirlich: `sql_query_join.DBIConnection()` still uses `sql_clause("ON", on, sep = " AND", parens = TRUE)` which doesn't add any parentheses.
But `sql_query_semi_join.DBIConnection()` uses `sql_clause_where(on, lvl = 1)` which adds the parentheses.
Would you prefer to also have parens here?

tests/testthat/_snaps/verb-fill.md
mgirlich: That's the translation of `ifelse()`

``` r
translate_sql(ifelse(is.na(x), 0L, 1L))
#> <SQL> CASE WHEN (((`x`) IS NULL)) THEN (0) WHEN NOT(((`x`) IS NULL)) THEN (1) END
```

<sup>Created on 2021-12-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I think this is because `ELSE` also catches `NULL` (which of course does not matter in this case).

The alternative with `case_when()` is a bit shorter but produces too many lines

``` r
translate_sql(case_when(is.na(x) ~ 0L, TRUE ~ 1L))
#> <SQL> CASE
#> WHEN (((`x`) IS NULL)) THEN (0)
#> ELSE (1)
#> END
```

<sup>Created on 2021-12-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mgirlich: Indenting the window might make sense. I'll have a look at it.

R/db-sql.R
mgirlich: 👍  Renamed to `sql_indent_subquery()` and exported.

tests/testthat/_snaps/backend-sqlite.md
mgirlich: I indented `ON` because this was a bit easier

tests/testthat/_snaps/verb-fill.md
mgirlich: I created a new PR for a nicer translation of the conditionals #738

mgirlich: The window clauses currently cannot be indented. They are translated in `sql_build.op_select()` but the level is only known later in the `sql_render()` stage. It might be possible to tackle this with the lazy render pipeline but I'm not sure it's worth it.

tests/testthat/_snaps/verb-expand.md
hadley: No strong preference, but it feels better to be consistent?

R/sql-clause.R
hadley: You can make the call — it'll be up to you to add it back in anyway if we later discover it is necessary 😉

tests/testthat/_snaps/verb-fill.md
hadley: Ok, makes sense. Not a big deal.

hadley (approved):
Feel free to merge whenever you think it's done. Just don't forget to add a news bullet, and squash merge.

R/sql-clause.R
mgirlich: Then I hope it won't be necessary 😉

PR #738, opened 2021-12-20
Nicer conditional SQL

mgirlich:
This slightly improves the SQL translations of `is.na()`, `if()`, `ifelse()`, `if_else()` and `case_when()`

Before

```r
translate_sql(ifelse(is.na(x), 0L, 1L))
#> <SQL> CASE WHEN (((`x`) IS NULL)) THEN (0) WHEN NOT(((`x`) IS NULL)) THEN (1) END

translate_sql(case_when(is.na(x) ~ 0L, TRUE ~ 1L))
#> <SQL> CASE
#> WHEN (((`x`) IS NULL)) THEN (0)
#> ELSE (1)
#> END
```

After

```r
translate_sql(ifelse(is.na(x), 0L, 1L))
#> <SQL> CASE WHEN ((`x` IS NULL)) THEN 0 WHEN NOT ((`x` IS NULL)) THEN 1 END

translate_sql(case_when(is.na(x) ~ 0L, TRUE ~ 1L))
#> <SQL> CASE WHEN ((`x` IS NULL)) THEN 0 ELSE 1 END
```


R/backend-.R
```diff
@@ -150,9 +150,15 @@ base_scalar <- sql_translator(
     sql_expr(ROUND(!!x, !!as.integer(digits)))
   },
 
-  `if` = sql_if,
-  if_else = function(condition, true, false, missing = NULL) sql_if(condition, true, false, missing),
-  ifelse = function(test, yes, no) sql_if(test, yes, no),
+  `if` = function(cond, if_true, if_false = NULL) sql_if(
```
hadley, line 153: Since these now span multiple lines, I'd add `{}`

R/translate-sql-conditional.R
```diff
@@ -66,5 +81,18 @@ sql_switch <- function(x, ...) {
 }
 
 sql_is_null <- function(x) {
-  sql_expr((((!!x)) %is% NULL))
+  x_expr <- enexpr(x)
+  if (is_call(x_expr)) {
```
hadley, line 85: Can you now use `enpar` here?

R/translate-sql-conditional.R
```diff
@@ -1,14 +1,23 @@
...
-      build_sql(" WHEN NOT(", cond, ") THEN (", if_false, ")"),
-    if (!is.null(missing)) {
-      missing_cond <- translate_sql(is.na(!!cond), con = sql_current_con())
-      build_sql(" WHEN ", missing_cond, " THEN (", missing, ")")
-    },
-    " END"
-  )
+  out <- build_sql("CASE WHEN ", enpar(cond), " THEN ", enpar(if_true))
+
+  if (!is.null(if_false) && identical(if_false, missing)) {
```
hadley, line 4: I think this deserves a comment to briefly describe the reasoning.

tests/testthat/test-translate-sql.R
```diff
@@ -26,11 +26,11 @@ test_that("between translated to special form (#503)", {
 
 test_that("is.na and is.null are equivalent", {
   # Needs to be wrapped in parens to ensure correct precedence
-  expect_equal(translate_sql(is.na(x)), sql("((`x`) IS NULL)"))
-  expect_equal(translate_sql(is.null(x)), sql("((`x`) IS NULL)"))
+  expect_equal(translate_sql(is.na(x)), sql("(`x` IS NULL)"))
```
hadley, line 29: I think this probably needs a new test for `translate_sql(is.na(x + y))` to capture the spirit of the original test.

R/translate-sql-conditional.R
hadley: And I don't think I saw a test that covered this branch?

hadley (approved):

R/translate-sql-conditional.R
mgirlich: Yes, and thanks to that I found an issue with this approach and the pipe

mgirlich:
I had to change from `enexpr()` to `enquo()` due to an issue with the pipe: `translate_sql(x %>% ifelse(1, 2))` didn't work but it does now.

hadley (approved):

PR #741, opened 2021-12-26
fix typo, dplyr -> dbplyr

eitsupi:
Judging from the context, these seem to be typos.


mgirlich:
Thanks for catching that!

PR #744, opened 2022-01-05
Update `pivot_longer()` args taking single/multiple values

mgirlich:
This PR prepares `pivot_longer.tbl_lazy()` for the planned changes in `tidyr` (https://github.com/tidyverse/tidyr/issues/1296).

@DavisVaughan Feel free to have a look at the PR.


DESCRIPTION
```diff
@@ -57,7 +57,7 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 Language: en-gb
 Roxygen: list(markdown = TRUE)
```
DavisVaughan, line 59: Might want to go ahead and put a Remote on dev tidyr.

Also, while it probably won't make much of a difference, it might be best to change from Suggesting just `tidyr` to `tidyr (>= <the-current-dev-version>)`.

When tidyr is on CRAN we can remove the Remote and bump to the new CRAN version, since that will be required for this all to work right

tests/testthat/test-verb-pivot-longer.R
```diff
@@ -84,6 +84,31 @@ test_that("can override default output column type", {
   )
 })
 
+test_that("`values_transform` works with single functions (#1284)", {
```
DavisVaughan, line 87: Miiiight be worth similar tests for `names_ptypes` and `names_transform`. Those are obviously a little different because they are handled by `tidyr::build_longer_spec()` directly, but still may be good to validate that it is working right.

DavisVaughan (approved):
Looks good, thanks! I imagine that this will just sit until the next version of tidyr is released.

DESCRIPTION
mgirlich: 👍

tests/testthat/test-verb-pivot-longer.R
mgirlich: I added one test to see that they are passed through correctly.

PR #746, opened 2022-01-05
`pivot_*()` works with formula functions

### LINKED ISSUE #745: Support formula in `values_fn` and `values_transform`
``` r
library(dbplyr)

lazy_frame(x = 1) %>%
  tidyr::pivot_longer(x, values_transform = list(value = as.character))
#> <SQL>
#> SELECT 'x' AS `name`, CAST(`x` AS TEXT) AS `value`
#> FROM `df`

lazy_frame(x = 1) %>%
  tidyr::pivot_longer(x, values_transform = list(value = ~ as.character(.x)))
#> Error: Result 1 must be a single string, not NULL of length 0


spec1 <- tibble::tibble(.name = "x", .value = "val", key = "x")
lazy_frame(a = c(1, 1, 2), key = c("x", "x", "x"), val = c(1, 10, 100)) %>% 
  dbplyr:::dbplyr_pivot_wider_spec(spec1, values_fn = sum)
#> <SQL>
#> SELECT `a`, SUM(CASE WHEN (`key` = 'x') THEN (`val`) END) AS `x`
#> FROM `df`
#> GROUP BY `a`

lazy_frame(a = c(1, 1, 2), key = c("x", "x", "x"), val = c(1, 10, 100)) %>% 
  dbplyr:::dbplyr_pivot_wider_spec(spec1, values_fn = ~ sum(.x, na.rm = TRUE))
#> Error: Can't convert NULL to a string
```

<sup>Created on 2022-01-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mgirlich:
Fixes #745.

Previously, the `na.rm` argument was just added in `dbplyr_pivot_wider_spec()` so that `values_fn = max` did not produce a warning. After this PR it does produce the warning `Missing values are always removed in SQL.` which is slightly annoying.

@hadley What do you think? Should this try to add the `na.rm = TRUE` argument again?


hadley:
I think this is sufficient; I think it's probably better to be consistent with other dbplyr functions even if it create a little extra function for the user. (And the warning isn't that bad since you only see it once a session.)

PR #748, opened 2022-01-11
Add `rows_*()` verbs

### LINKED ISSUE #736: Add `rows_*.tbl_dbi()` methods
These methods currently live in the `dm` package but it would make more sense to have them in `dbplyr`.

mgirlich:
Closes #736.

`returning`: This argument is still experimental and not (yet?) in the generic. ~To support `tidyselect` and not trigger a warning by `rlang::check_dots_used()` (in the generic) the user has to quote it manually (see https://github.com/tidyverse/dplyr/issues/5985).~ (Edit: not necessary anymore after `check_dots_used()` was removed in the generic).
Theoretically, one could execute arbitrary SQL in the `RETURNING` clause but I'm not sure whether there are sufficient use cases to support this.

`nonexistent`: The current implementation is `nonexistent = "ignore"`. To support `nonexistent = "error"` one could use a foreign key constraint on `y`. I'll tackle this in another PR. (Update: renamed to `unmatched` and "error" is not supported for now)

`rows_upsert()`: This is the most tricky and needs different methods:
* Some databases (e.g. Oracle or MS SQL) support `MERGE INTO ... WHEN (NOT) MATCHED`
* Some databases (e.g. SQLite, Postgres) have an upsert if there is a unique constraint, basically `INSERT ... ON CONFLICT DO UPDATE/NOTHING`
* If there is no unique constraint this can still be done in one query with a CTE (e.g. Postgres) or two queries (e.g. SQLite)

It might make sense to add an argument so that the user can choose whether a unique constraint exists or not. See
https://github.com/cynkra/dm/pull/616#issuecomment-920613435 for some more information on the Pros/Cons. I will do this in another PR.


R/rows.R
```diff
@@ -0,0 +1,610 @@
...
+
+  rows_check_key(x, by, "x")
+  rows_check_key(y, by, "y", unique = TRUE)
+
+  new_columns <- setdiff(colnames(y), by)
+
+  returning_cols <- rows_check_returning(x, returning, enexpr(returning))
+
+  if (!is_null(name)) {
+    if (conflict == "error") {
```
mgirlich, line 41: I think it should be possible to mute this message. But how? Add another argument `quiet`? And should this inform how to add an index via `db_create_index()`?

mgirlich: It would also be possible to add an option `check` that collects the data and then checks this but I'm not really sure this is worth it as the unique index makes much more sense.

mgirlich:
@hadley It would be great to merge most of the remaining PRs and then make a CRAN release. Do you think we can manage that in the next couple of weeks?

hadley (commented):
Is there anything in particular you'd like me to look at? I'm unlikely to have time to review it depth, but I trust that you've implemented it correctly (and we'll discover and fix any mistakes as people start to use it).

krlmlr:
I'll take a closer look here and double-check if it's consistent/compatible with what's in dm. It might be worth to release a dm version that uses the dbplyr implementation if it's available before releasing dbplyr.

R/rows.R
krlmlr: `conflict = "allow"` or `conflict = "silent"` ?

R/rows.R
```diff
@@ -0,0 +1,610 @@
...
+  #
+  #   message <- c(
+  #     glue("`{arg}` key values must be unique."),
+  #     i = glue("The following rows contain duplicate key values: {duplicated}.")
+  #   )
+  #
+  #   abort(message, call = error_call)
+  # }
+  #
+  # out
```
krlmlr, line 491: I don't think we should be doing anything here. We have tools in dm to check referential integrity and key constraints "after the fact", and we can simulate with `in_place = FALSE` . Actively checking here can be very expensive.

_pkgdown.yml
```diff
@@ -32,6 +32,8 @@ reference:
   - intersect.tbl_lazy
   - mutate.tbl_lazy
   - pull.tbl_sql
+  - rows_update.tbl_lazy
```
krlmlr, line 35: There is one help page for all "rows" functions, right?

```suggestion
  - "rows-db"
```

R/db-sql.R
```diff
@@ -112,6 +114,31 @@ sql_translate_env.DBIConnection <- function(con) {
...
+sql_named_cols <- function(con, cols, table = NULL) {
+  nms <- names2(cols)
+  nms[nms == cols] <- ""
+
+  cols <- sql_table_prefix(con, cols, table)
+  cols <- set_names(cols, nms)
+  escape(ident_q(cols), collapse = NULL, con = con)
+}
+
 sql_random <- function(con) {
```
krlmlr, line 142: Looks like this was accidentally moved up to `sql_returning_cols()`:

```suggestion
#' @export
sql_random <- function(con) {
```

krlmlr (commented):
Thanks, looks great -- using the internal dbplyr tools surely gives much better code than using `paste()` in dm. This is the first review round, looking only at code. Now installing the branch and checking compatibility with dm.

R/rows.R
krlmlr: I've proposed to change this to `conflict = "drop"` in the dplyr PR, and have `conflict = "ignore"` allow duplicate row. Hope we can reach a consensus.

R/rows.R
```diff
@@ -0,0 +1,610 @@
...
+      by = by,
+      ...,
+      conflict = conflict,
+      returning_cols = returning_cols
+    )
+
+    rows_get_or_execute(x, sql, returning_cols)
+  } else {
+    if (conflict == "error") {
+      abort('`conflict = "error"` is not supported for `in_place = FALSE`.')
```
krlmlr, line 58: We still could run a query like `union_all() %>% count() %>% filter(n > 1) %>% head(1) %>% collect()` and error if we see at least one row here, along with a message that this is expensive and users can use `conflict = "ignore"` (meaning that conflicts are allowed). The alternative `conflict = "drop"` can also be expensive on the database.

krlmlr:
Are you planning to support the `conflict`/`nonexistent` arguments for all verbs? How about the following simplified approach:

- We include the argument with the same defaults as in dplyr
- Any other value than the "natural" thing on the database (i.e. `conflict = "ignore"` meaning "allow duplicate rows", or `nonexistent = "ignore"`) gives a warning once every 8 hours (via `rlang::warn(.frequency = "regularly")`)
- Exception: `rows_insert(conflict = "drop")` already has code, we can keep using that

This behavior is consistent with `sum(na.rm = TRUE)` which also doesn't work on databases. I also don't mind an error instead of a warning, as long as there's a way to support what's natural on the database. Later we could work on perhaps supporting other options for `conflict` and `nonexistent` .

Of course this means that dplyr must support what's natural on the database.

krlmlr:
Otherwise, compatibility with dm looks great: https://github.com/cynkra/dm/pull/912 succeeds locally, now testing on databases.

R/lazy-ops.R
```diff
@@ -29,8 +29,10 @@ lazy_query_base <- function(x, vars, class = character()) {
   )
 }
 
-lazy_query_local <- function(df) {
-  lazy_query_base(df, names(df), class = "local")
+lazy_query_local <- function(df, name) {
+  out <- lazy_query_base(df, names(df), class = "local")
+  out$name <- name
```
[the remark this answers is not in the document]
mgirlich: Hmm, I was pretty sure I already added `...` in `lazy_query_base()` at some point 🤷 But yeah, makes sense

R/db-sql.R
```diff
@@ -399,6 +426,126 @@ sql_set_op.DBIConnection <- function(con, x, y, method) {
...
+#' @export
+#' @rdname db-sql
+sql_query_upsert <- function(con, x_name, y, by, update_cols, ...,
+                                  returning_cols = NULL) {
+  rlang::check_dots_used()
+  UseMethod("sql_query_upsert")
+}
+
+#' @export
+sql_query_upsert.DBIConnection <- function(con, x_name, y, by,
```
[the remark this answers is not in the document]
mgirlich: I think it should work in many databases but I also think most offer better or easier proprietary alternatives (`MERGE`, `INSERT INTO ... ON CONFLICT`). Even in 2022 upserts are a bit of a mess in SQL...

R/rows.R
mgirlich: I think it is worth supporting `conflict = "error"` but `collecting()` by default does not sound good to me. So, maybe `conflict` should not be an optional argument for `rows_insert.tbl_lazy(in_place = FALSE)`.

R/rows.R
mgirlich: What do you mean by "after the fact"?

There should definitely be tools to easiliy check the uniqueness of `y` and they should be well-documented.

If `y` is a local table the check is easy, so it would be an option to check this before copying.

R/rows.R
krlmlr: We pay extra for the checks in dplyr too, and we collect at most one row. We could give a message if the `conflict` argument is missing, to ensure that users are aware?

krlmlr:
Unfortunately, double quoting is an issue with dm. We use `ident_q` to mark identifiers as quoted, this requires using `escape()` instead of `sql_escape_ident()` -- also in existing code. Figuring out the best way to handle this.

krlmlr:
Do we want a message if `in_place` is left unspecified?

krlmlr:
There are several ways to create a writable `tbl` object to a remote table: by name, via `in_schema()`, `ident()`, or `ident_q()` . All these work with `rows_insert()` and SQLite, not so much with SQL Server (and Postgres, judging by the test failures in dm). I'll take a closer look.

## SQLite

``` r
library(dbplyr)
library(tidyverse)

con <- src_memdb()$con

con_frame <- function(...) {
  copy_inline(con, tibble(...))
}

DBI::dbExecute(con, "CREATE TABLE test (a INTEGER)")
#> [1] 0

tbl_pure <- tbl(con, "test")
tbl_pure
#> # Source:   table<test> [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#> # … with 1 variable: a <int>
rows_insert(tbl_pure, memdb_frame(a = 1L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
rows_insert(tbl_pure, memdb_frame(a = 1L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"

tbl_in_schema <- tbl(con, in_schema("main", "test"))
tbl_in_schema
#> # Source:   table<`main`.`test`> [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
rows_insert(tbl_in_schema, memdb_frame(a = 2L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
#> 2     2
rows_insert(tbl_in_schema, memdb_frame(a = 2L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"

tbl_ident <- tbl(con, ident("test"))
tbl_ident
#> # Source:   table<test> [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
#> 2     2
rows_insert(tbl_ident, memdb_frame(a = 3L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
#> 2     2
#> 3     3
rows_insert(tbl_ident, memdb_frame(a = 3L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"

tbl_ident_q <- tbl(con, ident_q("`test`"))
tbl_ident_q
#> # Source:   table<`test`> [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
#> 2     2
#> 3     3
rows_insert(tbl_ident_q, memdb_frame(a = 4L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
#> 2     2
#> 3     3
#> 4     4
rows_insert(tbl_ident_q, memdb_frame(a = 4L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"
```

<sup>Created on 2022-04-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

## SQL Server (with #823)

``` r
library(dbplyr)
library(tidyverse)

# FIXME: Specify server connection:
con <- sql_server_con()

con_frame <- function(...) {
  copy_inline(con, tibble(...))
}
con_frame(a = 1)
#> # Source:   [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <dbl>
#> 1     1

DBI::dbExecute(con, "CREATE TABLE test (a INTEGER)")
DBI::dbExecute(con, "DELETE FROM test")
#> [1] 0

tbl_pure <- tbl(con, "test")
tbl_pure
#> # Source:   table<test> [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#> # … with 1 variable: a <int>
rows_insert(tbl_pure, con_frame(a = 1L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198[@SV0376/test]
#>       a
#>   <int>
#> 1     1
rows_insert(tbl_pure, con_frame(a = 1L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"

tbl_in_schema <- tbl(con, in_schema("dbo", "test"))
tbl_in_schema
#> # Source:   table<"dbo"."test"> [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
rows_insert(tbl_in_schema, con_frame(a = 2L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
#> 2     2
rows_insert(tbl_in_schema, con_frame(a = 2L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"
#> Error in new_result(connection@ptr, statement, immediate): nanodbc/nanodbc.cpp:1594: 00000: [FreeTDS][SQL Server]The multi-part identifier ""dbo"."test".a" could not be bound.

tbl_ident <- tbl(con, ident("test"))
tbl_ident
#> # Source:   table<test> [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
rows_insert(tbl_ident, con_frame(a = 3L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
#> 2     3
rows_insert(tbl_ident, con_frame(a = 3L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"

tbl_ident_q <- tbl(con, ident_q("[test]"))
tbl_ident_q
#> # Source:   table<[test]> [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
#> 2     3
rows_insert(tbl_ident_q, con_frame(a = 4L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
#> 2     3
#> 3     4
rows_insert(tbl_ident_q, con_frame(a = 4L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"
#> Error in new_result(connection@ptr, statement, immediate): nanodbc/nanodbc.cpp:1594: 00000: [FreeTDS][SQL Server]The multi-part identifier "[test].a" could not be bound.
```

<sup>Created on 2022-04-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

krlmlr:
There are several ways to create a writable `tbl` object to a remote table: by name, via `in_schema()`, `ident()`, or `ident_q()` . All these work with `rows_insert()` and SQLite, not so much with SQL Server (and Postgres, judging by the test failures in dm). I'll take a closer look.

## SQLite

``` r
library(dbplyr)
library(tidyverse)

con <- src_memdb()$con

con_frame <- function(...) {
  copy_inline(con, tibble(...))
}

DBI::dbExecute(con, "CREATE TABLE test (a INTEGER)")
#> [1] 0

tbl_pure <- tbl(con, "test")
tbl_pure
#> # Source:   table<test> [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#> # … with 1 variable: a <int>
rows_insert(tbl_pure, memdb_frame(a = 1L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
rows_insert(tbl_pure, memdb_frame(a = 1L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"

tbl_in_schema <- tbl(con, in_schema("main", "test"))
tbl_in_schema
#> # Source:   table<`main`.`test`> [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
rows_insert(tbl_in_schema, memdb_frame(a = 2L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
#> 2     2
rows_insert(tbl_in_schema, memdb_frame(a = 2L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"

tbl_ident <- tbl(con, ident("test"))
tbl_ident
#> # Source:   table<test> [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
#> 2     2
rows_insert(tbl_ident, memdb_frame(a = 3L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
#> 2     2
#> 3     3
rows_insert(tbl_ident, memdb_frame(a = 3L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"

tbl_ident_q <- tbl(con, ident_q("`test`"))
tbl_ident_q
#> # Source:   table<`test`> [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
#> 2     2
#> 3     3
rows_insert(tbl_ident_q, memdb_frame(a = 4L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: sqlite 3.38.0 [:memory:]
#>       a
#>   <int>
#> 1     1
#> 2     2
#> 3     3
#> 4     4
rows_insert(tbl_ident_q, memdb_frame(a = 4L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"
```

<sup>Created on 2022-04-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

## SQL Server (with #823)

``` r
library(dbplyr)
library(tidyverse)

# FIXME: Specify server connection:
con <- sql_server_con()

con_frame <- function(...) {
  copy_inline(con, tibble(...))
}
con_frame(a = 1)
#> # Source:   [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <dbl>
#> 1     1

DBI::dbExecute(con, "CREATE TABLE test (a INTEGER)")
DBI::dbExecute(con, "DELETE FROM test")
#> [1] 0

tbl_pure <- tbl(con, "test")
tbl_pure
#> # Source:   table<test> [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#> # … with 1 variable: a <int>
rows_insert(tbl_pure, con_frame(a = 1L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
rows_insert(tbl_pure, con_frame(a = 1L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"

tbl_in_schema <- tbl(con, in_schema("dbo", "test"))
tbl_in_schema
#> # Source:   table<"dbo"."test"> [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
rows_insert(tbl_in_schema, con_frame(a = 2L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
#> 2     2
rows_insert(tbl_in_schema, con_frame(a = 2L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"
#> Error in new_result(connection@ptr, statement, immediate): nanodbc/nanodbc.cpp:1594: 00000: [FreeTDS][SQL Server]The multi-part identifier ""dbo"."test".a" could not be bound.

tbl_ident <- tbl(con, ident("test"))
tbl_ident
#> # Source:   table<test> [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
rows_insert(tbl_ident, con_frame(a = 3L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
#> 2     3
rows_insert(tbl_ident, con_frame(a = 3L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"

tbl_ident_q <- tbl(con, ident_q("[test]"))
tbl_ident_q
#> # Source:   table<[test]> [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
#> 2     3
rows_insert(tbl_ident_q, con_frame(a = 4L), conflict = "ignore")
#> Matching, by = "a"
#> # Source:   [?? x 1]
#> # Database: Microsoft SQL Server 15.00.4198
#>       a
#>   <int>
#> 1     1
#> 2     3
#> 3     4
rows_insert(tbl_ident_q, con_frame(a = 4L), conflict = "ignore", in_place = TRUE)
#> Matching, by = "a"
#> Error in new_result(connection@ptr, statement, immediate): nanodbc/nanodbc.cpp:1594: 00000: [FreeTDS][SQL Server]The multi-part identifier "[test].a" could not be bound.
```

<sup>Created on 2022-04-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

krlmlr:
`sql_output_cols()` has been dropped in 71d035ddd8dfa22af9b5bc5af01cfdfe49d70217. We introduced it in dm because the `OUTPUT` clause in SQL Server is inserted in a different place from the `RETURNING` clause for other databases. It seemed like a smaller price to pay compared to reimplementing all SQL generators separately for SQL Server. What am I missing?

krlmlr:
How do you feel about failing unconditionally with `rows_insert(conflict = "error", in_place = TRUE)`, provided that we have an escape hatch that inserts unconditionally and lets the backend decide (e.g. `conflict = "backend"` or a dedicated `rows_append()`, see https://github.com/DavisVaughan/dplyr/pull/1)?

mgirlich:
> `sql_output_cols()` has been dropped in [71d035d](https://github.com/tidyverse/dbplyr/commit/71d035ddd8dfa22af9b5bc5af01cfdfe49d70217). We introduced it in dm because the `OUTPUT` clause in SQL Server is inserted in a different place from the `RETURNING` clause for other databases. It seemed like a smaller price to pay compared to reimplementing all SQL generators separately for SQL Server. What am I missing?
> 
> `rows_insert()` no longer works with the `returning` argument on SQL Server.

Hmm, I'm not entirely sure on this one. With `sql_output_cols()` we don't need a separate implementation for two relatively simple functions `sql_query_insert()` and `sql_query_delete()` (but we still need `sql_query_update()` and `sql_query_upsert()`). On the other hand I find it a bit confusing to have two generics for one kind of functionality (`sql_returning_cols()` and `sql_output_cols()`) and to have them both in one query.

mgirlich:
> How do you feel about failing unconditionally with `rows_insert(conflict = "error", in_place = TRUE)`, provided that we have an escape hatch that inserts unconditionally and lets the backend decide (e.g. `conflict = "backend"` or a dedicated `rows_append()`, see [DavisVaughan/dplyr#1](https://github.com/DavisVaughan/dplyr/pull/1))?

This seems to be a good solution. Of course this implies that non-unique local tables might error quite late but the user can still do the check himself. Is there any easy to use tool for this? I know there is `dm::dm_add_pk()` but this requires to first create a dm object which seems a bit too complicated for such a simple use case.

mgirlich:
The issue for Postgres should be fixed. I didn't expect `ident_q(set_names(cols, nms))` to be different from `set_names(ident_q(cols), nms)` but it is. Another example for issues with `ident()` and friends due to them escaping their input...

krlmlr:
I fixed tests after #819, and added four more commits. Please feel free to revert/edit as necessary. Tests pass locally for all commits individually.

krlmlr:
For `rows_append()` we'll need https://github.com/DavisVaughan/dplyr/pull/1 and https://github.com/tidyverse/dplyr/pull/6203, let's wait until they land in the main branch or even on CRAN.

mgirlich:
Todo
* [ ] `rows_append()` -> wait for dplyr
* [ ] add `rows_insert_returning()` and friends and remove `returning` argument
* [ ] `rows_insert(conflict = “error”)` -> Raise an error unconditionally
* [ ] `rows_update/delete/patch(nonexistent = “error”)` -> Raise an error unconditionally
* [ ] Raise an error unconditionally – add a `check_unique` option to dbplyr only, handled by `check_dots_empty()` in dplyr?

krlmlr:
Are we keeping the `returning` argument, or do we introduce new `rows_*_returning()` functions?

I saw that the default for `in_place` has changed to `FALSE`, I guess this means we no longer give a message if `in_place` is omitted?

Regarding integration with dm, I think we can release dbplyr first and then adapt dm -- the dbplyr update doesn't break dm's tests, because the latter still overwrites the methods. Should dbplyr give a warning mentioning that the `rows_*()` API behaves differently when dm <= 0.2.8 is loaded?

It looks like the SQL Server implementation needs more work, I'll propose something today. Postgres looks fine.

krlmlr:
MSSQL seems good now.

mgirlich:
> Are we keeping the `returning` argument, or do we introduce new `rows_*_returning()` functions?

I'm okay with both. Do you have a preference?

> I saw that the default for `in_place` has changed to `FALSE`, I guess this means we no longer give a message if `in_place` is omitted?

Yes, I don't see a need to give a message.
 
> Regarding integration with dm, I think we can release dbplyr first and then adapt dm -- the dbplyr update doesn't break dm's tests, because the latter still overwrites the methods. Should dbplyr give a warning mentioning that the `rows_*()` API behaves differently when dm <= 0.2.8 is loaded?

Feel free to add a message though it will probably affect only few users.

> MSSQL seems good now.

Thanks!

krlmlr (approved):
Thanks! Looks good to me, with https://github.com/cynkra/dm/pull/988 we get the new behavior also for the legacy code in dm. I can release dm before or shortly after dbplyr, doesn't really matter.

krlmlr:
So, let's keep the `returning` argument and avoid adding new messages.

PR #751, opened 2022-01-13
Add `ON COMMIT PRESERVE ROWS` when creating temporary table in Oracle

### LINKED ISSUE #750: Missing `ON COMMIT PRESERVE ROWS` in Oracle `CREATE GLOBAL TEMPORARY TABLE`
I believe the bug is in https://github.com/tidyverse/dbplyr/blob/main/R/backend-oracle.R line 120:

image

Should look more like this https://github.com/r-dbi/odbc/blob/74ad62c275b1c95a70814707901bdafd8fc2d2b6/R/db.R#L32 line 15:

image

The effect of this is that anything with `temporary = TRUE` (e.g. `copy_to` or `compute` will result in an empty temporary table. Using `{odbc}` to connect.

admivsn:
Closes #750 

Tested this out locally and fixes the issue - without this additional statement the temporary table is blank.

https://stackoverflow.com/questions/27093937/how-to-create-a-temporary-table-as-copy-of-another-table-in-oracle

### DIFF
--- a/R/backend-oracle.R
+++ b/R/backend-oracle.R
@@ -120,7 +120,7 @@ sql_query_wrap.Oracle <- function(con, from, name = unique_subquery_name(), ...,
 sql_query_save.Oracle <- function(con, sql, name, temporary = TRUE, ...) {
   build_sql(
     "CREATE ", if (temporary) sql("GLOBAL TEMPORARY "), "TABLE \n",
-    as.sql(name, con), " ON COMMIT PRESERVE ROWS AS\n", sql,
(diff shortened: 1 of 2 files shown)


hadley:
Isn't this comment a bit concerning?

> Generally, I think that a policy of CREATE GLOBAL TEMPORARY TABLE using SELECT * FROM indicates a misunderstanding of the construct. GTTs in Oracle are permanent data structures; only the records are temporary.

Like won't we be leaving a bunch of empty tables around?

PR #753, opened 2022-01-14
Added missing connection parameter to the dbplyr_fill0 implementation

rsund:
The connection parameter was not passed to the first translate_sql() call in the dbplyr_fill0 implementation for databases without support for IGNORE NULLS. Because of that the quotes other than \`  \` were not correct for that particular translate_sql() call.

Following the example in the [fill documentation](https://dbplyr.tidyverse.org/reference/fill.tbl_lazy.html) the produced SQL was as follows with a database that uses " " as quotes:

```sql
<SQL>
SELECT "group", "name", "role", MAX("n_squirrels") OVER (PARTITION BY "..dbplyr_partion_1") AS "n_squirrels", 
MAX("n_squirrels2") OVER (PARTITION BY "..dbplyr_partion_2") AS "n_squirrels2", "id" FROM
(SELECT "group", "name", "role", "n_squirrels", "n_squirrels2", "id", SUM(CASE WHEN (((`n_squirrels`) IS NULL)) THEN 0 
ELSE 1 END) OVER (ORDER BY "id" ROWS UNBOUNDED PRECEDING) AS "..dbplyr_partion_1", 
SUM(CASE WHEN (((`n_squirrels2`) IS NULL))  THEN 0 ELSE 1 END) OVER (ORDER BY "id" ROWS UNBOUNDED PRECEDING) AS
"..dbplyr_partion_2" FROM "squirrels") "q01"
```

So following the `SUM(CASE WHEN...` there are wrong quotes in two places (around the variable names).

The fix was simple and only requires that the missing connection parameter is added to the translate_sql() call as shown in the commit.

### DIFF
--- a/R/verb-fill.R
+++ b/R/verb-fill.R
@@ -141,6 +141,7 @@ dbplyr_fill0.SQLiteConnection <- function(.con,
     cols_to_fill,
     ~ translate_sql(
       cumsum(case_when(is.na(!!.x) ~ 0L, TRUE ~ 1L)),
+      con = .con,
       vars_order = translate_sql(!!!order_by_cols, con = .con),
       vars_group = op_grps(.data),
     )


rsund:
Interestingly the tests failed, but while running the same example with RSQlite the produced SQL is fine:

```r
squirrels <- tibble::tribble(
  ~group,    ~name,     ~role,     ~n_squirrels, ~ n_squirrels2,
  1,      "Sam",    "Observer",   NA,                 1,
  1,     "Mara", "Scorekeeper",    8,                NA,
  1,    "Jesse",    "Observer",   NA,                NA,
  1,      "Tom",    "Observer",   NA,                 4,
  2,     "Mike",    "Observer",   NA,                NA,
  2,  "Rachael",    "Observer",   NA,                 6,
  2,  "Sydekea", "Scorekeeper",   14,                NA,
  2, "Gabriela",    "Observer",   NA,                NA,
  3,  "Derrick",    "Observer",   NA,                NA,
  3,     "Kara", "Scorekeeper",    9,                 10,
  3,    "Emily",    "Observer",   NA,                NA,
  3, "Danielle",    "Observer",   NA,                NA
)
squirrels$id <- 1:12

con <- DBI::dbConnect(RSQLite::SQLite())

DBI::dbWriteTable(con, "squirrels", squirrels, overwrite=TRUE)

library(dplyr)
tbl(con, "squirrels") %>%
    dbplyr::window_order(id) %>%
    tidyr::fill(
      n_squirrels,
      n_squirrels2,
    ) %>%
    show_query()

```
```sql
<SQL>
SELECT
  `group`,
  `name`,
  `role`,
  MAX(`n_squirrels`) OVER (PARTITION BY `..dbplyr_partion_1`) AS `n_squirrels`,
  MAX(`n_squirrels2`) OVER (PARTITION BY `..dbplyr_partion_2`) AS `n_squirrels2`,
  `id`
FROM (
  SELECT
    `group`,
    `name`,
    `role`,
    `n_squirrels`,
    `n_squirrels2`,
    `id`,
    SUM(CASE WHEN ((`n_squirrels` IS NULL)) THEN 0 ELSE 1 END) OVER (ORDER BY `id` ROWS UNBOUNDED PRECEDING) AS `..dbplyr_partion_1`,
    SUM(CASE WHEN ((`n_squirrels2` IS NULL)) THEN 0 ELSE 1 END) OVER (ORDER BY `id` ROWS UNBOUNDED PRECEDING) AS `..dbplyr_partion_2`
  FROM `squirrels`
)

```

rsund:
Interestingly the tests failed. I should have been testing more carefully before sending a PR, but I couldn't anticipate that such a simple change had such consequences, especially as things were working just fine with the database with " " quotes after the commit.

Any idea what should be changed?

Maybe instead of adding a connection parameter, the current connection could be stored with the function `set_current_con()` in the beginning of the `dbplyr_fill0()` function so that the engine with " " quotes could fetch the connection (in the `translate_sql()` there is a line for fetching the con if parameter is missing:  `con <- con %||% sql_current_con() %||% simulate_dbi()`).

But I still don't quite get that why adding a clearly missing (and important) parameter changes the behavior in `CASE WHEN` translation?!

mgirlich:
Good catch, thanks for the PR. The tests fail due to an incorrect check when translating `case_when()` that only happens for SQLite. I created a PR to fix this. Once this is merged, the tests pass.

It would be great if you could also add a NEWS entry.

rsund:
OK, good to hear. A commit for the NEWS entry added as requested.

PR #766, opened 2022-01-21
Align `partialeval` with `dtplyr`

### LINKED ISSUE #669: dbplyr `across(...)` doesn't work with `.cols = character()`
`across()` fails on an empty character vector on `tbl_sql` objects, rather than doing nothing

```r
library(dplyr)

memdb_frame(x = 1:10) %>% 
  mutate(
    across(.cols = character(), .fns = as.character)
  )
```

Error: Can't rename variables in this context.

On `tibble` works as expected: 

```r
tibble(x = 1:10) %>% 
  mutate(
    across(.cols = character(), .fns = as.character)
  )
```

```

### LINKED ISSUE #760: Test `across()` better
Adapt big test set from `dplyr`

* `across()` should not compute on grouping columns (though this might change in `dplyr`: https://github.com/tidyverse/dplyr/issues/6127)
``` r
memdb_frame(g = 1, x = 1, y = 1) %>% 
  group_by(g) %>% 
  summarise(across(everything(), ~ 0)) %>% 
  show_query()
#> <SQL>
#> SELECT `g`, 0.0 AS `g`, 0.0 AS `x`, 0.0 AS `y`
#> FROM `dbplyr_001`
#> GROUP BY `g`
```

<sup>Created on 2022-01-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

* `across()` should default to `everything()`
``` r
suppressMessages(devtools::load_all("~/GitHub/dbplyr/"))

tibble(g = 1, x = 1, y = 1) %>% group_by(x) %>% summarise(across())
#> # A tibble: 1 × 3
#>       x     g     y
#>   <dbl> <dbl> <dbl>
#> 1     1     1     1
memdb_frame(g = 1, x = 1, y = 1) %>% group_by(x) %>% summarise(across())
#> # Database: sqlite 3.37.0 [:memory:]
#>       x
#>   <dbl>
#> 1     1
```

<sup>Created on 2022-01-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mgirlich:
Fixes #760 and fixes #669.

This aligns the partial eval file with `dtplyr` so that fixes and improvements can be transferred more easily from one to the other. Most functions are now basically identical. Still different:

* some function names. Should we align names? If so which system do you prefer?
  * `partial_eval()` vs `dt_squash()`
  * `partial_eval_dots()` vs `capture_dots()`
  * `partial_eval_call()` vs `dt_squash_call()`
  * `partial_eval_across()` vs `dt_squash_across()`
* `dbplyr` functions take a `vars` argument but it might make more sense to pass `data` (e.g. because `across()` does not compute on group columns). Changing this would require changing the interface of the exported function `partial_eval()` but might still be worth it.


tests/testthat/test-tidyeval-across.R
```diff
@@ -0,0 +1,331 @@
...
+    summarise(gf, across(1:2, list(mean = mean, sum = sum))) %>% op_vars(),
+    c("x", "y_mean", "y_sum", "z_mean", "z_sum")
+  )
+  expect_equal(
+    summarise(gf, across(1:2, list(mean = mean, sum = sum), .names = "{.fn}_{.col}")) %>% op_vars(),
+    c("x", "mean_y", "sum_y", "mean_z", "sum_z")
+  )
+
+  skip("not yet decided")
+  expect_equal(
```
mgirlich, line 110: Here, `dbplyr` is smarter than `dplyr` and automatically creates names. What do you prefer?

tests/testthat/test-tidyeval-across.R
```diff
@@ -0,0 +1,331 @@
...
+    summarise(gdf, x = mean(x, na.rm = TRUE)) %>% remote_query()
+  )
+})
+
+test_that("across() can use named selections", {
+  df <- lazy_frame(x = 1, y = 2)
+
+  # no fns
+  expect_equal(
+    df %>% summarise(across(c(a = x, b = y))) %>% remote_query(),
```
mgirlich, line 226: I used `expect_equal()` because the queries are very simple. Alternatively, this could use `op_vars()` though this does not fully capture what's going on. What do you think?

R/tidyeval.R
```diff
@@ -55,35 +55,48 @@
 #' x <- 1
 #' partial_eval(quote(x ^ y))
 partial_eval <- function(call, vars = character(), env = caller_env()) {
-  if (is_null(call)) {
-    NULL
-  } else if (is_atomic(call) || blob::is_blob(call)) {
+  # corresponds to `dt_squash()`
+  # TODO change `vars` to `data`
```
hadley, line 59: I think it's worth making this change; I doubt there are many existing uses, and we could provide an informative deprecation based on either the argument name or the type (string vs. data frame or whatever it actually is).

R/tidyeval.R
```diff
@@ -133,32 +154,26 @@ partial_eval_call <- function(call, vars, env) {
...
-  if (is_tidy_pronoun(call)) {
-    if (is_symbol(call[[1]], "$")) {
-      idx <- call[[3]]
-    } else {
-      idx <- as.name(eval_bare(call[[3]], env))
+  if (is_mask_pronoun(call)) {
+    var <- call[[3]]
+    if (is_call(call, "[[")) {
+      var <- sym(eval(var, env))
+      # var <- as.name(eval_bare(var, env))
```
hadley, line 169:
```suggestion
```

tests/testthat/test-tidyeval-across.R
hadley: Probably best to be consistent with dplyr.

tests/testthat/test-tidyeval-across.R
hadley: `expect_equal()` seems fine to me.

hadley (approved):
I like the idea of consistently using `partial_eval_` as prefix.

hadley:
Thanks for working on this!

PR #768, opened 2022-02-02
Update to rlang 1.0.0

mgirlich:
* Update snapshots
* Update documentation
* Use @examplesIf for `tidyr` verbs
* Use `expect_snapshot(error = TRUE)` pattern instead of `expect_snapshot_error()` to show error call
* Improve some error messages from helper functions by passing a call to `abort()`


mgirlich:
Closed in favour of #769.

PR #772, opened 2022-02-03
Increase test coverage

mgirlich:
* Replace `stop()` by `abort()`
* Mark some old or hard to test code as `nocov`
* Try to reach test cover of 100% on most files. Unfortunately, some lines are marked as not covered even though they are. In part this is probably due to [this bug in covr](https://github.com/r-lib/covr/issues/482). But there seem to be some other issues as well.


R/backend-hive.R
```diff
@@ -68,4 +68,9 @@ sql_table_analyze.Hive <- function(con, table, ...) {
   )
 }
 
+#' @export
```
mgirlich, line 71: Moved from `verb-fill.R` to the backend file

R/backend-mysql.R
```diff
@@ -55,7 +55,7 @@ sql_translation.MariaDBConnection <- function(con) {
   sql_variant(
     sql_translator(.parent = base_scalar,
       as.logical = function(x) {
-        sql_expr(IF(x, TRUE, FALSE))
+        sql_expr(IF(!!x, TRUE, FALSE))
```
mgirlich, line 58: One minor bug I found while increasing the test coverage

R/sql.R
```diff
@@ -21,7 +21,7 @@ sql <- function(...) {
 c.sql <- function(..., drop_null = FALSE, con = simulate_dbi()) {
   input <- list(...)
 
-  if (drop_null) input <- purrr::compact(input)
+  if (drop_null) input <- purrr::compact(input) # nocov
```
mgirlich, line 24: The `drop_null` argument is not used anywhere. Could probably be removed...

R/translate-sql-conditional.R
```diff
@@ -73,7 +73,7 @@ sql_case_when <- function(...) {
 sql_switch <- function(x, ...) {
   input <- list2(...)
 
-  named <- names(input) != ""
+  named <- names2(input) != ""
```
mgirlich, line 76: Another minor but for completely unnamed input to `switch()`

R/translate-sql-window.R
```diff
@@ -253,19 +252,15 @@ local_context <- function(x, env = parent.frame()) {
 # Where translation -------------------------------------------------------
 
 uses_window_fun <- function(x, con) {
-  if (is.null(x)) return(FALSE)
```
mgirlich, line 256: `uses_window_fun()` is only called by `sql_build.op_filter()` and there it is always passed quosures.

Issue #598, opened 2021-02-09
pivot_wider doesn't seem to be lazy


Janlow:
Connecting R to an Oracle database using DBI and ODBC. I do a lot of querying using dplyr syntax and avoid using collect() until the very last moment, which I really like and works as it should. However, I used pivot_wider for the first time now inside a long query, without collect() at the end, and it seems as if it actually performs the query, which I have never experienced before, guess that this must be a bug? Do I have to specify somewhere myself that I want it to be executed lazy?

I am including an example so that you can see what I mean. Even though I am saving it as an object it executes, which I only expect if I use collect() or don't save it as an object. Using dbplyr 2.1.0 installed from CRAN.

```r
test <- memdb_frame(
  id = rep(1:5000000, each = 2),
  key = rep(c("x", "y"), 5000000),
  value = 1:10000000
) %>%
  pivot_wider(id_cols = id,
              names_from = key, 
              values_from = value)
```

abalter:
@Janlow -- I'm guessing @hadley  is going to ask you to create a [reprex](https://www.tidyverse.org/help/).

One method is to create a local database, but doubt there is an Oracle version--might have to use RSQLite.

https://dbplyr.tidyverse.org/articles/reprex.html

mgirlich:
Unfortunately, the data needs to be collected for `pivot_wider()`. The reason is that, at least in the standard SQL, the columns from pivotting have to be explicitly specified. It seems that this has slipped the documentation, sorry about that.

In `tidyr` you can first create a specification and then pass it to `pivot_wider_spec()`. Unfortunately, `tidyr::pivot_wider_spec()` is not a generic so we didn't export so we didn't export the internally used equivalent of it. But it might be an option for the future.

100gle:
I met same problem.

Even I can reproduce this problem with official document example...

```R
if (require("tidyr", quietly = TRUE)) {
  memdb_frame(
    id = 1,
    key = c("x", "y"),
    value = 1:2
  ) %>%
    tidyr::pivot_wider(
      id_cols = id,
      names_from = key,
      values_from = value
    )
}
```

the error message like:

```
Error in UseMethod("pivot_wider") : 
  "pivot_wider"没有适用于"c('tbl_SQLiteConnection', 'tbl_dbi', 'tbl_sql', 'tbl_lazy', 'tbl')"目标对象的方法
```

kmishra9:
[...]

PR #794, opened 2022-03-16
Update `pivot_wider()` documentation

mgirlich:
Users are confused that `pivot_wider()` is not lazy (e.g. #598, #703). Collecting is only necessary to generate the spec.
As long as `tidyr::pivot_wider_spec()` is not made a generic I see two ways of handling this:

* just update the documentation (like it is done here) or
* export `dbplyr_pivot_wider_spec()` (which is lazy) so people can generate the spec locally.

@hadley Do you have any preference?

### DIFF
--- a/R/verb-pivot-wider.R
+++ b/R/verb-pivot-wider.R
@@ -1,11 +1,13 @@
 #' Pivot data from long to wide
 #'
+#' @description
 #' `pivot_wider()` "widens" data, increasing the number of columns and
 #' decreasing the number of rows. The inverse transformation is
 #' `pivot_longer()`.
+#' Learn more in `vignette("pivot", "tidyr")`.
+#'
 #' Note that `pivot_wider()` is not and cannot be lazy because we need to look
 #' at the data to figure out what the new column names will be.
-#' Learn more in `vignette("pivot", "tidyr")`.
 #'
 #' @details
 #' The big difference to `pivot_wider()` for local data frames is that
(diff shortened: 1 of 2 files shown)


hadley:
I think just updating the documentation is fine. Maybe it's worth mentioning why? (i.e. "`pivot_wider()` isn't lazy, and in fact can't be lazy, because we need to look at the data to figure out what the new column names will be).  I guess we could also provide an additional argument that would let you supply the new names?

mgirlich:
To me adding an argument for the names does not make much sense. This feels like a super simplified specification and this would require additional testing and implementation. So, I just add some more information to the documentation.

R/verb-pivot-wider.R
```diff
@@ -3,6 +3,8 @@
 #' `pivot_wider()` "widens" data, increasing the number of columns and
 #' decreasing the number of rows. The inverse transformation is
 #' `pivot_longer()`.
+#' Note that `pivot_wider()` is not and cannot be lazy because we need to look
+#' at the data to figure out what the new column names will be.
 #' Learn more in `vignette("pivot", "tidyr")`.
```
hadley, line 8:
```suggestion
#' @description
#' `pivot_wider()` "widens" data, increasing the number of columns and
#' decreasing the number of rows. The inverse transformation is
#' `pivot_longer()`.
#' Learn more in `vignette("pivot", "tidyr")`.
#'
#' Note that `pivot_wider()` is not and cannot be lazy because we need to look
#' at the data to figure out what the new column names will be.
```

hadley (approved):

PR #801, opened 2022-03-18
Translate `cut()`

### LINKED ISSUE #697: support for cut() function in dbplyr (snowflake)
Hi, 

I'm using dbplyr to interface with Snowflake, and a very common task is to group numeric columns by bands to summarise output. With classic dplyr one would do group_by(cut(factor, c(1, 2, 3))) and it works wonders, but dbplyr does not appear to support it (at least in snowflake). 

So this is a post of two halves:
1) Could we please build in support for cut() function in SQL translations?
2) In the mean time, what would be the best workaround for this? 

e.g. i can build a function that creates an appropriate sql phrase, but how do I get it to be evaluated properly?

```
db_cut <- function(fact, bins) {
  k = length(bins)
  string = paste0("CASE WHEN ", fact, " < ", bins[1], " THEN '<", bins[1], "' ")   
  for (i in 2:k) {
    string = paste(string,
                   paste0("WHEN ", fact, " < ", bins[i], " THEN '[",
                          bins[i-1], "-", bins[i], ")' "))}
  string = paste0(string, " END")
  return(string)
}

this does not work...:
data %>%
  mutate(bands = db_cut(column_name, c(1, 2, 3))) %>%
  group_by(bands) %>%
  summarise(n()) 

```

am I missing something simple, or going down the wrong path?

Thanks!

mgirlich:
Closes #697


R/translate-sql-cut.R
```diff
@@ -0,0 +1,61 @@
...
+  breaks <- sort.int(as.double(breaks))
+  if (anyDuplicated(breaks)) {
+    abort("`breaks` are not unique.")
+  }
+  n <- length(breaks)
+  if (n == 1) {
+    abort("`breaks` must have size 2 or bigger.")
+  }
+
+  labels <- labels %||% TRUE
```
hadley, line 11: Should allow labels to be a character vector too:

```R
cut(1:3, breaks = c(0, 1.5, 2.5), labels = c("a", "b"))
```

tests/testthat/test-translate-sql-cut.R
```diff
@@ -0,0 +1,97 @@
...
+  expect_equal(
+    translate_sql(cut(x, c(0, 1, Inf))),
+    sql_lines(
+      "CASE",
+      "WHEN (`x` <= 0.0) THEN NULL",
+      "WHEN (`x` > 0.0 AND `x` <= 1.0) THEN '(0,1]'",
+      "WHEN (`x` > 1.0) THEN '(1,Inf]'",
+      "END"
+    )
+  )
```
hadley, line 78: Do these hit different code paths to the case with -Inf and Inf limits? If so, I think you could drop them.

R/translate-sql-cut.R
```diff
@@ -0,0 +1,61 @@
+sql_cut <- function(x, breaks, labels = NULL, right = TRUE) {
```
hadley, line 1: Worth implementing `include.lowest`?

R/translate-sql-cut.R
```diff
@@ -0,0 +1,61 @@
...
+  breaks <- sort.int(as.double(breaks))
```
hadley, line 2: This has the side-effects of stripping any `NA` from `breaks`? Is that deliberate? If so, should add a test.

hadley (approved):

tests/testthat/test-translate-sql-cut.R
mgirlich: 👍

R/translate-sql-cut.R
mgirlich: It is what `base::cut()` does but it looks terrible to me... So I'm not completely sure it is a good idea to also do that?

R/translate-sql-cut.R
mgirlich: The case work was a bit annoying but it makes sense after all I guess.

R/translate-sql-cut.R
hadley: If you don't have any strong feelings, I'd suggest adding `na.last = TRUE` to `sort.int` and then erroring if `breaks` contains any NAs. I implemented something similar for scales very recently: https://github.com/r-lib/scales/blob/main/R/label-number.r#L323-L332

hadley (approved):

hadley:
BTW I fixed the build failure on macOS-latest

PR #809, opened 2022-04-02
Use copy_inline() by default in memdb_frame()

krlmlr:
Should `auto_copy.tbl_sql()` also use `copy_inline()` ?

### DIFF
--- a/R/memdb.R
+++ b/R/memdb.R
@@ -4,8 +4,9 @@
 #' data frame in R, it creates a table in [src_memdb()].
 #'
 #' @inheritParams tibble::tibble
-#' @param name,.name Name of table in database: defaults to a random name that's
-#'   unlikely to conflict with an existing table.
+#' @param name,.name Name of table in database: if provided, [copy_to()]
+#'   will be used to create the table, otherwise [copy_inline()] creates
+#'   an anonymous table.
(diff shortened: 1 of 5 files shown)


mgirlich:
The following test fails

```r
test_that("can get 0 rows", {
  db <- memdb_frame(x = 1)
  out <- collect(head(db, 0))
  expect_equal(out, tibble(x = double()))
})
```

and shows one issue with SQLite and the values clause: it is not completely type stable :-/

mgirlich:
The tests fail because of type instability for 0 rows tables in SQLite. I did not yet have a closer look whether this is because of SQLite itself or because of RSQLite. @krlmlr can you have a look at that?

If this cannot be fixed in RSQLite then we could use `copy_to()` for 0 row data frames.

krlmlr:
I'm no longer sure that `copy_inline()` is really the right thing to do for SQLite databases, let's shelve it for now. Sorry for the noise.

mgirlich:
This might still be an option for `auto_copy()` though. Maybe as opt-in via `*_join(copy = "inline")`? What do you think?

krlmlr:
Still not sure. I prefer preparing the RHS of the join in advance and store it in a variable before joining, and never use auto-copy.

PR #810, opened 2022-04-04
Indentation `copy_inline()`

mgirlich:
@krlmlr I fixed the indentation for `copy_inline()`. This should also make it easier to force rendering the values clause in a CTE. But let's check this in another PR (see https://github.com/tidyverse/dbplyr/pull/668#discussion_r841025763).

## Examples with CTE

```r
dbplyr_uncount(df, w) %>% show_query(cte = TRUE)
```

```sql
WITH `q01` AS (
  SELECT CAST(`..dbplyr_row_id` AS INTEGER) AS `..dbplyr_row_id`
  FROM (
    SELECT NULL AS `..dbplyr_row_id`
    WHERE (0 = 1)
    UNION ALL
    VALUES (1)
  ) AS `values_table`
),
`q02` AS (
  SELECT `x`, `w`, `..dbplyr_row_id`
  FROM `test` AS `LHS`
  INNER JOIN `q01` AS `RHS`
    ON (`RHS`.`..dbplyr_row_id` <= `LHS`.`w`)
)
SELECT `x`
FROM `q02`
```

```r
tmp <- copy_inline(src_memdb()$con, tibble(x = 1:2, y = "c"))
tmp2 <- memdb_frame(x = 1:3)

left_join(tmp2, tmp, by = "x") %>% 
  show_query(cte = TRUE)
```

```sql
WITH `q01` AS (
  SELECT CAST(`x` AS INTEGER) AS `x`, CAST(`y` AS TEXT) AS `y`
  FROM (
    SELECT NULL AS `x`, NULL AS `y`
    WHERE (0 = 1)
    UNION ALL
    VALUES (1, 'c'), (2, 'c')
  ) AS `values_table`
)
SELECT `LHS`.`x` AS `x`, `y`
FROM `dbplyr_216` AS `LHS`
LEFT JOIN `q01` AS `RHS`
  ON (`LHS`.`x` = `RHS`.`x`)
```


krlmlr (approved):
Thanks, looks good.

IIUC, passing `lvl` to `sql_values()` is needed only for the non-CTE case? If CTE is becoming the default at some point, do we even need to thread it around everywhere? An alternative would be to add indentation after rendering. (Careful with embedded newlines!)

``` r
dbplyr:::sql_values(dbplyr::simulate_mssql(), data.frame(a = "\n"))
#> <SQL> SELECT TRY_CAST(`a` AS VARCHAR(MAX)) AS `a`
#> FROM (
#>   (
#>     SELECT NULL AS `a`
#>     WHERE (0 = 1)
#>   )
#>   UNION ALL
#>   (
#>   VALUES ('
#> ')
#>   )
#> ) `values_table`
```

<sup>Created on 2022-04-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mgirlich:
> IIUC, passing `lvl` to `sql_values()` is needed only for the non-CTE case?

Yes, `lvl` is only needed for the indentation, so it should be unnecessary if only CTEs are used.

> If CTE is becoming the default at some point, do we even need to thread it around everywhere?

This depend

An alternative would be to add indentation after rendering. (Careful with embedded newlines!)

PR #819, opened 2022-04-08
Avoid sql_escape_ident() in sql_table_prefix()

krlmlr:
This helps #748: without this change, updating dm to be compatible with the new `rows_*()` implementation will be hard.

### DIFF
--- a/R/backend-sqlite.R
+++ b/R/backend-sqlite.R
@@ -129,7 +129,11 @@ sql_query_wrap.SQLiteConnection <- function(con, from, name = unique_subquery_na
     if (is.null(name)) {
       build_sql(sql_indent_subquery(from, con, lvl), con = con)
     } else {
-      build_sql(sql_indent_subquery(from, con, lvl), " AS ", ident(name), con = con)
+      ident_name <- name %||% unique_subquery_name()
+      if (!is.ident(ident_name)) {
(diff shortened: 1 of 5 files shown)


R/backend-sqlite.R
```diff
@@ -129,7 +129,11 @@ sql_query_wrap.SQLiteConnection <- function(con, from, name = unique_subquery_na
     if (is.null(name)) {
       build_sql(sql_indent_subquery(from, con, lvl), con = con)
     } else {
-      build_sql(sql_indent_subquery(from, con, lvl), " AS ", ident(name), con = con)
+      ident_name <- name %||% unique_subquery_name()
+      if (!is.ident(ident_name)) {
```
hadley, line 133: Since this seems to be a repeated pattern, might be worth extracting it out?

```R
as_subquery_name <- function(x) {
  if (is.ident(x)) {
    x
  } else if (is.null(x) {
    unique_subquery_name()
  } else {
    ident(x)
  }
}
```

hadley (approved):

krlmlr:
Good idea, done. I wonder if we want change the default to `name = NULL` instead of `name = unique_subquery_name()`, #826.

mgirlich (approved):

PR #823, opened 2022-04-09
Fix copy_inline() for MSSQL

krlmlr:
The `VALUES` keyword needs a table name and column names.


R/verb-copy-to.R
```diff
@@ -227,11 +227,25 @@ sql_values_clause <- function(con, df, row = FALSE, lvl = 0) {
...
 
-  union_query <- set_op_query(
-    null_row_query,
-    sql_format_clauses(list(sql_clause("VALUES", rows_sql)), lvl = lvl + 1, con = con),
-    type = "UNION ALL",
-  )
+  rows_clauses <- list(sql_clause("VALUES", rows_sql))
+
+  if (derived) {
+    rows_query <- sql_format_clauses(rows_clauses, lvl = 0, con = con)
```
mgirlich, line 233: The formatting is incorrect when there are many values

```sql
SELECT CAST("x" AS INTEGER) AS "x", CAST("y" AS TEXT) AS "y"
FROM (
  (
    SELECT NULL AS "x", NULL AS "y"
    WHERE (0 = 1)
  )
  UNION ALL
  (
    SELECT *
    FROM (
VALUES
  (1, 'a'),
  (2, 'b'),
  (3, 'c'),
  (4, 'd'),
  (5, 'e'),
  (6, 'f'),
  (7, 'g'),
  (8, 'h')
    ) drvd("x", "y")
  )
) "values_table"
```

but I don't think it's worth spending much time on fixing this.
Maybe add a comment that this is a known issue?

mgirlich (approved):

mgirlich:
Ah, it's so nice to see how many different standards there are in SQL. In particular with MS SQL always being the odd one out...

Implementing `copy_inline()` correctly turns out to be way more complex than I thought initially.

R/verb-copy-to.R
```diff
@@ -227,11 +227,25 @@ sql_values_clause <- function(con, df, row = FALSE, lvl = 0) {
...
+
+    derived_sql <- sql(paste0("drvd(", escape(ident(colnames(df)), con = con), ")"))
+
+    rows_query <- sql_query_select(
+      con,
+      sql("*"),
+      # sql_subquery() can't use sql() or ident_q() as `name` argument
```
mgirlich, line 240: It is quite annoying that `ident(ident(x))` actually does the quoting twice. Might be worth trying to fix this at some point...

krlmlr: `ident(ident_q(x))` or `ident(sql(x))` are problematic. Can we really fix this without breaking anything?

mgirlich: I'm not sure. But I don't see where I actually want `ident()` to overwrite the `ident_q` or `sql` class. So, this feels more like a common source for errors than for intended behaviour. But this is something for a different PR.

PR #826, opened 2022-04-11
Use name = NULL default for sql_query_wrap()

krlmlr:
`unique_subquery_name()` is private, do we really want to expose it?


mgirlich:
Good catch. I don't think we want to expose it.

mgirlich (approved):

PR #841, opened 2022-04-25
Highlight keywords in `show_query()`

### LINKED ISSUE #644: Code highlighting in console
Would be nice to have code highlighting. Together with #642 it would be much easier to skim over SQL code generated by `dbplyr`.

mgirlich:
Closes #644.
It uses a global option to avoid having to add an argument `colour` to lots of functions (as proposed by Hadley https://github.com/tidyverse/dbplyr/issues/644#issuecomment-1095590348). Probably, this misses some places but I like how simple this was to add.

### DIFF
--- a/R/utils-format.R
+++ b/R/utils-format.R
@@ -21,6 +21,14 @@ indent_print <- function(x) {
   indent(utils::capture.output(print(x)))
 }
 
+style_kw <- function(x) {
+  if (dbplyr_use_colour()) {
+    cli::col_blue(x)
+  } else {
+    x
+  }
+}
+
 # function for the thousand separator,
 # returns "," unless it's used for the decimal point, in which case returns "."
 'big_mark' <- function(x, ...) {
--- a/R/utils-format.R
+++ b/R/utils-format.R
@@ -27,3 +27,7 @@ indent_print <- function(x) {
   mark <- if (identical(getOption("OutDec"), ",")) "." else ","
   formatC(x, big.mark = mark, ...)
 }
+
+dbplyr_use_colour <- function() {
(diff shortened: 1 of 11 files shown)


R/utils-format.R
```diff
@@ -27,3 +27,7 @@ indent_print <- function(x) {
   mark <- if (identical(getOption("OutDec"), ",")) "." else ","
   formatC(x, big.mark = mark, ...)
 }
+
+dbplyr_use_colour <- function() {
+  getOption("dbplyr_use_colour", FALSE) && cli::num_ansi_colors() > 1L
+}
```
hadley, line 33: I'd consider also adding

```R
style_kw <- function(x) {
  if (dbplyr_use_colour()) {
    cli::col_blue(x)
  } else {
    x
  }
}
```

And you don't need to check `num_ansi_colors()` since cli already does that for you.

hadley (approved):
It's probably worth adding one test for this where you do `local_reproducible_output(crayon = TRUE)` to force the ANSI escapes to appear in the snapshot test.

This also definitely deserves a news bullet 😄

tests/testthat/test-verb-select.R
```diff
@@ -122,6 +122,19 @@ test_that("mutate collapses over nested select", {
   expect_snapshot(lf %>% mutate(a = 1, b = 2) %>% select(x))
 })
 
+test_that("", {
```
hadley, line 125:
```suggestion
test_that("output is styled", {
```

hadley (approved):

tests/testthat/test-verb-select.R
mgirlich: Ooops, thanks for catching!

PR #850, opened 2022-04-29
Fix `remote_name()`

### LINKED ISSUE #639: [question] expected output of `x %>% dbplyr::compute() %>% dbplyr::remote_name()`
For example, consider

```
df1 <- memdb_frame(x = 1)
df2 <- df1 %>% mutate(y = 0)
dfc <- df2 %>% compute()
print(dfc %>% remote_name())
```

With dbplyr version 2.1.1 the output is `NULL`, but I seem to remember some earlier version of dbplyr would produce a non-`NULL` output, and also based on how I understood the `dbplyr` documentation, a non-`NULL` output is expected, right?

BTW the aforementioned `NULL` / non-`NULL` outputs also apply to Spark dataframes in `sparklyr` or other DB backends, not just `memdb_frame`s.

mgirlich:
Closes #639.


tests/testthat/test-remote.R
```diff
@@ -6,6 +6,30 @@ test_that("remote_name returns null for computed tables", {
   expect_equal(remote_name(mf2), NULL)
 })
 
+test_that("remote_name returns name for grouped tables", {
```
hadley, line 9: I'd probably combine all of these into a single test since they're each rather simple and are strongly related.

hadley (approved):

R/remote.R
```diff
@@ -24,10 +24,29 @@
...
+  }
+
+  vars_base <- op_vars(lq$from)
+  if (!is_select_trivial(lq$select, vars_base)) {
+    return()
+  }
+
+  if (!(is_empty(lq$where) && is_empty(lq$order_by) && is_false(lq$distinct) && is_empty(lq$limit))) {
     return()
+  }
```
krlmlr, line 47: Longer, but easier to read?

```suggestion
  if (!is_empty(lq$where)) {
    return()
  }

  if (!is_empty(lq$order_by)) {
    return()
  }

  if (!is_false(lq$distinct)) {
    return()
  }

  if (!is_empty(lq$limit)) {
    return()
  }
```

PR #853, opened 2022-04-30
New in_catalog()

### LINKED ISSUE #806: [FR] Allow database name in `in_schema`
When connecting to an MSSQL Server, we can have several databases from which we can pull data. To get the quoting straight, one has to add an additional `sql` call to `in_schema` to be able to get the data in a particular database, which is not that clear:

```r
library(dbplyr)

(s1 <- in_schema("mydatabase.myschema", "mytable"))
## <SCHEMA> `mydatabase.myschema`.`mytable`
(s2 <- in_schema(sql("mydatabase.myschema"), "mytable"))
## <SCHEMA> mydatabase.myschema.`mytable`
# library(dplyr)
# con <-  dbConnect(odbc::odbc(), "mydatabase_server")
# tbl(con, s1) ## will not work
# tbl(con, s2) ## will work
```

Admittedly, my knowledge of standard SQL is mediocre at best, so there may be very good reasons for not adding a `catalog/database` parameter to `in_schema`, but at least the proper way of addressing a table in a schema in a catalog / database should be documented somewhere. w/o  [SO](https://stackoverflow.com/questions/71410349/store-new-permanent-table-in-schema-using-compute) I would yet not have any idea of how to pull data with the fully qualified name. 

Thus, either we add that to the docs or even add a new parameter to `in_schema`.

krlmlr:
Creating objects of a new `"dbplyr_catalog"` class. We could also make the `catalog` component optional in the existing `"dbplyr_schema"` class.

Closes #806.


mgirlich:
I wonder how many people would rather search for `in_database()` instead of `in_catalog()`.

mgirlich (approved):

krlmlr:
Good point, though everything in dbplyr is about databases, one way or another. We could provide both.

krlmlr:
@hadley: Okay with a new class? Should we establish a class hierarchy, with a common base class, or perhaps use only the `"dplyr_schema"` class (i.e., revert d8ab76d)?

hadley (approved):
The implementation looks fine to me.

PR #855, opened 2022-05-02
Add `sql_query_insert.PqConnection()`

mgirlich:
This method uses `ON CONFLICT DO NOTHING`

PR #856, opened 2022-05-02
`rows_insert/upsert()`: add argument `method`

### LINKED ISSUE #840: Add argument to `rows_insert()` and `rows_upsert()` whether to use unique constraint
Some databases (e.g. SQLite and Postgres) have special support for inserting/upserting when there is a unique constraint on the `by` columns. The user should be able to choose whether to use them or not.

mgirlich:
Closes #840. This also affects #855.

This gives the user the possibility to use a fallback method. This can be useful for older database versions or e.g. in Postgres there is no unique index.


tests/testthat/_snaps/backend-mssql.md
```diff
@@ -369,3 +369,31 @@
...
+    Output
+      <SQL> WITH `updated` AS (
+        UPDATE `df_x`
+        SET `c` = `...y`.`c`, `d` = `...y`.`d`
+        FROM (
+        SELECT `a`, `b`, `c` + 1.0 AS `c`, `d`
+        FROM `df_y`
+      ) `...y`
+        WHERE (`...y`.`a` = `df_x`.`a`) AND (`...y`.`b` = `df_x`.`b`)
+        OUTPUT 'INSERTED'.`*`
```
krlmlr, line 386: Does the `*` need to be escaped here?

R/backend-postgres.R
```diff
@@ -250,10 +250,24 @@ sql_query_explain.PqConnection <- function(con, sql, format = "text", ...) {
...
 sql_query_explain.PostgreSQL <- sql_query_explain.PqConnection
 
+#' Upsert using the `ON CONFLICT ... DO UPDATE` clause
+#'
+#' The `ON CONFLICT ... DO UPDATE` clause requires a unique index on the columns
+#' in `by`. One of the advantages is better concurrency behaviour. See the
+#' [Postgres Wiki](https://wiki.postgresql.org/wiki/UPSERT#SQL_MERGE_syntax) for
+#' more information on the advantages and disadvantages compared to other
+#' upsert implementations.
+#'
```
krlmlr, line 260: Did you mean to add this documentation block to the generic?

krlmlr (commented):
Thanks. Can we also add an integration test for MSSQL?

I see fairly little use for "upsert" to begin with, others might disagree. Should we settle for one method, `MERGE` or not, that "just works" in a lab setting? Then we can see if it's used and if it works in practice, and improve.

mgirlich:
> Thanks. Can we also add an integration test for MSSQL?

👍 

> I see fairly little use for "upsert" to begin with, others might disagree.

Then I am one of the others as I frequently use upsert 😄 

> Should we settle for one method, `MERGE` or not, that "just works" in a lab setting? Then we can see if it's used and if it works in practice, and improve.

I would be one of the happy users for both Postgres methods which is why I added them in the first place.

krlmlr:
Should the `vendor_method` be a string instead, which later might allow us to choose other methods? Do we really need new `sql_` generics?

R/backend-postgres.R
mgirlich: What's the best way to document the valid strings for `method` of the `sql_query_upsert()` methods?

mgirlich:
I agree that it makes more sense if `method` is a string.

mgirlich:
Argh, too bad, SQL Server does [not allow `UPDATE` in a CTE](https://docs.microsoft.com/en-us/sql/t-sql/queries/with-common-table-expression-transact-sql?view=sql-server-ver15#arguments)

> CTE_query_definition
Specifies a SELECT statement whose result set populates the common table expression.

R/backend-mssql.R
```diff
@@ -93,7 +93,10 @@ simulate_mssql <- function(version = "15.0") {
 #' @export
 `sql_query_insert.Microsoft SQL Server` <- function(con, x_name, y, by, ...,
                                                     conflict = c("error", "ignore"),
-                                                    returning_cols = NULL) {
+                                                    returning_cols = NULL,
+                                                    method = NULL) {
+  method <- method %||% "where_not_exists"
+  arg_match(method, "where_not_exists", error_arg = "method")
```
hadley, line 99: I wonder if it's worth just leaving this unchecked so that the user can use whatever the database supports. Otherwise I worry that we'd need to keep this in sync with database version, which is going to be quite a bit of work for little gain.

hadley (approved):

R/backend-mssql.R
mgirlich: I thought about leaving this unchecked but I don't really see how this is going to work.

Let's say someone wants to support `method = "fancy"` for Postgres. Then `sql_query_insert.PqConnection()` needs to know about `method = "fancy"` to generate the correct SQL for this, i.e. we need something like this:

```r
sql_query_insert.PqConnection <- function(...) {
  if (method == "fancy") {
    sql_query_insert_fancy(...)
  }

  ...
}
```

R/backend-postgres.R
mgirlich: @hadley Do you have an idea

* how/where to document which methods are supported by which backend
* and where to describe the SQL code for a method

?

R/backend-mssql.R
hadley: Ooooh, I just thought this affected the value of some component of the generated SQL.

R/backend-postgres.R
hadley: It's probably easiest to document the methods with the generic (i.e. `@rdname sql_query_upsert`) since that way the majority of arguments are the same, but then you'll need to pull that out in to its own `.Rd` topic. Then I'd construct a hand written `@param method` that explained the basic options, with more details available in the `@details`. In the interests of getting this release out the door, I'd suggest focussing on the minimum possible for now, and we can come back to it after the release.

PR #860, opened 2022-05-05
Add more translations for Snowflake

fh-afrachioni:
This request adds a number of additional translations for Snowflake backends, with the intent to translate all functions currently available for Postgres.  Specifically, support is added for:

**`base`**
- `log10`
- `grepl`
- `round`
- `paste` and `paste0`

**`stringr`**
- `str_c`
- `str_locate`
- `str_detect`
- `str_replace` and `str_replace_all`
- `str_remove` and `str_remove_all`
- `str_trim` and `str_squish`

**`lubridate`**
- `day`, `mday`, `wday`, `yday`, `week`, `isoweek`, `month`, `quarter`, `isoyear`
- `seconds`, `minutes`, `hours`, `days`, `weeks`, `months`, `years`
- `floor_date`

This fixes issue #935


fh-afrachioni:
Closing for the time being to add a few more translations.

fh-afrachioni:
Reopening after adding backend testing.  Any feedback would be appreciated.

R/backend-snowflake.R
```diff
@@ -37,3 +217,40 @@ simulate_snowflake <- function() simulate_dbi("Snowflake")
...
+snowflake_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed = FALSE, useBytes = FALSE) {
+  # https://docs.snowflake.com/en/sql-reference/functions/regexp.html
+  if (any(c(perl, fixed, useBytes))) {
+    abort("`perl`, `fixed` and `useBytes` parameters are unsupported")
+  }
+  # REGEXP on Snowflaake "implicitly anchors a pattern at both ends", which
+  # grepl does not.  Left- and right-pad `pattern` with .* to get grepl-like
+  # behavior
+  if (ignore.case) {
+    sql_expr(LOWER((!!x)) %REGEXP% (".*" || LOWER((!!pattern)) || ".*"))
```
hadley, line 230: I don't think this is quite right because it would convert (e.g.) `\D` to `\d`, changing the meaning of the regexp. I think you probably should just error if `ignore.case` is `TRUE`.

R/backend-snowflake.R
```diff
@@ -37,3 +217,40 @@ simulate_snowflake <- function() simulate_dbi("Snowflake")
...
+  } else {
+    sql_expr(((!!x)) %REGEXP% (".*" || (!!pattern) || ".*"))
+  }
+}
+snowflake_round <- function(x, digits = 0L) {
+  digits <- as.integer(digits)
+  sql_expr(round(((!!x)) %::% FLOAT, !!digits))
+}
+
+# On Snowflake, CONCAT_WS is null if any of its arguments are null.  Paste
```
hadley, line 240: That's the same behaviour as `str_c()` so you might want to use a different translation there.

R/backend-snowflake.R
```diff
@@ -18,13 +18,193 @@ NULL
...
+        if (!label) {
+          sql_expr(EXTRACT("month", !!x))
+        } else {
+          if (abbr) {
+            sql_expr(MONTHNAME(!!x))
+          } else {
+            sql_expr(
+              DECODE(
+                EXTRACT("month", !!x),
+                1, "January  ",
```
hadley, line 122: Should these have trailing spaces?

R/backend-snowflake.R
```diff
@@ -18,13 +18,193 @@ NULL
...
+          )
+        } else if (label && abbr) {
+          sql_expr(DAYNAME(!!x))
+        } else {
+          abort("Unrecognized arguments to `wday`")
+        }
+      },
+      yday = function(x) sql_expr(EXTRACT("dayofyear", !!x)),
+      week = function(x) {
+        sql_expr(FLOOR((EXTRACT("dayofyear", !!x) - 1L) / 7L) + 1L)
```
hadley, line 108: Are you sure that's the correct calculation? I think it might be too simple.

R/backend-snowflake.R
```diff
@@ -18,13 +18,193 @@ NULL
...
+      },
+      # On Snowflake, REGEXP_REPLACE is used like this:
+      # REGEXP_REPLACE( <subject> , <pattern> [ , <replacement> ,
+      #                <position> , <occurrence> , <parameters> ] )
+      # so we must set <occurrence> to 1 if not replacing all.  See:
+      # https://docs.snowflake.com/en/sql-reference/functions/regexp_replace.html
+      # Also, Snowflake needs backslashes escaped, so we must increase the
+      # level of escaping by 1
+      str_replace = function(string, pattern, replacement) {
+        pattern <- gsub("\\\\", "\\\\\\\\", pattern)
```
hadley, line 51: I'd use `fixed = TRUE` here and make sure you cover in the unit tests.

R/backend-snowflake.R
```diff
@@ -18,13 +18,193 @@ NULL
...
+        sql_expr(DATE_TRUNC(!!unit, !!x))
+      }
+    ),
+    sql_translator(
+      .parent = base_agg,
+      cor = sql_aggregate_2("CORR"),
+      cov = sql_aggregate_2("COVAR_SAMP"),
+      all = sql_aggregate("BOOLAND_AGG", "all"),
+      any = sql_aggregate("BOOLOR_AGG", "any"),
+      exactly_one = sql_aggregate("BOOLXOR_AGG", "exactly_one"),
```
hadley, line 188: There's no R function `exactly_one` so I wouldn't include this in the translation.

R/backend-snowflake.R
```diff
@@ -37,3 +217,40 @@ simulate_snowflake <- function() simulate_dbi("Snowflake")
 # Link to full list: https://docs.snowflake.com/en/sql-reference/sql-all.html
 #' @export
 sql_table_analyze.Snowflake <- function(con, table, ...) {}
+
+snowflake_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed = FALSE, useBytes = FALSE) {
+  # https://docs.snowflake.com/en/sql-reference/functions/regexp.html
+  if (any(c(perl, fixed, useBytes))) {
```
hadley, line 223: I'd write `perl || fixed || useBytes` just to reinforce that these should be scalars.

hadley (commented):
Thanks for working on this! A bunch of small feedback follows.

R/backend-snowflake.R
fh-afrachioni: Good point, I'll implement this.

R/backend-snowflake.R
fh-afrachioni: I was trying to be consistent with the postgres implementation, which uses `TO_CHAR(., "Month")`, which yields
> full upper case month name (blank-padded to 9 chars)

As described here: https://www.postgresql.org/docs/current/functions-formatting.html

But maybe this isn't a good reason; I'll strip the whitespace in my update.

R/backend-snowflake.R
fh-afrachioni: Sounds good, will remove!

R/backend-snowflake.R
fh-afrachioni: From the `lubridate` docs,
> `week()` returns the number of complete seven day periods that have occurred between the date and January 1st, plus one.
>
> `isoweek()` returns the week as it would appear in the ISO 8601 system, which uses a reoccurring leap week.

I think it _should_ be this simple, and I tested it against both the `lubridate` and postgres-dbplyr implementations for a bunch of dates.

hadley: Ok, great.

hadley:
@mgirlich could you please take a quick look too? (just to make sure I haven't forgotten anything major since I haven't worked on dbplyr in a while)

R/backend-snowflake.R
```diff
@@ -18,13 +18,198 @@ NULL
...
+        build_sql("INTERVAL '", x, " week'")
+      },
+      months = function(x) {
+        build_sql("INTERVAL '", x, " month'")
+      },
+      years = function(x) {
+        build_sql("INTERVAL '", x, " year'")
+      },
+      # https://docs.snowflake.com/en/sql-reference/functions/date_trunc.html
+      floor_date = function(x, unit = "seconds") {
```
mgirlich, line 182: The default is plural, below is always singular. I see this is also the case for the postgres backend...
Either you simply fix the default to `"seconds"` or if you want to be more user friendly you could also allow the plural version. Up to you.

R/backend-snowflake.R
```diff
@@ -37,3 +222,30 @@ simulate_snowflake <- function() simulate_dbi("Snowflake")
 # Link to full list: https://docs.snowflake.com/en/sql-reference/sql-all.html
 #' @export
 sql_table_analyze.Snowflake <- function(con, table, ...) {}
+
+snowflake_grepl <- function(pattern, x, ignore.case = FALSE, perl = FALSE, fixed = FALSE, useBytes = FALSE) {
+  # https://docs.snowflake.com/en/sql-reference/functions/regexp.html
+  if (perl || fixed || useBytes || ignore.case) {
+    abort("`perl`, `fixed`, `useBytes`, and `ignore.case` parameters are unsupported")
```
mgirlich, line 229: For consistency

```suggestion
    cli_abort("{.arg {c('perl', 'fixed', 'useBytes')}} parameters are unsupported.")
```

R/backend-snowflake.R
```diff
@@ -18,13 +18,198 @@ NULL
...
+    sql_translator(
+      .parent = base_odbc_scalar,
+      log10 = function(x) sql_expr(log(10, !!x)),
+      grepl = snowflake_grepl,
+      round = snowflake_round,
+      paste = snowflake_paste(" "),
+      paste0 = snowflake_paste(""),
+      str_c = function(..., sep = "", collapse = NULL) {
+        if (!is.null(collapse)) {
+          abort(c(
```
mgirlich, line 31: For consistency this should use `cli_abort()` (with `{.arg collapse}`, `{.fn str_c}`, ...)

mgirlich (changes requested):

mgirlich:
@fh-afrachioni Looks great overall. Thanks!

R/backend-snowflake.R
fh-afrachioni: Good catch, maybe I'll follow up with an update to the postgres backend.  Since the singular and plural versions are all supported on both Snowflake ([here](https://docs.snowflake.com/en/sql-reference/functions-date-time.html#label-supported-date-time-parts)) and lubridate, I'm expanding the arg list to include both.

fh-afrachioni:
Thanks for the review, @mgirlich; I just updated with the requested changes.  Happy to address anything else!

mgirlich (approved):

fh-afrachioni:
Thanks for the review and approval, @hadley and @mgirlich!  Do you know when the next release is planned?  I have a team that would love to use these features within the next few weeks.

hadley:
We've accumulated quite a few new features lately thanks to @mgirlich's hard work. So it would be nice to get a release out in the not too distant future. But no concrete plans yet.

PR #872, opened 2022-05-13
Metadata in lazy query

mgirlich:
This PR

* adds a common generator function `lazy_query()` that supports all lazy queries.
* adds input checks on the `lazy_*_query()` functions
* The `from` field in `lazy_select_query()` is renamed to `x` for consistency with the other lazy queries. This might require another small PR in sparklyr.
* The `x` and `y` field of the queries are now always lazy queries instead of lazy tables. This might require another small PR in sparklyr.
* allows to store "metadata" (`group_vars`, `order_vars`, and `frame`) in each lazy query (previously it was only really supported in `<lazy_select_query>`
* This allows to simplify `add_order()`, `add_group_by()`
* This allows to remove some methods for `op_frame()`, `op_grps()`, and `op_sort()`

If this is merged then sparklyr might need some minor changes.


R/lazy-join-query.R
```diff
@@ -1,34 +1,92 @@
...
+  }
+
+  if (!identical(names(vars), c("alias", "x", "y", "all_x", "all_y"))) {
+    # TODO use `cli_abort()` after https://github.com/r-lib/rlang/issues/1386
+    # is fixed
+    abort("`vars` must have fields `alias`, `x`, `y`, `all_x`, and `all_y`", .internal = TRUE)
+  }
+
+  n <- vctrs::vec_size(vars$alias)
+  vctrs::vec_assert(vars$alias, character())
```
hadley, line 45: These all should to use `arg` and `call` to generate maximally informative errors. We might need to pass the call env through a few functions, but it's probably worth it.

hadley (approved):

PR #875, opened 2022-05-18
Inline `select()` in `*_join()`

### LINKED ISSUE #867: Combine `select()` + `*_join()` in a single query
``` r

lf2 <- lazy_frame(x = 1, b = 2, c = 3)

left_join(
  lf,
  lf2 %>% select(x, b),
  by = "x"
)

```

Currently

```sql
FROM `df` AS `LHS`
LEFT JOIN (
  SELECT `x`, `b`
  FROM `df`
) `RHS`
  ON (`LHS`.`x` = `RHS`.`x`)
```

Should be

```sql
FROM `df` AS `LHS`
LEFT JOIN `df` AS `RHS`
  ON (`LHS`.`x` = `RHS`.`x`)
```

mgirlich:
Closes #867.

```r

lf <- lazy_frame(x1 = 10, a = 1, y = 3, .name = "lf1")
lf2 <- lazy_frame(x2 = 10, b = 2, z = 4, .name = "lf2")
```

`left/right/full/inner_join()`

```r
left_join(
  lf %>% select(a2 = a, x = x1),
  lf2 %>% select(x = x2, b),
  by = "x"
)
```

Before

```sql
SELECT `a2`, `LHS`.`x` AS `x`, `b`
FROM (
  SELECT `a` AS `a2`, `x1` AS `x`
  FROM `lf1`
) `LHS`
LEFT JOIN (
  SELECT `x2` AS `x`, `b`
  FROM `lf2`
) `RHS`
  ON (`LHS`.`x` = `RHS`.`x`)
```

After

```sql
SELECT `a` AS `a2`, `x1` AS `x`, `b`
FROM `lf1` AS `LHS`
LEFT JOIN `lf2` AS `RHS`
  ON (`LHS`.`x1` = `RHS`.`x2`)
```

`semi/anti_join()`

```r
semi_join(
  lf %>% select(a2 = a, x = x1),
  lf2 %>% select(x = x2, b),
  by = "x"
)
```

Before

```sql
SELECT * FROM (
  SELECT `a` AS `a2`, `x1` AS `x`
  FROM `lf1`
) `LHS`
WHERE EXISTS (
  SELECT 1 FROM (
  SELECT `x2` AS `x`, `b`
  FROM `lf2`
) `RHS`
  WHERE (`LHS`.`x` = `RHS`.`x`)
)
```

After

```sql
SELECT `a` AS `a2`, `x1` AS `x`
FROM `lf1` AS `LHS`
WHERE EXISTS (
  SELECT 1 FROM (
  SELECT `x2` AS `x`, `b`
  FROM `lf2`
) `RHS`
  WHERE (`LHS`.`x1` = `RHS`.`x`)
)
```

The `semi/anti_join()` could be improved further but I'm not sure it's worth it (mostly interesting when the join columns have different names and one uses `rename()` instead of using a named `by`).


R/lazy-join-query.R
```diff
@@ -7,6 +7,9 @@ lazy_join_query <- function(x,
                             by,
                             suffix = c(".x", ".y"),
                             na_matches = c("never", "na"),
+                            group_vars = NULL,
```
mgirlich, line 10: `add_group_by()` adds the field `group_vars` to any `lazy_query` object. Similarly for `order_vars` and `frame`. So it makes sense to be able to explicitly set them when creating these objects.

mgirlich:
@hadley Would be great if you could give feedback

* whether you see problems with inlining
* you see edge cases that need to be tested

R/lazy-select-query.R
```diff
@@ -94,6 +94,42 @@ update_lazy_select <- function(select, vars) {
...
+  if (!inherits(x, "lazy_select_query")) {
+    return(FALSE)
+  }
+
+  if (!purrr::every(x$select$expr, is_symbol)) {
+    return(FALSE)
+  }
+
+  vars_prev <- op_vars(x$x)
+  if (select == "trivial" && !is_select_trivial(x$select, vars_prev)) {
```
hadley, line 110: Put `is_select_trivial()` near this function?

R/lazy-select-query.R
```diff
@@ -94,6 +94,42 @@ update_lazy_select <- function(select, vars) {
   select
 }
 
+is_lazy_select_query_simple <- function(x,
```
hadley, line 97: I think a brief comment explaining what "trivial" means here would be useful

tests/testthat/_snaps/verb-arrange.md
```diff
@@ -177,7 +177,8 @@
       i Do you need to move arrange() later in the pipeline or use window_order() instead?
     Output
       <SQL>
-      SELECT * FROM (
+      SELECT *
+      FROM (
```
hadley, line 181: I definitely like the `FROM` on its own line.

hadley (approved):
Your logic makes sense to me, and I can see any obvious holes in it.

Is it worth including a test for some of the suffix cases? I'm not sure how much that overlaps with this code.

R/lazy-select-query.R
mgirlich: I added a comment and renamed `trivial` to `identity` which feels a bit clearer.

R/lazy-select-query.R
mgirlich: I added a comment and renamed `trivial` to `identity` which feels a bit clearer.

Issue #870, opened 2022-05-11
Consider support for `HAVING` clause?


hadley:
Low priority, but it would be nice to avoid the subquery if you do a `filter()` after a `summarise()`.

mgirlich:
I gave this some thought and toyed around a little with `HAVING`. I think the scope of it is quite limited but still useful.
From the [Postgres documentation](https://www.postgresql.org/docs/current/sql-select.html#SQL-HAVING)

> Each column referenced in condition must unambiguously reference a grouping column, unless the reference appears within an aggregate function or the ungrouped column is functionally dependent on the grouping columns.

So, the SQL query must look like

```sql
SELECT <grouping columns>, <aggregated columns>
GROUP BY <grouping columns>
HAVING ...
```

and with dplyr verbs I think this can only be achieved via

1. `summarise() + filter()`

```r
df %>% 
  group_by(<grouping columns>) %>% 
  summarise(<aggregated columns>) %>% 
  filter(<condition>)
```

where `<condition>` may only refer to `<grouping columns>` and `<aggregated columns>`. E.g.

```r
lazy_frame(g = 1, x = 1, y = 1) %>% 
  group_by(g) %>% 
  summarise(x_mean = mean(x)) %>% 
  filter(x_mean > 1, g == 1)
```

And one cannot use `x` or `y` in `filter()` (though in SQL this would be possible).
This case should be relatively simple. It could be done directly in `add_filter()` but also later in `sql_optimise()` (though this is probably a bit more difficult).

2. `filter() + summarise()`

```r
lazy_frame(g = 1, x = 1, y = 1) %>% 
  group_by(g) %>% 
  filter(sum(x) > 1, mean(y) == 1, g == 1) %>% 
  summarise(x_mean = mean(x))
```

Now, it is also possible to use `x` and `y` in `filter()` if they are both used in an aggregate function. This can be done in `add_summarise()` or in `sql_optimise()`.

@hadley Do you think these are all cases where `HAVING` can be used?

hadley:
Yeah, I think that's right.

We could also consider supporting `QUALIFY` at the same time. I know less about whether it's supported in most databases, but it does a similar thing but for filtering the result of window functions (i.e. it would be useful for `summarise()` after a grouped `mutate()`). Again, not terribly important, but if it fit in the same framework as `HAVING` it might make implementing both slightly more worthwhile.

mgirlich:
[...]

PR #877, opened 2022-05-18
Use `HAVING` for `summarise()` + `filter()`

mgirlich:
This solves the first of two cases where `HAVING` can be used according to #870.

```r

lazy_frame(g = 1, x = 1, y = 1) %>% 
  group_by(g) %>% 
  summarise(x_mean = mean(x, na.rm = TRUE)) %>% 
  filter(x_mean > 1, g == 1)
```

Before

```sql
SELECT *
FROM (
  SELECT `g`, AVG(`x`) AS `x_mean`
  FROM `df`
  GROUP BY `g`
) `q01`
WHERE (`x_mean` > 1.0) AND (`g` = 1.0)
```

After

```sql
SELECT `g`, AVG(`x`) AS `x_mean`
FROM `df`
GROUP BY `g`
HAVING (AVG(`x`) > 1.0) AND (`g` = 1.0)
```


R/tidyeval.R
```diff
@@ -231,7 +231,11 @@ fun_name <- function(fun) {
 
 replace_sym <- function(call, sym, replace) {
   if (is_symbol(call, sym)) {
-    replace
+    if (is_list(replace)) {
+      replace[[which(call == syms(sym))]]
```
hadley, line 235: I'm mildly surprised that this equality works because it's comparing a symbol to a list of symbols?

tests/testthat/_snaps/verb-filter.md
```diff
@@ -28,3 +28,85 @@
...
+      Windowed expression `SUM(`x_mean`)` does not have explicit order.
+      i Please use `arrange()` or `window_order()` to make deterministic.
+    Output
+      <SQL>
+      SELECT `g`, `h`, `x_mean`
+      FROM (
+        SELECT
+          `g`,
+          `h`,
+          `x_mean`,
```
hadley, line 103: Unrelated to this PR, have we considered continuing to use `*` when you only add new columns?

hadley (commented):
This is looking great!

tests/testthat/_snaps/verb-filter.md
mgirlich: No, we haven't. Good point, that would be quite nice.

R/tidyeval.R
mgirlich: Yes, it is comparing a symbol to a list of symbols and it is indeed a bit surprising. But it works nicely:

```r
sym("b") == syms(c("a", "b", "c"))
#> FALSE  TRUE FALSE
```

To make this less surprising it probably makes sense to just use

```r
replace[[match(as_string(call), sym)]]
```

tests/testthat/_snaps/verb-filter.md
mgirlich: I added PR #881 to tackle this issue.

R/tidyeval.R
hadley: Yeah, I'd prefer that.

PR #881, opened 2022-05-20
Use `*` in `SELECT` whenever possible

mgirlich:
This PR simplifies the `SELECT <new columns>, <prev. columns>, <new columns>` to `SELECT <new columns>, *, <new columns>`, e.g.

```r
lf <- lazy_frame(x = 1, y = 1)
lf %>% transmute(a = 1L, x, y, z = 2L)
```

Before

```sql
SELECT 1 AS `a`, `x`, `y`, 2 AS `z`
FROM `df`
```

After

```sql
SELECT 1 AS `a`, *, 2 AS `z`
FROM `df`
```

It also uses `*` in some more situations because it simplifies symbol quosures to symbols in `get_mutate_layerys()`

```r
lf %>% distinct(x, y)
```

Before

```sql
SELECT DISTINCT `x`, `y`
FROM `df`
```

After

```sql
SELECT DISTINCT *
FROM `df`
```


hadley (approved):
Nice!

hadley:
Do you think we could squeeze this in 2.2.0? This example is particularly nice for the blog post:

```R
lf <- lazy_frame(!!!setNames(as.list(1:26), letters))
lf |> 
  mutate(x2 = x + 1, x2 = x2 + 1)
```

Or do you think this is likely to cause more revdep failures?

mgirlich:
👍  I don't think there will be more revdep failures. So, yeah, let's get this into 2.2.0.

hadley:
Thanks!

apalacio9502:
Hello,

This modification is very interesting, however in the case of Oracle it is not fully functional.

In Oracle, if I want to select all the columns and add a new one, I must write the query in the following way 
`SELECT q01.*, COLUM_1+COLUM_2 AS VALUE
FROM TEST q01;`

And not as dbplyr is currently doing, which is this way.

`SELECT *, COLUM_1+COLUM_2 AS VALUE
FROM TEST q01;`

The big difference is that the * is preceded by the name that the table is being called, which in this case is q01

Thank you.

hadley:
@apalacio9502 can you please file a new issue instead of commenting on a closed one?

PR #890, opened 2022-05-23
Use `tbl()` in `tbl_src_dbi()`

### LINKED ISSUE #889: `collapse()` and `compute()` do not use `tbl()` dispatch anymore
PR #668 was actually a breaking change. Previously, `compute()`, `collapse()` and `copy_to()` used the generic `tbl()`. Now, they use `tbl_src_dbi()` which is not a generic. The sparklyr package has defined its own method for `tbl()`. So, this change broke their `compute()` function.

mgirlich:
Fixes #889.
It still makes sense to use `tbl_src_dbi()` to make sure the `vars` argument is provided but we should keep the dispatch via `tbl()`.

### DIFF
--- a/R/src_dbi.R
+++ b/R/src_dbi.R
@@ -89,8 +89,8 @@ tbl.src_dbi <- function(src, from, ...) {
 # The former may query the database for column names if `vars` is omitted,
 # the latter always requires `vars`.
 tbl_src_dbi <- function(src, from, vars) {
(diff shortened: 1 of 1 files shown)


hadley (approved):

krlmlr (approved):
Thanks, sorry for the mess.

mgirlich:
@krlmlr Well, you created a PR and I also didn't see it coming. But easy enough to fix 😄

PR #892, opened 2022-05-24
Only use table alias in joins if necessary

mgirlich:
A bit more work than expected due to some edge cases:

1. `left_join(x, x)` still need a table alias. The question is whether they should both use an alias or only one of the tables?
2. `left_join(x, y, sql_on = ...)` should still use the table alias `LHS` and `RHS` to not be a breaking change

In some cases the error message might not be helpful but I think this is mostly in edge cases so I didn't want to put in too much effort for now.

```r

lf <- lazy_frame(x = 1, a = 1, .name = "df1")
lf2 <- lazy_frame(x = 1, b = 2, .name = "df2")

left_join(lf, lf2, by = "x")
```

Before

```sql
FROM `df1` AS `LHS`
LEFT JOIN `df2` AS `RHS`
  ON (`LHS`.`x` = `RHS`.`x`)
```

After

```sql
SELECT `df1`.`x` AS `x`, `a`, `b`
FROM `df1`
LEFT JOIN `df2`
  ON (`df1`.`x` = `df2`.`x`)
```

```r
lf %>% 
  filter(a == 1) %>% 
  left_join(lf2, by = "x")
```

Before

```sql
FROM (
  SELECT *
  FROM `df1`
  WHERE (`a` = 1.0)
) `LHS`
LEFT JOIN `df2` AS `RHS`
  ON (`LHS`.`x` = `RHS`.`x`)
```

After

```sql
FROM (
  SELECT *
  FROM `df1`
  WHERE (`a` = 1.0)
) `LHS`
LEFT JOIN `df2`
  ON (`LHS`.`x` = `df2`.`x`)
```


R/verb-joins.R
```diff
@@ -359,12 +355,36 @@ add_semi_join <- function(x, y, anti = FALSE, by = NULL, sql_on = NULL, copy = F
...
-check_join_as <- function(x_as, y_as, call) {
-  vctrs::vec_assert(x_as, character(), size = 1, arg = "x_as", call = call)
-  vctrs::vec_assert(y_as, character(), size = 1, arg = "y_as", call = call)
-  if (x_as == y_as) {
+check_join_as <- function(x_as, x, y_as, y, sql_on, call) {
+  x_name <- query_name(x)
+  y_name <- query_name(y)
+  if (is_null(x_as) && is_null(y_as)) {
+    if (identical(x_name, y_name)) {
+      return(list(x_as = ident("LHS"), y_as = ident("RHS")))
```
hadley, line 363: Another option would be `paste(x_name, "_RHS")` + `paste(x_name, "_LHS")`

R/remote.R
```diff
@@ -37,6 +37,54 @@ remote_name <- function(x) {
...
+  NULL
+}
+
+#' @export
+query_name.lazy_select_query <- function(x) {
+  if (!is_empty(x$where)) {
+    return()
+  }
+
+  if (!is_empty(x$order_by)) {
```
hadley, line 70: Do you mind briefly explaining the reasoning in a comment? This is detecting if it will need to become a subquery?

hadley: Actually, doesn't this function always return `NULL`?

mgirlich: I can't recall why I added this method 😕 Maybe it also got obsolete after inlining `select()`. So, I just removed it 😄

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # dbplyr (development version)
 
+* Joins now don't use the table alias "LHS" and "RHS" unless necessary (@mgirlich).
+
+* `fill()` now works for connections from RPostgreSQL (@mgirlich).
+
+* `explain()` now works for connections from ROracle (@mgirlich).
```
hadley, line 7: Remnant of merge conflict?

hadley (approved):

NEWS.md
mgirlich: Probably...

PR #900, opened 2022-05-27
Add more detail to rows_insert() docs

krlmlr:
Follow-up to #887.

### DIFF
--- a/R/rows.R
+++ b/R/rows.R
@@ -1,4 +1,4 @@
-#' Manipulate individual rows, optionally modifying the underlying table.
+#' Manipulate individual rows on the database
 #'
 #' @description
 #' These are methods for the dplyr [rows_insert()], [`rows_append()`],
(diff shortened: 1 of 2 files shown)


R/rows.R
```diff
@@ -8,7 +8,17 @@
...
+#' the database: the connection needs permission to insert, modify or delete
+#' rows, but not to alter the structure of the table.
+#'
+#' The default, `in_place = FALSE`, generates equivalent lazy tables (using
+#' `SELECT` queries) that allow previewing the result without actually
+#' modifying the underlying table on the database.
+#'
+#' @details
+#' For `in_place = TRUE`, the `x` argument must refer to a table instantiated
+#' with [tbl()], not to a lazy query. The [remote_name()] function is used
```
mgirlich, line 20: I think this is fine though not 100% correct:
* one can do something like `tbl(con, "SELECT ... FROM ...")` which does not work (I just noticed `remote_name()` gives a name for such a table but it is quite an edge case anyway...)
* it also works after `compute()`

But I think it is clear enough as it is.

R/rows.R
```diff
@@ -1,4 +1,4 @@
-#' Manipulate individual rows, optionally modifying the underlying table.
+#' Manipulate individual rows on the database
```
hadley, line 1: I think it's really important to include "modifying the underlying" table in the title, since that makes these functions different to every other dbplyr function.

R/rows.R
```diff
@@ -8,7 +8,17 @@
...
+#' rows, but not to alter the structure of the table.
+#'
+#' The default, `in_place = FALSE`, generates equivalent lazy tables (using
+#' `SELECT` queries) that allow previewing the result without actually
+#' modifying the underlying table on the database.
+#'
+#' @details
+#' For `in_place = TRUE`, the `x` argument must refer to a table instantiated
+#' with [tbl()] or [compute()], not to a lazy query. The [remote_name()]
+#' function is used to determine the name of the table to be updated.
```
hadley, line 21: Maybe move this into `@param x,y`? I think you could just say something like "In most cases, `x`, will be a table, not a lazy query". Since the point of `in_place = FALSE` is just to preview the change.

hadley (approved):

hadley (approved):

PR #901, opened 2022-05-27
Nicer copy inline query

### LINKED ISSUE #893: Simplify `copy_inline()` using column aliases
eg. 
```sql
SELECT CAST(`x` AS INTEGER) AS `x`, CAST(`y` AS TEXT) AS `y`
FROM 
  (VALUES (1, 'a'), (2, 'b'), (3, 'c'), (4, 'd'), (5, 'e')) AS inline(x, y)
```

At least duckdb and postgresl support this (although SQLite does not). Postgres calls it a "column alias list", the [SQL standard](http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt) calls it "derived column list".

mgirlich:
Fixes #893.

This now produces

```sql
SELECT CAST(`x` AS INTEGER) AS `x`, CAST(`y` AS TEXT) AS `y`
FROM (  VALUES (1, 'a'), (2, 'b')) AS inline(`x`, `y`)
```

(the extra white spaces are a bit annoying but not really worth to remove I think)

and

```sql
SELECT CAST(`x` AS INTEGER) AS `x`, CAST(`y` AS TEXT) AS `y`
FROM (
  VALUES
    (1, 'a'),
    (2, 'b'),
    (3, 'c'),
    (4, 'd'),
    (5, 'e'),
    (6, 'f'),
    (7, 'g'),
    (8, 'h')
) AS inline(`x`, `y`)
```

We do not have a DuckDB backend yet. Instead, DuckDB implemented their own dbplyr backend (though not yet on CRAN). So, I did not add the nicer looking SQL for DuckDB.


R/verb-copy-to.R
```diff
@@ -249,21 +224,64 @@ sql_values_clause <- function(con, df, row = FALSE, derived = FALSE, lvl = 0) {
...
+  rows_sql <- sql(paste0(if (row) "ROW", "(", rows, ")"))
+
+  list(sql_clause("VALUES", rows_sql))
+}
+
+sql_values_zero_rows <- function(con, df, lvl) {
+  if (nrow(df) != 0L) {
+    # TODO use `cli_abort()` after https://github.com/r-lib/rlang/issues/1386
+    # is fixed
+    abort("`df` does not have 0 rows", .internal = TRUE)
```
hadley, line 247:
```suggestion
    abort("{.arg df} must have 0 rows", .internal = TRUE)
```

hadley (approved):

R/backend-sqlite.R
```diff
@@ -179,6 +179,12 @@ sql_query_join.SQLiteConnection <- function(con, x, y, vars, type = "inner", by
   }
 }
 
+#' @export
+values_prepare.SQLiteConnection <- function(con, df) {
```
krlmlr, line 183: Is this also applicable with the new date+time extension via `extended_types = TRUE` ? Is there a way to find out if this has been passed to the connection?

https://rsqlite.r-dbi.org/reference/sqlite

R/verb-copy-to.R
```diff
@@ -166,40 +166,20 @@ op_vars.lazy_values_query <- function(op) {
...
 
-#' @export
-sql_values.SQLiteConnection <- function(con, df, lvl = 0, ...) {
-  needs_escape <- purrr::map_lgl(df, ~ is(.x, "Date") || inherits(.x, "POSIXct"))
-  purrr::modify_if(df, needs_escape, ~ escape(.x, con = con, parens = FALSE, collapse = NULL)) %>%
-    sql_values_clause(con = con, lvl = lvl)
-}
-
-sql_values_clause <- function(con, df, row = FALSE, derived = FALSE, lvl = 0) {
+sql_values_subquery_default <- function(con, df, lvl, row, derived) {
```
krlmlr, line 179: There are valid reasons to avoid exporting this machinery. Should we make it as easy as possible for third-party packages to vendor this code, e.g. by putting it into a separate file and mentioning a "do whatever you want" license?

krlmlr (approved):

R/backend-sqlite.R
mgirlich: I didn't manage to create a date column without actually creating a new table...

``` r
con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:", extended_types = TRUE)

# this does not produce columns of type DATE
out <- DBI::dbGetQuery(
  con,
  "SELECT
  date(`date`) AS `date`,
  CAST(`date` AS DATE) AS `date2`
FROM (
  SELECT
    NULL AS `date`
  WHERE (0 = 1)
  UNION ALL
  VALUES ('2020-01-01')
) AS `values_table`"
)

tibble::as_tibble(out)
#> # A tibble: 1 × 2
#>   date       date2
#>   <chr>      <int>
#> 1 2020-01-01  2020

# this works nicely
DBI::dbExecute(con, "create table tbl2 (aDate DATE)")
#> [1] 0
tibble::as_tibble(DBI::dbReadTable(con, "tbl2"))
#> # A tibble: 0 × 1
#> # … with 1 variable: aDate <date>
```

<sup>Created on 2022-05-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/verb-copy-to.R
mgirlich: Rather not yet. It uses another unexported function `sql_format_clauses()`. So, we would also have to export this function. But then we should also export `sql_clause()` and maybe some others. Would be nice if we could keep them internal for now.

krlmlr: I'm all for keeping them internal, just collect in one place so that implementers can grab them. Not important though.

R/backend-sqlite.R
krlmlr: Oh yeah, I'm not sure this will ever work. But then `copy_inline()` isn't necessarily faster or better on SQLite, so we could forward to `copy_to()`?

mgirlich: Hmm, as far as I can see there are three reasons to use `copy_inline()`:

1. No permission to create a table -> SQLite does not have a concept of permissions apart from file permissions. So, this might not be relevant.
2. Performance -> probably not relevant for SQLite but one might test this.
3. Fully understandable queries in the sense that one can actually see in the query all the values used to create a table. This can sometimes be nice for tables with only a couple of rows.

The first two are probably irrelevant for SQLite. The third point can sometimes be nice.
But then it is a bit confusing to use `copy_inline()` and then not ending up with a `VALUES` clause. So, I'm not really sure this is a good idea.

PR #906, opened 2022-05-31
Mark sd(), var(), cor(), and cov() as not supported

hadley:
These are common R summary functions, but are not standard SQL.


R/backend-.R
```diff
@@ -394,11 +402,18 @@ base_win <- sql_translator(
   # Recycled aggregate fuctions take single argument, don't need order and
   # include entire partition in frame.
   mean  = win_aggregate("AVG"),
-  var   = win_aggregate("VARIANCE"),
   sum   = win_aggregate("SUM"),
   min   = win_aggregate("MIN"),
   max   = win_aggregate("MAX"),
 
+  all   = win_aggregate("MAX"),
```
mgirlich, line 409: I think you switched `all` and `any` 😄 

```r
memdb_frame(x = c(TRUE, FALSE), y = TRUE) %>% 
  mutate(
    across(.fns = c(all = all, any = any))
  )
```

```
# Source:   SQL [2 x 6]
# Database: sqlite 3.38.5 [:memory:]
      x     y x_all x_any y_all y_any
  <int> <int> <int> <int> <int> <int>
1     1     1     1     0     1     1
2     0     1     1     0     1     1
```

hadley: Oops yes. I will add a test 😄

krlmlr (approved):

Issue #822, opened 2022-04-08
Improve error messages


mgirlich:
## Missing Variables

The error message for missing variables is generated by the database. Therefore, the error message does not mention the verb where the issue is.

``` r
library(dbplyr)
library(dplyr, warn.conflicts = FALSE)

df <- memdb_frame(x = 1)

df %>% 
  mutate(x = y + 1)
#> Error: no such column: y

df %>% 
  group_by(x = y + 1)
#> Error: no such column: y

df %>% 
  distinct(x = y + 1)
#> Error: no such column: y
```

<sup>Created on 2022-04-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

## Functions without translation

The error message also does not mention the verb where the issue comes from:

``` r
library(dbplyr)
library(dplyr, warn.conflicts = FALSE)

df <- memdb_frame(x = 1)

df %>% 
  mutate(x = y %/% 1)
#> Error: %/% is not available in this SQL variant
```

<sup>Created on 2022-04-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

## `across()`

``` r
library(dbplyr)
library(dplyr, warn.conflicts = FALSE)

df <- memdb_frame(x = 1)

df %>% 
  mutate(across(c(x), "a"))
#> Error: no such function: a
```

<sup>Created on 2022-04-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

## How to improve

* translate to SQL directly when calling the verb
* do not translate unknown symbols
* Consider using `cli_abort()`

hadley:
I think we should switch to `cli::cli_abort()` for this, and then the problem mostly becomes about tracking which top-level verb is responsible for the error. I suspect we'll need to use a global-state hack for this, as otherwise we'll have to thread the `call` variable through a very large number of functions.

PR #907, opened 2022-06-01
Improved error messages

### LINKED ISSUE #803: Bug: can use `distinct()` on removed variable
The issue exists in the CRAN and the dev version of `dbplyr`.

``` r
library(dbplyr)

lazy_frame(x = 1, y = 1) %>% 
  select(-x) %>% 
  distinct(x)
#> <SQL>
#> SELECT DISTINCT `x`
#> FROM `df`
```

<sup>Created on 2022-03-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

The issue is in `partial_eval_sym()`:

```r
partial_eval_sym <- function(sym, data, env) {
  vars <- op_vars(data)
  name <- as_string(sym)
  if (name %in% vars) {
    sym
  } else if (env_has(env, name, inherit = TRUE)) {
    eval_bare(sym, env)
  } else { # this else clause is the issue
    sym
  }
}
```

Currently, this is needed in three places:

* `across()`: this calls `partial_eval_call()` where we might use `.x` or function arguments.
* `mutate()` with references to previously created variables (this won't be necessary after merging #802).
* one test for `partial_eval()` where this behaviour is explicitly tested ("existing non-variables get inlined"). But this test was only added recently when aligning `partial_eval()` with `dtplyr`, so I think it can be removed.

So, overall I think this can be removed and the evaluation can be made a bit safer.

### LINKED ISSUE #905: Should strict_sql also error on unknown variables?
``` r
library(dbplyr)

options(dplyr.strict_sql = TRUE)

lazy_frame(x = 1) |> 
  mutate(z = foo(x))
#> Error in `foo()`:
#> ! Don't know how to translate `foo()`
```

<sup>Created on 2022-05-31 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

But 

``` r
library(dbplyr)

options(dplyr.strict_sql = TRUE)

lazy_frame(x = 1) |> 
  mutate(z = x + y)
#> <SQL>
#> SELECT *, `x` + `y` AS `z`
#> FROM `df`
```

<sup>Created on 2022-05-31 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mgirlich:
This is the first step to much more helpful error messages.

This fixes #803, fixes #905, and parts of #822.

* Unknown variables now produce an error. I don't think there are many (or even any?) reasons that this should not produce an error.
* `fix_call()` is currently needed because tidyselect/vctrs do not always handle the call correctly.


tests/testthat/test-tidyeval.R
```diff
@@ -19,7 +19,6 @@ test_that("existing non-variables get inlined", {
 
   n <- 10
   expect_equal(capture_dot(lf, x + n), expr(x + 10))
-  expect_equal(capture_dot(lf, x + m), expr(x + m))
```
mgirlich, line 22: I don't think this test made much sense.

tests/testthat/test-verb-summarise.R
```diff
@@ -17,6 +17,7 @@ test_that("summarise performs partial evaluation", {
 })
 
 test_that("can't refer to freshly created variables", {
+  skip("TODO need to adapt error")
```
mgirlich, line 20: I will have to think how to keep the error message...

R/backend-hana.R
```diff
@@ -20,7 +20,7 @@
 #' library(dplyr, warn.conflicts = FALSE)
 #'
 #' lf <- lazy_frame(a = TRUE, b = 1, c = 2, d = "z", con = simulate_hana())
-#' lf %>% transmute(x = paste0(z, " times"))
+#' lf %>% transmute(x = paste0(d, " times"))
```
hadley, line 23: All of these minor fixes definitely imply that the old behaviour was silently obscuring a lot of problems.

R/rows.R
```diff
@@ -724,9 +724,9 @@ tick <- function(x) {
 
 # other helpers -----------------------------------------------------------
 
-eval_select2 <- function(expr, data) {
+eval_select2 <- function(expr, data, error_call) {
   sim_data <- simulate_vars(data)
-  locs <- tidyselect::eval_select(expr, sim_data)
+  locs <- fix_call(tidyselect::eval_select(expr, sim_data), error_call)
```
hadley, line 729: Are you deliberately not using the `error_call` argument to `eval_select()`?

R/tidyeval-across.R
```diff
@@ -60,7 +60,10 @@ across_funs <- function(funs, env, data, dots, names_spec, fn, evaluated = FALSE
     funs <- eval(funs, env)
     return(across_funs(funs, env, data = data, dots = dots, names_spec = NULL, fn = fn, evaluated = TRUE))
   } else {
-    cli_abort("{.arg .fns} argument to {.fun dbplyr::across} must be a NULL, a function, formula, or list")
+    cli_abort(
+      "{.arg .fns} argument to {.fun dbplyr::across} must be a NULL, a function, formula, or list",
+      call = NULL
```
hadley, line 65: For consistency, you could maybe try call = `quote(dbplyr::across())` and then remove from the message

R/tidyeval-across.R
```diff
@@ -109,21 +116,21 @@ partial_eval_fun <- function(fun, env, data) {
 }
 
 partial_eval_body <- function(x, env, data, sym, replace = quote(!!.x)) {
-  call <- replace_sym(x, sym, replace)
-  if (is_call(call)) {
```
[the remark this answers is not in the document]
hadley: Then maybe we can remove `partial_eval_body()` altogether?

R/tidyeval.R
```diff
@@ -117,7 +131,10 @@ partial_eval_sym <- function(sym, data, env) {
   } else if (env_has(env, name, inherit = TRUE)) {
     eval_bare(sym, env)
   } else {
-    sym
+    cli::cli_abort(
+      "object {.var {name}} not found.",
```
hadley, line 135:
```suggestion
      "Object {.var {name}} not found.",
```

R/tidyeval-across.R
```diff
@@ -132,16 +139,28 @@ across_setup <- function(data,
...
-  names_spec <- eval(call$.names, env)
-  funs_across_data <- across_funs(
-    funs = call$.fns,
-    env = env,
-    data = data,
-    dots = dots,
-    names_spec = names_spec,
-    fn = fn
-  )
+  try_fetch({
```
hadley, line 142: It'd be slightly nicer to flip the `lapply()` and the `try_fetch()` so that the error message could reference the specific element of `...` that failed. (If you did this, for simplicity, I'd suggest coverting the `lapply()` to a for loop)

tests/testthat/_snaps/verb-group_by.md
```diff
@@ -12,7 +12,28 @@
       (expect_error(lazy_frame(x = 1:3) %>% group_by(y)))
     Output
       <error/rlang_error>
-      Error in `group_by.tbl_lazy()`:
```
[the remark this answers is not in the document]
hadley: No this seems fine to me.

tests/testthat/test-tidyeval.R
hadley: Agreed. I'm not sure why I made it work this way. The only reason I can think of is to make it possible to refer to database specific constants, but that seems like a pretty weak argument.

hadley (approved):

R/tidyeval-across.R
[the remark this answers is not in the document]
mgirlich: 👍

R/verb-summarise.R
```diff
@@ -46,21 +45,55 @@ summarise.tbl_lazy <- function(.data, ..., .groups = NULL) {
...
+
+summarise_bind_error <- function(cur_data, dots, i, error_env) {
+  quos <- dots[[i]]
+  if (!is.list(quos)) {
+    quos <- set_names(list(quos), names(dots)[[i]])
+  }
+
+  for (j in seq_along(quos)) {
+    dot_name <- names(quos)[[j]]
+    summarise_bind_error1(error_env, dot_name)
```
mgirlich, line 81: It is a bit of a hack but works nicely.
When the next dot is evaluated `partial_eval_sym()` searches for the variable in `cur_data`. As we just removed it, it is evaluated in the environment of the dot. By lazily binding an error to the variable in the environment we get the old behaviour - even slightly improved.

R/tidyeval-across.R
mgirlich: > then remove from the message

I'm not quite sure how I could remove it.

hadley: I just meant that you no longer need `to {.fun dbplyr::across}` in the message.

PR #917, opened 2022-06-20
Do not use `*` in `SELECT` queries for Oracle

### LINKED ISSUE #908: The new version is not compatible with Oracle
Hello,

The modification in this commit https://github.com/tidyverse/dbplyr/commit/8dd7639cc84759b6c30c88f5661e97a225a2637f  is very interesting, however in the case of Oracle it is not fully functional.

In Oracle, if I want to select all the columns and add a new one, I must write the query in the following way
SELECT q01.*, COLUM_1+COLUM_2 AS VALUE FROM TEST q01;

And not as dbplyr is currently doing, which is this way.

SELECT *, COLUM_1+COLUM_2 AS VALUE FROM TEST q01;

The big difference is that the * is preceded by the name that the table is being called, which in this case is q01

Thank you.

mgirlich:
Fixes #908.


mgirlich:
I think it would be great to have a bugfix release of dbplyr for this issue and #916. What do you think?

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # dbplyr (development version)
 
+* Querying Oracle databases works again. Unfortunately, the fix caused to
+  explicitly select every column again for Oracle databases (@mgirlich, #908).
```
hadley, line 4:
```suggestion
* Querying Oracle databases works again. Unfortunately, the fix requires every
  column to be explicitly selected again (@mgirlich, #908).
```

hadley (approved):
Do you want to take the lead on the release? You'd just work your way through `usethis::use_release_issue()` until you get to submission, when I'd take over.

PR #918, opened 2022-06-20
Create `src_` classes from DBI class extensions

noamross:
When a `src` is created from a DBI-compliant connection, it gets the S3 class `src_TYPEConnection`.  However, a DBI connection can extend, and thus inherit, from other DBI classes and thus the `src_` type should reflect these.  In my case, my [`doltr` package](https://github.com/ecohealthalliance/doltr) connects to Dolt databases, which are MySQL/MariaDB compatible and thus `DoltConnection` extends `MariaDBConnection`.  The `dm` package [queries the class of `src_` objects to determine database type](https://github.com/cynkra/dm/blob/main/R/db-helpers.R#L67-L70), but since `dplyr` doesn't include class extensions when creating `src_` types, this compatibility is invisible.

This PR changes the construction of `src_dbi` objects so they inherit all the types of DBI connections they extend.

@krlmlr (refers to https://github.com/cynkra/dm/issues/366)


R/src_dbi.R
```diff
@@ -123,7 +123,9 @@ src_dbi <- function(con, auto_disconnect = FALSE) {
     disco <- db_disconnector(con, quiet = is_true(auto_disconnect)) # nocov
   }
 
-  subclass <- paste0("src_", class(con)[[1]])
+  con_classes <- extends(class(con))[
+    grepl("(?<!DBI)Connection$", extends(class(con)), perl = TRUE)]
```
krlmlr, line 127: Did you mean to exclude only `DBIConnection` ?

```suggestion
    grepl("(?<!^DBI)Connection$", extends(class(con)), perl = TRUE)]
```

krlmlr (approved):

Issue #627, opened 2021-02-11
Translate `window_order` to `arrange` for local data frames


iangow:
Feature request: Add `window_order` to `dplyr` for local data frames. 

Following on from tidyverse/dplyr#2874, it seems that using `arrange()` to produce windows using SQL sources yields a warning like this:

```
## Warning: ORDER BY is ignored in subqueries without LIMIT
## ℹ Do you need to move arrange() later in the pipeline or use window_order() instead?
```

This is fine, as either I can just swap out `arrange` for `window_order` and the code runs or I can select `warning=FALSE` in the code chunk options.

There are two downsides to the former approach:

1. Because `window_order` lives in `dbplyr`, its use requires either `dbplyr::window_order` or `library(dbplyr)`.
2. The code becomes adapted to SQL sources, as it will no longer work with local data frames. (I find myself often inserting or removing a `collect()` call as I mess around with data, as **if one is careful and uses standard `dplyr` verbs**, code works in either place. This is a really nice feature.)

If `window_order` existed for local data frames, then both problems would be solved. And it seems it has a well-defined meaning in that new context.

iangow:
> Maybe dbplyr could allow `window_order()` to work on data frames? (Just be calling `arrange`).

I think that would work for my use case. I will edit the issue title accordingly.

> It doesn't feel right to me to expose this in dplyr because it's so database focused. 

*In principle*, there are benefits from being able to run window functions for particular variables without sorting the whole data frame. That said, I couldn't find any compelling examples of this *in practice* (my sample may be too limited; I couldn't find the examples I vaguely remember from many years ago).

hadley:
Right, but to make `window_order()` work like that would be a bunch of implementation work in dplyr, which definitely wouldn't be worth it.

hadley:
Do you want to do a PR for this?  You'd need to turn `window_order()` into an S3 generic, then add `tbl_lazy()` and `data.frame()` methods.

iangow:
Let me see what I can do. Probably in a month or so.

(I'm up to p.61 of "Advanced R" … it seems another 240 pages of reading will allow me to make sense of what you're saying. I made my [first R package](https://github.com/iangow/farr) today---consisting of nothing but the #628 function---so there's hope.)

hadley:
@iangow I can also do it 😄

mgirlich:
@iangow Do you want to have a go at a PR now? 😄

iangow:
[...]

PR #926, opened 2022-06-27
Turned window_order into S3 generic.

iangow:
- Imported from dplyr.
- Relates to #627.


iangow:
> Thanks for your PR. By now there is a check in place whether `.data` in `window_order()` is a lazy table or not. So, there is no immediate need to turn `window_order()` into an S3 generic.

Thanks. Does this mean one can use `window_order()` with non-lazy tables?

mgirlich:
No, but it gives a nicer error message for non-lazy tables. Hadley pointed out this should rather throw an error https://github.com/tidyverse/dplyr/issues/6146#issuecomment-1192825426.

mgirlich:
It depends on the circumstances whether `arrange()` creates a warning or not. As there is no subquery generated in the example it would not make sense to create a warning. In this case it is fine to use `arrange()` as the result is also supposed to be arranged. Overall, it would probably make more sense to call `window_order()` before the `mutate()` and `arrange()` as last step.

iangow:
Got it. The example below (same as above, but with a join to trigger a subquery) triggers the warning.

There are a lot of subtleties with `dbplyr` that I think are challenging for a single chapter in R4DS to get into (perhaps why they don't even discuss `window_order()` there).

To this end, I have done some work "translating" the O'Reilly book "SQL for Data Analysis" to `dbplyr` [here](https://iangow.github.io/sql_book). Using `dbplyr` means the reader can make all the plots included in the original book (these were done "behind the scenes" using Python or Tableau in the original book). Also I conjecture that many users would write more accurate SQL using `dbplyr` rather than SQL (see [here](https://github.com/iangow/notes/blob/main/ctes.qmd)).

``` r
library(DBI)
library(dbplyr)
library(tidyverse)

con <- DBI::dbConnect(duckdb::duckdb())

dbplyr::copy_nycflights13(con)
#> Creating table: airlines
#> Creating table: airports
#> Creating table: flights
#> Creating table: planes
#> Creating table: weather
flights <- tbl(con, "flights")
airlines <- tbl(con, "airlines")

flights |> 
  group_by(dest) |>  
  window_order(time_hour) |> 
  mutate(lead = lead(arr_delay),
         lag = lag(arr_delay)) |>
  inner_join(airlines, by = "carrier") |>
  show_query()
#> <SQL>
#> SELECT LHS.*, "name"
#> FROM (
#>   SELECT
#>     *,
#>     LEAD(arr_delay, 1, NULL) OVER (PARTITION BY dest ORDER BY time_hour) AS lead,
#>     LAG(arr_delay, 1, NULL) OVER (PARTITION BY dest ORDER BY time_hour) AS lag
#>   FROM flights
#> ) LHS
#> INNER JOIN airlines
#>   ON (LHS.carrier = airlines.carrier)

flights |> 
  group_by(dest) |>  
  arrange(time_hour) |> 
  mutate(lead = lead(arr_delay),
         lag = lag(arr_delay)) |>
  ungroup() |>
  inner_join(airlines, by = "carrier") |>
  show_query()
#> <SQL>
#> SELECT LHS.*, "name"
#> FROM (
#>   SELECT
#>     *,
#>     LEAD(arr_delay, 1, NULL) OVER (PARTITION BY dest ORDER BY time_hour) AS lead,
#>     LAG(arr_delay, 1, NULL) OVER (PARTITION BY dest ORDER BY time_hour) AS lag
#>   FROM flights
#> ) LHS
#> INNER JOIN airlines
#>   ON (LHS.carrier = airlines.carrier)
```

<sup>Created on 2023-06-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #930, opened 2022-07-04
WIP: Support new dbplyr.trace option

### LINKED ISSUE #827: Debug mode
I'd like to propose a mode where all SQL queries and statements sent to the database are logged, perhaps using classed messages. We have {dblog}, but that log is too verbose -- during analysis we're mostly interested in the generated SQL. This debug mode could perhaps also automatically be enabled during tests, so that we get SQL code for each query we run, for free.

krlmlr:
Print SQL before querying or executing, and a corresponding message when done.

Closes #827.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -132,6 +132,7 @@ Collate:
     'testthat.R'
     'tidyeval-across.R'
     'tidyeval.R'
+    'trace.R'
     'translate-sql.R'
     'utils-format.R'
     'verb-arrange.R'
--- a/R/trace.R
+++ b/R/trace.R
@@ -1,0 +1,58 @@
+tracing_level <- function() {
+  trace <- getOption("dbplyr.trace")
+  if (is.null(trace)) {
+    return(0L)
+  }
+  if (is.logical(trace)) {
+    return(as.integer(trace))
+  }
+  if (!is_integerish(trace)) {
+    warn("Invalid value for dbplyr.trace option, resetting.")
+    options(dbplyr.trace = NULL)
+    return(0)
+  }
+  as.integer(trace)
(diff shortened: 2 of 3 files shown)


krlmlr:
What do you want this to look like?

mgirlich:
I'm basically quite happy with the approach already. As this was your idea: is there anything else you want/need?

What's missing from this PR in my opinion is:

* NEWS
* It might make sense to add a better visual indicator between queries. In the following example it is a bit too difficult to quickly distinguish the two queries:

```
[17]: dbExecute()
ANALYZE `dbplyr_011`
[17]: dbExecute() done
[18]: dbExecute()
INSERT INTO `dbplyr_010` (`x`, `y`)
SELECT *
FROM (
  SELECT *
  FROM `dbplyr_011`
) AS `...y`
WHERE NOT EXISTS (
  SELECT 1 FROM `dbplyr_010`
  WHERE (`dbplyr_010`.`x` = `...y`.`x`)
)
[18]: dbExecute() done
```

hadley:
Closing, since if you're still interested, you'd now want to implement on top of `db_execute()`/`db_get_query()`.

krlmlr:
I'm confused. Are `db_execute()` and `db_get_query()` becoming generics? They don't seem to be in 267a12e080806482f85be5a73b5fb5372b14af1f .

hadley:
Why would they need to be generics?

krlmlr:
If they are not, then this may mean that you're asking me to reimplement this PR using new idioms. I'd be happy to review a PR instead.

hadley:
I'm not sure this feature is necessary any more, now that errors will include the generated SQL.

Issue #897, opened 2022-05-25
Reconsider header


hadley:
Currently the header looks something like:

```R
# Source:   SQL [?? x 5]
# Database: duckdb_connection
```

where a tibble looks like:

```R
# A tibble: 53,940 × 10
```

The idea of a "source" is mostly superseded, so it might be better to display something like:

```R
# A duckdb_connection query: ?? × 5
```

PR #936, opened 2022-07-07
Use the `tbl_sum()` generic for printing the header

krlmlr:
For https://github.com/cynkra/dm/issues/1187. We could take a look at #897 as part of it too.

The current implementation never forwards to `pillar::tbl_sum()`, which makes it difficult to implement mixins that contribute information to the header.


mgirlich:
Thanks for the PR and sorry for the long wait. Not sure whether you still need it for dm but this definitely makes sense to me.

PR #959, opened 2022-08-04
Use `LHS.*` after `left_join()`

### LINKED ISSUE #898: * in joins
Would be nice if:

``` r
library(dbplyr)

lf1 <- lazy_frame(!!!setNames(as.list(1:26), letters))
lf2 <- lazy_frame(a = 1, a1 = 1, a2 = 1)

lf1 |> left_join(lf2)
#> Joining, by = "a"
#> <SQL>
#> SELECT
#>   `LHS`.`a` AS `a`,
#>   `b`,
#>   `c`,
#>   `d`,
#>   `e`,
#>   `f`,
#>   `g`,
#>   `h`,
#>   `i`,
#>   `j`,
#>   `k`,
#>   `l`,
#>   `m`,
#>   `n`,
#>   `o`,
#>   `p`,
#>   `q`,
#>   `r`,
#>   `s`,
#>   `t`,
#>   `u`,
#>   `v`,
#>   `w`,
#>   `x`,
#>   `y`,
#>   `z`,
#>   `a1`,
#>   `a2`
#> FROM `df` AS `LHS`
#> LEFT JOIN `df` AS `RHS`
#>   ON (`LHS`.`a` = `RHS`.`a`)
```

<sup>Created on 2022-05-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Instead produced:

```sql
SELECT
  `LHS`.*, `a1`, `a2`
FROM `df` AS `LHS`
LEFT JOIN `df` AS `RHS`
  ON (`LHS`.`a` = `RHS`.`a`)
```

No idea how feasible this is.

mgirlich:
Closes #898. There should be no issues with Oracle this time as this uses a table alias.

``` r
library(dbplyr)

lf1 <- lazy_frame(!!!setNames(as.list(1:26), letters))
lf2 <- lazy_frame(a = 1, a1 = 1, a2 = 1)

lf1 |> 
  left_join(lf2)
#> Joining, by = "a"
#> <SQL>
#> SELECT `LHS`.*, `a1`, `a2`
#> FROM `df` AS `LHS`
#> LEFT JOIN `df` AS `RHS`
#>   ON (`LHS`.`a` = `RHS`.`a`)

# and a small bonus thanks to inlining `select()` into joins
lf1 |> 
  left_join(lf2) %>% 
  relocate(a1)
#> Joining, by = "a"
#> <SQL>
#> SELECT `a1`, `LHS`.*, `a2`
#> FROM `df` AS `LHS`
#> LEFT JOIN `df` AS `RHS`
#>   ON (`LHS`.`a` = `RHS`.`a`)
```

<sup>Created on 2022-08-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


R/query-join.R
```diff
@@ -77,6 +79,43 @@ sql_join_var <- function(con, alias, x, y, all_x, all_y, x_as, y_as) {
   }
 }
 
+join_use_star <- function(con,
```
hadley, line 82: Would you mind adding a few comments here to explain the reasoning?

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # dbplyr (development version)
 
+* When possible, dbplyr now uses `SELECT *` after a left join instead of
```
hadley, line 3: Or inner join?

Could we apply the same logic to right joins too?

hadley (approved):
Looks great!

NEWS.md
mgirlich: Not that simple to extend this to all type of joins but now they are all supported (note that for `full_join()` one can never use `*`).

tests/testthat/_snaps/verb-expand.md
```diff
@@ -4,7 +4,7 @@
       lazy_frame(x = 1, y = 1) %>% tidyr::expand(x, y)
     Output
       <SQL>
-      SELECT `x`, `y`
+      SELECT `LHS`.*, `RHS`.*
```
hadley, line 7: If it's easy, maybe we should only use `*` if there is more than one variable selected?

mgirlich: Yes, this looks a bit better now.

hadley: Nice!

PR #962, opened 2022-08-06
Wrap DBI errors

mgirlich:
This PR catches errors produced by the database (i.e. by `dbGetQuery()` and `dbExecute()`) and wraps them to produce more informative errors.

``` r
library(dbplyr)

DBI::dbWriteTable(con, "mtcars", tibble(x = 1, y = 1))
DBI::dbExecute(con, sql_table_index(con, "mtcars", "x", unique = TRUE))
#> [1] 0

rows_append(
  tbl(con, "mtcars"),
  tibble(x = 1),
  copy = TRUE,
  in_place = TRUE
)
# Before
#> Error: UNIQUE constraint failed: mtcars.x

# After
#> Error in `rows_append()`:
#> ! Can't modify database table "mtcars".
#> Caused by error:
#> ! UNIQUE constraint failed: mtcars.x
```

<sup>Created on 2022-08-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/db-sql.R
+++ b/R/db-sql.R
@@ -741,7 +741,7 @@ db_analyze.DBIConnection <- function(con, table, ...) {
   if (is.null(sql)) {
     return() # nocov
   }
-  dbExecute(con, sql)
+  safe_db_execute(con, sql, .msg = "Can't analyze table {.val {table}}.")
 }
 
 dbplyr_create_index <- function(con, ...) {
@@ -752,7 +752,7 @@ dbplyr_create_index <- function(con, ...) {
 db_create_index.DBIConnection <- function(con, table, columns, name = NULL,
                                           unique = FALSE, ...) {
   sql <- sql_table_index(con, table, columns, name = name, unique = unique, ...)
-  dbExecute(con, sql)
+  safe_db_execute(con, sql, .msg = "Can't create index on {.val {table}}.")
 }
 
 dbplyr_explain <- function(con, ...) {
@@ -762,7 +762,9 @@ dbplyr_explain <- function(con, ...) {
 #' @importFrom dplyr db_explain
 db_explain.DBIConnection <- function(con, sql, ...) {
   sql <- sql_query_explain(con, sql, ...)
-  expl <- dbGetQuery(con, sql)
+  call <- current_call()
+  expl <- safe_db_get_query(con, sql, .msg = "Can't explain query.")
+
   out <- utils::capture.output(print(expl))
   paste(out, collapse = "\n")
 }
@@ -774,7 +776,8 @@ dbplyr_query_fields <- function(con, ...) {
 #' @importFrom dplyr db_query_fields
 db_query_fields.DBIConnection <- function(con, sql, ...) {
   sql <- sql_query_fields(con, sql, ...)
-  names(dbGetQuery(con, sql))
+  df <- safe_db_get_query(con, sql, .msg = "Can't query fields.")
+  names(df)
 }
 
 dbplyr_save_query <- function(con, ...) {
@@ -784,7 +787,12 @@ dbplyr_save_query <- function(con, ...) {
 #' @importFrom dplyr db_save_query
 db_save_query.DBIConnection <- function(con, sql, name, temporary = TRUE, ...) {
   sql <- sql_query_save(con, sql, name, temporary = temporary, ...)
-  dbExecute(con, sql, immediate = TRUE)
+  safe_db_execute(
+    con = con,
+    sql = sql,
+    immediate = TRUE,
+    .msg = "Can't save query to {.val {name}}."
+  )
   name
 }
 
@@ -795,4 +803,36 @@ dbplyr_sql_subquery <- function(con, ...) {
 #' @importFrom dplyr sql_subquery
 sql_subquery.DBIConnection <- function(con, from, name = unique_subquery_name(), ..., lvl = 0) {
   sql_query_wrap(con, from = from, name = name, ..., lvl = lvl)
+}
+
+safe_db_get_query <- function(con, sql, ..., .msg, .call = caller_env()) {
+  glue_envir <- caller_env()
+  tryCatch(
+    DBI::dbGetQuery(con, sql, ...),
+    error = function(cnd) {
+      cli_abort(
+        .msg,
+        parent = cnd,
+        .envir = glue_envir,
+        call = .call,
+        class = "dbplyr_error_dbi"
+      )
+    }
+  )
+}
+
+safe_db_execute <- function(con, sql, ..., .msg, .call = caller_env()) {
+  glue_envir <- caller_env()
+  tryCatch(
+    DBI::dbExecute(con, sql, ...),
+    error = function(cnd) {
+      cli_abort(
+        .msg,
+        parent = cnd,
+        .envir = glue_envir,
+        call = .call,
+        class = "dbplyr_error_dbi"
+      )
+    }
+  )
 }
--- a/R/rows.R
+++ b/R/rows.R
@@ -775,12 +775,19 @@ rows_insert_prep <- function(con, x_name, y, by, lvl = 0) {
   out
 }
 
-rows_get_or_execute <- function(x, sql, returning_cols) {
+rows_get_or_execute <- function(x, sql, returning_cols, call = caller_env()) {
   con <- remote_con(x)
+  msg <- "Can't modify database table {.val {remote_name(x)}}."
   if (is_empty(returning_cols)) {
-    dbExecute(con, sql, immediate = TRUE)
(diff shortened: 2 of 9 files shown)


tests/testthat/_snaps/backend-postgres.md
```diff
@@ -102,7 +102,9 @@
       rows_insert(x, y, by = c("a", "b"), in_place = TRUE, conflict = "ignore",
       returning = everything(), method = "on_conflict")
     Condition
-      Error:
+      Error in `rows_insert()`:
+      ! Can't modify database table "df_x".
+      Caused by error:
```
mgirlich, line 107: Should this mention that the error is (probably) produced by the database?

R/db-sql.R
```diff
@@ -796,3 +804,35 @@ dbplyr_sql_subquery <- function(con, ...) {
...
+  glue_envir <- caller_env()
+  tryCatch(
+    DBI::dbGetQuery(con, sql, ...),
+    error = function(cnd) {
+      cli_abort(
+        .msg,
+        parent = cnd,
+        .envir = glue_envir,
+        call = .call,
+        class = "dbplyr_error_dbi"
```
mgirlich, line 818: Not sure whether you want a classed error here or not. I added this for testing but we might want to keep it. What do you think?

mgirlich:
Maybe the `db_*()` family should gain a `.call` argument as they rarely called by the user but rather by the developer. What do you think?

R/db-io.R
```diff
@@ -142,14 +152,14 @@ dbi_quote.ident <- function(x, con) DBI::dbQuoteIdentifier(con, as.character(x))
 dbi_quote.character <- function(x, con) DBI::dbQuoteString(con, x)
 dbi_quote.sql <- function(x, con) DBI::SQL(as.character(x)) # nocov
 
-create_indexes <- function(con, table, indexes = NULL, unique = FALSE, ...) {
```
hadley, line 145: I _think_ you should leave this as is since `.call` is only available in some methods.

R/db-sql.R
hadley: I think our experience so far is that classed errors tend to be more trouble than their worth, so I'd omit it for now.

I'm also not sure if these helper functions should exist — something about the way that `glue_envir` is needed makes me wonder if it would be better to just have individual `tryCatch()`s in each of the functions.

R/rows.R
```diff
@@ -775,12 +775,19 @@ rows_insert_prep <- function(con, x_name, y, by, lvl = 0) {
   out
 }
 
-rows_get_or_execute <- function(x, sql, returning_cols) {
+rows_get_or_execute <- function(x, sql, returning_cols, call = caller_env()) {
   con <- remote_con(x)
+  msg <- "Can't modify database table {.val {remote_name(x)}}."
   if (is_empty(returning_cols)) {
```
hadley, line 781: e.g. here the `tryCatch()` could wrap both the `dbExecute()` and `the `dbGetQuery()`

R/verb-compute.R
```diff
@@ -81,6 +81,12 @@ collect.tbl_sql <- function(x, ..., n = Inf, warn_incomplete = TRUE, cte = FALSE
   }
 
   sql <- db_sql_render(x$src$con, x, cte = cte)
-  out <- db_collect(x$src$con, sql, n = n, warn_incomplete = warn_incomplete)
+  call <- current_env()
+  tryCatch(
+    out <- db_collect(x$src$con, sql, n = n, warn_incomplete = warn_incomplete),
+    error = function(cnd) {
+      cli_abort("Can't collect lazy table.", parent = cnd, call = call)
```
hadley, line 88: Are you sure you need the `call` here?

tests/testthat/_snaps/backend-postgres.md
hadley: I think that would be useful, but I can't come up with any good phrasing for it.

I think the ideal would be something like this, but we can't customise like that currently:

```
Error in `rows_insert()`:
 ! Can't modify table "df_x".
Caused by database error:
! Failed to fetch row: ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification
```

tests/testthat/test-db-sql.R
```diff
@@ -19,3 +19,15 @@ test_that("sql_query_rows() works", {
     sql("SELECT COUNT(*) FROM `abc` AS `master`")
   )
 })
+
+test_that("handles DBI error", {
+  con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
+  on.exit(DBI::dbDisconnect(con))
+
+  expect_error(db_analyze(con, "tbl"), class = "dbplyr_error_dbi")
```
hadley, line 27: And I think these are probably better as snapshot tests as that helps you see whether or not the `call` has been correctly passed through.

R/verb-compute.R
hadley: And something about the wording seems off to me? Maybe "failed to collect lazy table" or "can't execute SQL"?

hadley (commented):
Overall, results look great. But lots of questions about the implementation 😄 

I don't know how to think about the consequences of adding a new argument to the generic (i.e. what do extension packages need to do and what happens if there's a version mismatch between installed dbplyr and the version they were developed with). So I'd prefer to just add it to individual methods, like you've done here.

tests/testthat/test-db-sql.R
mgirlich: The reason I didn't want to use snapshot tests is that the errors rely on part on the error produced by SQLite. But probably they are stable enough.

tests/testthat/_snaps/db-io.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+    Code
+      (expect_error(db_copy_to(con = con, table = "tmp2", values = data.frame(x = c(1,
+        1)), unique_indexes = list("x"))))
+    Output
+      <error/rlang_error>
+      Error in `db_copy_to()`:
+      ! Can't copy to table "tmp2"
+      Caused by error in `db_create_index.DBIConnection()`:
```
mgirlich, line 24: I'm a bit confused that this mentions the method and above it is the generic. But that seems to be something related to rlang and probably not really worth to fix.

R/verb-compute.R
mgirlich: Nice, I don't need `call`. rlang really does a great job 😄

tests/testthat/test-db-sql.R
mgirlich: I was a bit reluctant on snapshot tests as they rely on the database error as well. But probably they are stable enough.

tests/testthat/_snaps/backend-postgres.md
mgirlich: Maybe just prepend the `cnd$message` by something like `[Database error]` or `in database:`

```
Error in `rows_insert()`:
      ! Can't modify database table "df_x".
      Caused by error:
      ! [Database error] Failed to fetch row: ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification
```

```
Error in `rows_insert()`:
      ! Can't modify database table "df_x".
      Caused by error:
      ! in database: Failed to fetch row: ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification
```

It's not quite a nice as `Caused by database error` but should still help to understand the issue better.

tests/testthat/test-db-sql.R
hadley: Ooooh, that makes sense. You could maybe use `transform` to rip out the database specific stuff?

mgirlich: I simply replaced everything after the last `Caused by error:` with `dummy DBI error`. This should be sufficiently stable.

tests/testthat/_snaps/backend-postgres.md
hadley, line 107: Do you want to dummy these out too now?

tests/testthat/_snaps/db-io.md
```diff
@@ -0,0 +1,28 @@
+# db_copy_to() wraps DBI errors
+
+    Code
+      (expect_error(db_copy_to(con = con, table = "tmp", values = data.frame(x = c(1,
+        1)))))
+    Output
+      <error/rlang_error>
+      Error in `db_copy_to()`:
```
hadley, line 8: This chain is particularly nice IMO

tests/testthat/_snaps/db-io.md
hadley: Yeah, I wouldn't worry about it either.

hadley (approved):

PR #964, opened 2022-08-08
Add `types` to `copy_inline()`

### LINKED ISSUE #963: Add `types` argument to `copy_inline()`
Similar to `copy_to()` there should be an argument `types` in `copy_inline()`. This is useful when working with non-R column types, e.g. for `rows_insert()`. Also see #909.

mgirlich:
Closes #963.

Maybe we should support (also in `copy_to()`) for `types` to be a lazy table where the types can be inferred from.


R/verb-copy-to.R
```diff
@@ -261,41 +265,74 @@ sql_values_clause <- function(con, df, row = FALSE) {
...
   }
 
-  typed_cols <- purrr::map_chr(
-    vctrs::vec_init(df),
-    ~ {
-      cast_expr <- call2(sql_cast_dispatch(.x), NA)
-      translate_sql(!!cast_expr, con = con)
-    }
-  )
+  # typed_cols <- purrr::map_chr(
```
hadley, line 275: I think you can delete this?

R/verb-copy-to.R
```diff
@@ -107,6 +107,9 @@ auto_copy.tbl_sql <- function(x, y, copy = FALSE, ...) {
 #' @param con A database connection.
 #' @param df A local data frame. The data is written directly in the SQL query
 #'   so it should be small.
+#' @param types a character vector giving variable types to use for the columns.
```
hadley, line 110:
```suggestion
#' @param types A character vector giving variable types to use for the columns.
```

Are the types automatically translated to backend specific types? That's what the link to the SQL docs imply to me. But I don't think that's how the code works?

mgirlich: Ah, good point. I simply copied this from the `copy_to()` documentation but it does not really make sense to point to the SQLite documentation.

R/verb-copy-to.R
```diff
@@ -261,41 +274,67 @@ sql_values_clause <- function(con, df, row = FALSE) {
...
+        val <- if (na) NA else ident(.y)
+        sql_expr(cast(!!val %as% !!sql(.x)), con = con)
+      }
+    )
+  }
+
   sql_vector(typed_cols, parens = FALSE, collapse = NULL, con = con)
 }
 
+sql_values_select <- function(con, df, types) {
```
hadley, line 318: Can at least remove the commented code and maybe this entire function?

hadley (approved):
Looks good! 

(BTW I think I forgot to mention that if I "approve" a PR with some comments, that basically means I'd like you to look at and consider the comments, but it's up to your judgement and you should merge when you think it's ready)

PR #965, opened 2022-08-08
`rows_*()` casts `y` columns if it copies them

### LINKED ISSUE #909: How to handle other column types in `rows_append()` and `rows_insert()`?
Inserting in a `jsonb` column is not that easy

``` r

con <- DBI::dbConnect(RPostgres::Postgres())

if (DBI::dbExistsTable(con, "..json_test")) {
  DBI::dbRemoveTable(con, "..json_test")
}

DBI::dbExecute(
  con,
  'CREATE TABLE "..json_test" (
    id integer NOT NULL,
    json_field jsonb
  )'
)
#> [1] 0

data.frame(id = 1, json_field = "[1, 2, 3]") %>% 
  rows_append(
    x = tbl(con, "..json_test"),
    copy = TRUE,
    in_place = TRUE
  )
#> Error: Failed to fetch row: ERROR:  column "json_field" is of type jsonb but expression is of type text
#> LINE 2: SELECT *
#>                ^
#> HINT:  You will need to rewrite or cast the expression.
DBI::dbReadTable(con, "..json_test")
#> [1] id         json_field
#> <0 rows> (or 0-length row.names)

# need to explicitly copy and cast beforehand
copy_inline(con, data.frame(id = 1, json_field = "[1, 2, 3]")) %>% 
  mutate(json_field = sql("json_field::jsonb")) %>% 
  rows_append(
    x = tbl(con, "..json_test"),
    in_place = TRUE
  )
DBI::dbReadTable(con, "..json_test")
#>   id json_field
#> 1  1  [1, 2, 3]
```

<sup>Created on 2022-06-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Maybe we can add explicit casts. Otherwise, we should add an example how to deal with this yourself

mgirlich:
Closes #909.
It would also be possible to add explicit casts to `y` columns if `y` is a database table (using `mutate()` on `y`). What do you think?


R/rows.R
```diff
@@ -775,6 +775,27 @@ rows_insert_prep <- function(con, x_name, y, by, lvl = 0) {
...
+  if (inherits(con, "TestConnection")) {
+    return(NULL)
+  }
+  if (is_null(name)) {
+    return(NULL)
+  }
+  res <- DBI::dbSendQuery(con, paste0("SELECT * FROM ", name))
+  on.exit(DBI::dbClearResult(res))
+  DBI::dbFetch(res, n = 0)
+  out <- DBI::dbColumnInfo(res)
```
hadley, line 794:
```suggestion
  res <- DBI::dbGetQuery(con, paste0("SELECT * FROM ", name), n = 0)
  out <- DBI::dbColumnInfo(res)
```
?

R/rows.R
```diff
@@ -775,6 +775,27 @@ rows_insert_prep <- function(con, x_name, y, by, lvl = 0) {
...
+
+  set_names(out[[".typname"]] %||% out[["type"]], out[["name"]])
```
hadley, line 796: I think `type` should come first (since that's the spec that `dbColumnInfo()` documents.

I feel like it would be a good idea to somehow test that this still works even if `dbColumnInfo()` returns a column with a different name. I have no idea how common this is, but I do worry about the level of variability in the implementation of `dbColumnInfo()`. I think it currently works sort of by coincidence because `get_col_types()` returns `NULL`, then `NULL[colnames(y)]` is still `NULL`, and presumably `auto_copy(types = NULL)` activates some default behaviour.

hadley (approved):

R/rows.R
mgirlich: Unfortunately, this does not work. `dbColumnInfo()` needs a DBIResult, not a dataframe.

R/rows.R
mgirlich: I now had to realise this does not work for other databases 😞  So, the types are only inferred for Postgres and nothing is done for other databases. To make this work properly for other databases requires either fixes in the corresponding db package or custom SQL on our side.
I added a generic so that users can help us and provide such SQL.

PR #975, opened 2022-08-15
Multiple joins in one query

### LINKED ISSUE #865: Multiple joins in one query
In databases one often needs to join multiple tables. In dbplyr this produces rather many nested queries which one would more often write as a single query.

## Example

``` r

  left_join(lf3, by = "x")
```

Currently produces

```sql
SELECT `LHS`.`x` AS `x`, `a`, `b`, `c`
FROM (
  FROM `df1` AS `LHS`
  LEFT JOIN `df2` AS `RHS`
    ON (`LHS`.`x` = `RHS`.`x`)
) `LHS`
LEFT JOIN `df3` AS `RHS`
  ON (`LHS`.`x` = `RHS`.`x`)
```

It would be nicer if it could produce something like

```sql
SELECT `df`.`x` AS `x`, `a`, `b`, `c`
FROM `df1`
LEFT JOIN `df2`
  ON (`df1`.`x` = `df2`.`x`)
LEFT JOIN `df3`
  ON (`df1`.`x` = `df3`.`x`)
```

## Thoughts & Questions

* Is the result of joins in subqueries and multiple joins in one query necessarily the same?
* What about table aliases?
  * Now we have `x_as` and `y_as`. I think if they are not provided it might make more sense to not use a table alias.
  * If `x_as` is provided in a join which is not the first join, then maybe a subquery should be generated
  * If `y_as` is provided it can always be used
* A `FULL JOIN` can be tricky to combine with other joins:
  * SQLite does not directly support `FULL JOIN`
  * The columns joined by are `coalesce()`
* If `semi_join()` or `anti_join()` is followed by `left/right/inner/full_join()` they cannot be combined because `WHERE` is evaluated after `JOIN`

So, I think that
* a sequence of `left_join()` and `inner_join()` can be combined in one query
* a sequence of `semi_join()` and `anti_join()` might be combined (though nested queries might be more efficient)

mgirlich:
Closes #865. This PR replaces #894.

The code currently has some rough edges that still need to be improved. One has to be quite careful when a series of `left/inner_join()` can be done in a single query but it produces much nicer queries in the end.

## Remarks
* For simplicity, I just used `vec_as_names()` to produce unique names. We might want something else? I just remembered that we considered using `abbreviate()`. I'll take a look at that in another PR.
* The `lazy_join_query` structure is basically not needed in this our codebase anymore. It would make sense to remove it but as it is exported I wanted to check whether it is in use somewhere or not.

## Names for Subqueries

```r

left_join(
  lf %>% filter(x == 1),
  lf %>% filter(a > 1),
  by = "x"
)
```

Produces

```sql
SELECT `LHS`.`x` AS `x`, `LHS`.`a` AS `a.x`, `RHS`.`a` AS `a.y`
FROM (
  SELECT *
  FROM `df`
  WHERE (`x` = 1.0)
) `LHS`
LEFT JOIN (
  SELECT *
  FROM `df`
  WHERE (`a` > 1.0)
) `RHS`
  ON (`LHS`.`x` = `RHS`.`x`)
```

## Self-Join

```r

left_join(
  lf,
  lf,
  by = "x"
)
```

```sql
SELECT `df_LHS`.`x` AS `x`, `df_LHS`.`a` AS `a.x`, `df_RHS`.`a` AS `a.y`
FROM `df` AS `df_LHS`
LEFT JOIN `df` AS `df_RHS`
  ON (`df_LHS`.`x` = `df_RHS`.`x`)
```

## User Provided Alias

```r
left_join(
  lf,
  lf,
  by = "x",
  x_as = "df"
) %>% 
  left_join(lf, by = "x")
```

The alias provided by the user always wins. So this produces

```sql
SELECT
  `df`.`x` AS `x`,
  `df`.`a` AS `a.x`,
  `df...2`.`a` AS `a.y`,
  `df...3`.`a` AS `a`
FROM `df`
LEFT JOIN `df` AS `df...2`
  ON (`df`.`x` = `df...2`.`x`)
LEFT JOIN `df` AS `df...3`
  ON (`df`.`x` = `df...3`.`x`)
```

Note that an alias for `x` can also be provided later on. So, this produces the same query:

```r
left_join(
  lf,
  lf,
  by = "x"
) %>% 
  left_join(lf, by = "x", x_as = "df")
```

## Conflicting Alias

In case of a conflicting alias a subquery is used

```r
left_join(
  lf,
  lf,
  x_as = "df",
  by = "x"
) %>% 
  left_join(lf, by = "x", x_as = "df2")
```

```sql
SELECT `df2`.*, `a`
FROM (
  SELECT `df`.`x` AS `x`, `df`.`a` AS `a.x`, `df...2`.`a` AS `a.y`
  FROM `df`
  LEFT JOIN `df` AS `df...2`
    ON (`df`.`x` = `df...2`.`x`)
) `df2`
LEFT JOIN `df`
  ON (`df2`.`x` = `df`.`x`)
```


R/lazy-join-query.R
```diff
@@ -139,8 +180,12 @@ op_vars.lazy_join_query <- function(op) {
   op$vars$alias
 }
 #' @export
+op_vars.lazy_multi_join_query <- function(op) {
+  op$vars$name
+}
+#' @export
 op_vars.lazy_semi_join_query <- function(op) {
-  names(op$vars)
+  op$vars$name
```
mgirlich, line 188: I changed the `vars` field of `lazy_semi_join_query` to a tibble with columns `name` and `var`, so that it is more similar to the `vars` field of `lazy_multi_join_query`.

R/lazy-join-query.R
```diff
@@ -156,15 +201,83 @@ sql_build.lazy_join_query <- function(op, con, ...) {
...
 
+#' @export
+sql_build.lazy_multi_join_query <- function(op, con, ...) {
+  auto_name <- is.na(op$table_names$as)
+  table_names_out <- dplyr::coalesce(op$table_names$as, op$table_names$name)
+  if (length(table_names_out) == 2) {
+    table_names_out <- join_simple_table_alias(op$table_names$name, op$table_names$as)
+
+    if (op$joins$type %in% c("full", "right")) {
+      # construct a classical `join_query()` so that special handling of
```
mgirlich, line 212: I think `inner_join()`, `left_join()` and `cross_join()` should work the same in every database. But I think this fallback makes sense for `right_join()` and `full_join()`.

R/lazy-join-query.R
```diff
@@ -156,15 +201,83 @@ sql_build.lazy_join_query <- function(op, con, ...) {
...
+    all_vars_list = all_vars_list,
+    duplicated_vars = duplicated_vars
+  )
+}
+
 #' @export
 sql_build.lazy_semi_join_query <- function(op, con, ...) {
-  vars <- op$vars
   vars_prev <- op_vars(op$x)
-  if (identical(unname(vars), names(vars)) &&
```
mgirlich, line 163: Due to changing the `vars` field to a tibble.

R/verb-joins.R
```diff
@@ -211,20 +207,23 @@ anti_join.tbl_lazy <- function(x, y, by = NULL, copy = FALSE,
...
+                     copy,
+                     suffix,
+                     auto_index,
+                     na_matches,
+                     x_as,
+                     y_as,
                      call = caller_env()) {
   if (!is.null(sql_on)) {
-    by <- list(x = character(0), y = character(0), on = sql(sql_on))
+    by <- list(x = character(0), y = character(0), on = unclass(sql_on))
```
mgirlich, line 226: It would be great to implement `ident()`, `ident_q()` and `sql()` on top of vctrs at some point. They have some annoying issues that make it difficult to work with...

R/verb-joins.R
```diff
@@ -237,90 +236,216 @@ add_join <- function(x, y, type, by = NULL, sql_on = NULL, copy = FALSE,
...
+  x_lq <- inline_result$lq
+  x_vars <- inline_result$vars
+  by_x_org <- by$x
+  by$x <- inline_result$by
+
+  new_query <- join_needs_new_query(x$lazy_query, join_alias, type)
+  if (new_query) {
+    x_join_vars <- tibble(
+      name = op_vars(x),
+      table = list(1L),
```
mgirlich, line 256: At this point we don't know the names used for the tables in the multi join query. So, `table` uses an integer to create a reference to the table. It must be a list to support `full_join()`.
Also `vars` must be a list to support `full_join()` together with inlining `select()`. This is a bit ugly but the alternatives I tried caused way more trouble.

mgirlich:
I split this PR into multiple smaller ones to make reviews easier.

PR #977, opened 2022-08-16
Add option to control keyword highlighting

### LINKED ISSUE #974: feature request: Turn keyword highlighting off
I'd like keyword highlighting, but the dark blue is hard to read with a dark grey background. Would be nice if the option allowed specifying a color, but failing that I'd like to be able to turn it off.

``` r
library(dplyr, w = F)
library(testthat, w = F)
#> Warning: package 'testthat' was built under R version 4.1.2
copy_to(con, mtcars)
mtcars2 <- tbl(con, "mtcars")

local_reproducible_output(crayon = TRUE)
options(dbplyr_use_colour = FALSE)

mtcars2 %>% 
  select(mpg) %>% 
  show_query()
#> <SQL>
#> [34mSELECT[39m `mpg`
#> [34mFROM[39m `mtcars`
```

<sup>Created on 2022-08-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

My workaround for now is below, but it would be nice if there was an option 

```{r}
style_kw <- function(x) {
  if (dbplyr:::dbplyr_use_colour()) {
    # use `br_blue` which is more legible for a dark theme
    cli::col_br_blue(x)
  } else {
    x
  }
}
 
assignInNamespace(
  "style_kw", 
  style_kw,
  asNamespace("dbplyr")
)
 ```

### PR DESCRIPTION
**Pull request by mgirlich**
Fixes #974.

It would be even nicer if we could detect a dark theme ourselves but cli does not export `detect_dark_theme()`.
Where should this be documented so that users can easily find it?

### DIFF
--- a/tests/testthat/test-verb-select.R
+++ b/tests/testthat/test-verb-select.R
@@ -248,6 +248,7 @@ test_that("mutate collapses over nested select", {
 
 test_that("output is styled", {
   local_reproducible_output(crayon = TRUE)
+  withr::local_options(dbplyr_highlight = cli::combine_ansi_styles("blue"))
 
   lf <- lazy_frame(x = 1, y = 1, z = 1)
   out <- lf %>%
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # dbplyr (development version)
 
+* The keyword highlighting can now be customised via the option `dbplyr_highlight`.
+  Turn it off via `options(dbplyr_highlight = FALSE)` or pass a custom ansi
+  style, e.g. `options(dbplyr_highlight = cli::combine_ansi_styles("bold", "cyan"))`
+  (@mgirlich, #974).
+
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: tests/testthat/test-verb-select.R [hunk 1]
```diff
@@ -248,6 +248,7 @@ test_that("mutate collapses over nested select", {
 
 test_that("output is styled", {
   local_reproducible_output(crayon = TRUE)
+  withr::local_options(dbplyr_highlight = cli::combine_ansi_styles("blue"))
```
hadley, line 251: Why do we need this now?

hadley (approved):
The obvious place to document this would be `show_query()` but that's in dplyr. I don't think there is anywhere that users would think to look 😞

tests/testthat/test-verb-select.R
mgirlich: It is needed in case you use the `dbplyr_highlight` option and set it to something different than blue 😉

PR #980, opened 2022-08-19
Refactor `check_join_as()`

mgirlich:
I split the PR #975 into multiple smaller PRs to make it easier to review.

### DIFF
--- a/R/verb-joins.R
+++ b/R/verb-joins.R
@@ -243,14 +243,15 @@ add_join <- function(x, y, type, by = NULL, sql_on = NULL, copy = FALSE,
   # the table alias can only be determined after `select()` was inlined.
   # This works even though `by` is used in `inline_select_in_join()` and updated
   # because this does not touch `by$x_as` and `by$y_as`.
-  by[c("x_as", "y_as")] <- check_join_as(
-    x_as,
-    inlined_select_list$x,
-    y_as,
-    inlined_select_list$y,
-    sql_on = sql_on,
-    call = call
+  join_alias <- check_join_alias(x_as, y_as, sql_on, call)
+
+  table_names <- c(
+    unclass(query_name(inlined_select_list$x)) %||% NA,
+    unclass(query_name(inlined_select_list$y)) %||% NA
   )
+  by[c("x_as", "y_as")] <- join_simple_table_alias(table_names, join_alias)
+  by$x_as <- ident(by$x_as)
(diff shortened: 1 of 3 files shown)


R/verb-joins.R
```diff
@@ -435,6 +464,43 @@ join_vars <- function(x_names, y_names, type, by, suffix = c(".x", ".y"), call =
   )
 }
 
+join_simple_table_alias <- function(table_names, aliases) {
```
hadley, line 467: Would it be simpler to have arguments `x_alias`, `y_alias`, `x_name`, `y_name`? Or if this exists in preparation for having more than two table names, I think you could implement more generically in this PR.

R/verb-joins.R
```diff
@@ -435,6 +464,43 @@ join_vars <- function(x_names, y_names, type, by, suffix = c(".x", ".y"), call =
...
+  stopifnot(length(table_names) == 2)
+  stopifnot(length(aliases) == 2)
+  aliases <- unlist(aliases)
+
+  x_alias <- aliases[[1]]
+  y_alias <- aliases[[2]]
+  x_name <- table_names[[1]]
+  y_name <- table_names[[2]]
+
+  x_out <- dplyr::coalesce(x_alias, x_name, "LHS")
```
hadley, line 477: I think of `coalesce()` as a vector operation, so this feels a bit off to me. I'd prefer to see `x <- x_out %||% x_alias %||% x_name %||% "LHS"` but I can see why that would be painful to implement given the current structure. I'm not sure whether or not that suggests some bigger changes are necessary.

mgirlich: This was actually quite simple to change. I used `coalesce()` due to its origin in the multi join PR. But I also prefer using `%||%`. Let's see whether this will be annoying for the multi join PR or not.

R/verb-joins.R
mgirlich: Yes, simpler with these arguments and it also showed that the function name should be changed to `join_two_table_alias()`

R/verb-joins.R
```diff
@@ -435,6 +460,39 @@ join_vars <- function(x_names, y_names, type, by, suffix = c(".x", ".y"), call =
   )
 }
 
+join_two_table_alias <- function(x_name, y_name, x_alias, y_alias) {
```
hadley, line 463: I think this looks much better 😄

PR #961, opened 2022-08-04
Shim for redshift values table support

ejneer:
fixes #949 

```r
copy_inline(con, head(mtcars[c("mpg", "cyl", "disp")])) %>% show_query()
```

before:
```sql
<SQL>
SELECT
  CAST("mpg" AS FLOAT) AS "mpg",
  CAST("cyl" AS FLOAT) AS "cyl",
  CAST("disp" AS FLOAT) AS "disp"
FROM (
  (
    SELECT NULL AS "mpg", NULL AS "cyl", NULL AS "disp"
    WHERE (0 = 1)
  )
  UNION ALL
  (
  VALUES
    (21.0, 6.0, 160.0),
    (21.0, 6.0, 160.0),
    (22.8, 4.0, 108.0),
    (21.4, 6.0, 258.0),
    (18.7, 8.0, 360.0),
    (18.1, 6.0, 225.0)
  )
) "values_table"
```

after:
```sql
<SQL>
SELECT
  CAST("mpg" AS FLOAT) AS "mpg",
  CAST("cyl" AS FLOAT) AS "cyl",
  CAST("disp" AS FLOAT) AS "disp"
FROM (
  (
    SELECT NULL AS "mpg", NULL AS "cyl", NULL AS "disp"
    WHERE (0 = 1)
  )
  UNION ALL
  ( SELECT 21.0, 6.0, 160.0 UNION ALL SELECT 21.0, 6.0, 160.0 UNION ALL SELECT 22.8, 4.0, 108.0 UNION ALL SELECT 21.4, 6.0, 258.0 UNION ALL SELECT 18.7, 8.0, 360.0 UNION ALL SELECT 18.1, 6.0, 225.0)
) "values_table"
```

PR #982, opened 2022-08-22
Use `UNION ALL` instead of `VALUES` for some DBs

### LINKED ISSUE #949: copy_inline() and copy_to() not working in AWS Redshift
copy_inline() and copy_to() do not work in AWS Redshift.
Generated SQL appears to be syntactically incompatible with Redshift PostgreSQL:

Error: Failed to prepare query: ERROR:  syntax error at or near ","
LINE 30: ...f64faee3fc', 'nc', '27607', '3', '035', 'p', 's', 'h', '3'),
                                                                       ^

### LINKED ISSUE #950: `copy_inline` does not work for HANA DB.
Hi,
thanks a lot for implementing the extremely useful `copy_inline` method.
Unfortunately, the generated SQL fails for HANA DB.

I cannot create a reprex since it requires a real ocdb connection.
But here is the code:

```r
# establish a DBI connection and store in variable `con`

copy_inline(con, tibble(a=c(1,1), b=c(2,3)))
```

> Fehler in new_result(connection@ptr, statement, immediate) : 
  nanodbc/nanodbc.cpp:1412: 42000: [SAP AG][LIBODBCHDB DLL][HDBODBC] Syntax error or access violation;257 sql syntax error: incorrect syntax near "WHERE": line 7 col 7 (at pos 140)

The SQL it generates:
```r
copy_inline(con, tibble(a=c(1,1), b=c(2,3))) %>% show_query()
```
```SQL
<SQL>
SELECT CAST("a" AS DOUBLE) AS "a", CAST("b" AS DOUBLE) AS "b"
FROM (
  (
    SELECT NULL AS "a", NULL AS "b"
    WHERE (0 = 1)
  )
  UNION ALL
  (VALUES (1.0, 2.0), (1.0, 3.0))
) "values_table"
```

### LINKED ISSUE #972: copy_inline() not working on oracle backends
This is basically a repeat of the issues listed in https://github.com/tidyverse/dbplyr/issues/950, but for Oracle backends.

```
copy_inline(con, tibble(a=c(1,1), b=c(2,3)))
```
> Error in .oci.SendQuery(conn, statement, data = data, prefetch = prefetch,  : 
>  ORA-00903: invalid table name

As for the Hana DB it seems the `values` command does not yield the desired result and you need use `union all ` in combination with selecting from `dual` in the query. So instead of the previous query this works:

```
SELECT CAST("a" AS NUMBER) AS "a", CAST("b" AS NUMBER) AS "b"
FROM (
  (
    SELECT NULL AS "a", NULL AS "b"
    FROM dual 
    WHERE (0 = 1)
  )
  UNION ALL
  (select 1.0, 2.0 from dual union all select 1.0, 3.0 from dual)
) "values_table"

```

mgirlich:
Fixes #949, fixes #950 and fixes #972.
Replaces #961.


ejneer:
I believe you'll also need a `sql_values_subquery.Redshift` method. My redshift connection doesn't seem to inherit from `RedshiftConnection` so it falls through to the default.

```r
> con <- DBI::dbConnect(odbc::odbc(), dsn = "my-redshift-dsn")
> inherits(con, "Redshift")
[1] TRUE
> inherits(con, "RedshiftConnection")
[1] FALSE
```

ejneer:
looks good for redshift 👍

Issue #970, opened 2022-08-10
Include a little about sub-queries in translation-verb vignette


hadley:
i.e. show a couple of cases where they are always required (e.g. `df |> mutate(x1 = x + 1, x2 = x1 + 1)`) and discuss some of the optimisations that dbplyr now performs to avoid them

Issue #896, opened 2022-05-25
Rethink `tbl()`/`lazy_frame()` names


hadley:
Neither name really conveys connection to the database. Not sure what they should be instead, but could make use a `db_` prefix? Maybe `db_table()` and `db_simulate()`? Then could add `db_query()` as a more direct/discoverable equivalent to `tbl(con, SQL(...))`?

hadley:
The `db_` prefix isn't great because of the existing generics, so maybe `query_table()`, `query_sql()`, and `query_mimic()`?

SQL is only 3 letters, but it's close enough that it doesn't feel far away.

Issue #863, opened 2022-05-08
Option to use `copy_inline()` in `left_join()` and friends


hadley:
Maybe `copy = "inline"` as an additional option?

PR #984, opened 2022-08-25
Multiple joins in one query

### LINKED ISSUE #865: Multiple joins in one query

mgirlich:
Closes #865.

One has to be quite careful when a series of `left/inner_join()` can be done in a single query but it produces much nicer queries in the end.

## Remarks
* For simplicity, I just used `vec_as_names()` to produce unique names. We might want something else? I just remembered that we considered using `abbreviate()`. I'll take a look at that in another PR.
* The `lazy_join_query` structure is basically not needed in this our codebase anymore. It would make sense to remove it but as it is exported I wanted to check whether it is in use somewhere or not.

## Names for Subqueries

```r

left_join(
  lf %>% filter(x == 1),
  lf %>% filter(a > 1),
  by = "x"
)
```

Produces

```sql
SELECT `LHS`.`x` AS `x`, `LHS`.`a` AS `a.x`, `RHS`.`a` AS `a.y`
FROM (
  SELECT *
  FROM `df`
  WHERE (`x` = 1.0)
) `LHS`
LEFT JOIN (
  SELECT *
  FROM `df`
  WHERE (`a` > 1.0)
) `RHS`
  ON (`LHS`.`x` = `RHS`.`x`)
```

## Self-Join

```r

left_join(
  lf,
  lf,
  by = "x"
)
```

```sql
SELECT `df_LHS`.`x` AS `x`, `df_LHS`.`a` AS `a.x`, `df_RHS`.`a` AS `a.y`
FROM `df` AS `df_LHS`
LEFT JOIN `df` AS `df_RHS`
  ON (`df_LHS`.`x` = `df_RHS`.`x`)
```

## User Provided Alias

```r
left_join(
  lf,
  lf,
  by = "x",
  x_as = "df"
) %>% 
  left_join(lf, by = "x")
```

The alias provided by the user always wins. So this produces

```sql
SELECT
  `df`.`x` AS `x`,
  `df`.`a` AS `a.x`,
  `df...2`.`a` AS `a.y`,
  `df...3`.`a` AS `a`
FROM `df`
LEFT JOIN `df` AS `df...2`
  ON (`df`.`x` = `df...2`.`x`)
LEFT JOIN `df` AS `df...3`
  ON (`df`.`x` = `df...3`.`x`)
```

Note that an alias for `x` can also be provided later on. So, this produces the same query:

```r
left_join(
  lf,
  lf,
  by = "x"
) %>% 
  left_join(lf, by = "x", x_as = "df")
```

## Conflicting Alias

In case of a conflicting alias a subquery is used

```r
left_join(
  lf,
  lf,
  x_as = "df",
  by = "x"
) %>% 
  left_join(lf, by = "x", x_as = "df2")
```

```sql
SELECT `df2`.*, `a`
FROM (
  SELECT `df`.`x` AS `x`, `df`.`a` AS `a.x`, `df...2`.`a` AS `a.y`
  FROM `df`
  LEFT JOIN `df` AS `df...2`
    ON (`df`.`x` = `df...2`.`x`)
) `df2`
LEFT JOIN `df`
  ON (`df2`.`x` = `df`.`x`)
```


mgirlich:
@hadley I think there are enough nice new features to make another release. Is there anything else you want to include?
Maybe you want to have a look at the PR #918 and your issues #896, #897, #970, and #863.

hadley:
@mgirlich I don't have time for those issues at the moment (and I'd need to think more about #918) so think we're good to go for a release. Are you thinking 2.2.2 or 2.3.0? (The main difference is whether or not you think we should do a blog post). Do you want to handle most of the release (i.e. work through the bullets in `use_release_issue()`) or do you want me to do it?

mgirlich:
The release contains

* much better error messages
* a couple of small bugfixes and minor features
* new translations for some backends
* much nicer SQL

While I really like these features I don't necessarily see the need for a blog post. Also there are a couple of other improvements for nicer SQL still missing (e.g. this PR) and it feels a bit weird to blog about half finished work.

I created a release issue #994.

hadley:
Can you please give me a suggested reading guide for this PR?

R/verb-joins.R
```diff
@@ -297,19 +347,108 @@ join_inline_select <- function(lq, by, on) {
...
 
-update_join_vars <- function(vars, x_lq, y_lq, x_vars, y_vars, type, by) {
-  vars$all_x <- op_vars(x_lq)
-  vars$all_y <- op_vars(y_lq)
+join_needs_new_query <- function(x_lq, join_alias, type) {
+  if (!inherits(x_lq, "lazy_multi_join_query")) {
+    return(TRUE)
+  }
+
+  if (!type %in% c("left", "inner")) {
```
hadley, line 355: Maybe `type %in% c("full", "right")` to match the condition I discussed above?

R/verb-joins.R
```diff
@@ -297,19 +347,108 @@ join_inline_select <- function(lq, by, on) {
...
+    return(TRUE)
+  }
+
+  x_as <- join_alias$x
+  y_as <- join_alias$y
+
+  as_current <- x_lq$table_names$as
+  if (!is_null(x_as)) {
```
hadley, line 363: This feels a bit fiddly to me and I wonder if we're putting too much emphasis on `x_as`/`y_as` which should be relatively rarely used features. I think it would be fine if the last specified alias wins (assuming that's easier to implement)

tests/testthat/_snaps/verb-joins.md
```diff
@@ -63,16 +63,31 @@
...
+      remote_query(out)
+    Output
+      <SQL> SELECT `df3`.`x` AS `x`, `a`, `LHS`.`b` AS `b.x`, `df3`.`b` AS `b.y`
+      FROM (
+        SELECT `df1`.*, `b`
+        FROM `df1`
+        LEFT JOIN `df2`
+          ON (`df1`.`x` = `df2`.`x`)
+      ) `LHS`
+      RIGHT JOIN `df3`
```
hadley, line 78: It seems weird to have a result with multiple queries for a test with title "multiple joins create a single query"

tests/testthat/test-verb-joins.R
```diff
@@ -447,8 +564,9 @@ test_that("join verbs generate expected ops", {
 })
 
 test_that("can optionally match NA values", {
-  lf <- lazy_frame(x = 1)
-  expect_snapshot(left_join(lf, lf, by = "x", na_matches = "na"))
+  lf1 <- lazy_frame(x = 1, .name = "lf1")
```
hadley, line 567: Might be worth using `simulate_postgres()` here to get the slightly simpler SQL.

tests/testthat/test-verb-joins.R
```diff
@@ -415,27 +415,144 @@ test_that("select() before join is not inlined when using `sql_on`", {
...
+  out <- left_join(lf, lf2, by = "x") %>%
+    inner_join(lf3, by = "x")
+  lq <- out$lazy_query
+  expect_s3_class(lq, "lazy_multi_join_query")
+  expect_equal(lq$table_names, tibble(as = NA_character_, name = c("df1", "df2", "df3")))
+  expect_equal(lq$vars$name, c("x", "a", "b.x", "b.y"))
+  expect_equal(lq$vars$table, list(1L, 1L, 2L, 3L))
+  expect_equal(lq$vars$var, list("x", "a", "b", "b"))
+
+  expect_equal(
```
hadley, line 432: This seems long enough to make a snapshot test worthwhile?

R/verb-joins.R
```diff
@@ -234,44 +234,94 @@ add_join <- function(x, y, type, by = NULL, sql_on = NULL, copy = FALSE,
...
+      joins = joins_data,
+      table_names = vctrs::vec_rbind(table_names_x, table_names_y),
+      vars = vars
+    )
+    return(out)
+  }
+
+  # `x_lq` must be a `lazy_multi_join_query` so it can be modified directly
+  if (!is_null(join_alias$x)) {
+    x_lq$table_names$as[[1]] <- join_alias$x
```
hadley, line 317: When I worked through the example you suggested, I noticed that an `NA` was replaced by a `NULL` which seems mildly fishy.

hadley (commented):
I can't say I followed every detail, but overall the code makes sense to me and I don't get the sense it's more complicated than it needs to be. Thanks for working on this!!

R/db-sql.R
```diff
@@ -387,6 +387,106 @@ sql_join.DBIConnection <- function(con, x, y, vars, type = "inner", by = NULL, n
...
+#'   on = c(NA, NA),
+#'   na_matches = c("never", "never")
+#' )
+sql_query_multi_join.DBIConnection <- function(con,
+                                               x,
+                                               joins,
+                                               table_vars,
+                                               vars,
+                                               ...,
+                                               lvl = 0) {
```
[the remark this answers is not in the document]
mgirlich: Yes, only for indenting. Would you rather rename it to `indent_lvl`?

R/db-sql.R
```diff
@@ -387,6 +387,106 @@ sql_join.DBIConnection <- function(con, x, y, vars, type = "inner", by = NULL, n
...
+  )
+
+  on_clauses <- purrr::pmap(
+    vctrs::vec_cbind(
+      rhs = table_names[-1],
+      joins[c("by_x", "by_x_table_id", "by_y", "on", "na_matches")]
+    ),
+    function(rhs, by_x, by_x_table_id, by_y, on, na_matches) {
+      if (!is.na(on)) {
+        on <- sql(on)
```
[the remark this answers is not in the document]
mgirlich: Turned out to be unnecessary after refactoring the `joins` data structure.

R/verb-joins.R
mgirlich: It would still require a check if `sql_on` is used, e.g.

```r
left_join(x, y, x_as = "x", y_as = "y", sql_on = "x.a = y.a") %>% 
  left_join(
    z,
    x_as = "dummy",
    y_as = "z",
    sql_on = "dummy.a = z.a"
  )
```

R/verb-joins.R
mgirlich: I can't quite follow where an `NA` was replaced by a `NULL`.

R/db-sql.R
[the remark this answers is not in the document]
hadley: It's probably fine, given that it's used elsewhere. If you do rename it, a separate PR would be better, I think.

R/verb-joins.R
hadley: That feels pretty fiddly to me too; I'm not sure we need to support so many edge cases.

R/verb-joins.R
hadley: IIRC, when I stepped through the code ` x_lq$table_names$as[[1]]` was `NA` and `join_alias$x` was `NULL`

mgirlich: I refactored the table names logic and it is now a bit easier.

R/verb-joins.R
mgirlich: I agree that it is quite fiddly. But I'm not sure there is a nice alternative. We need some kind of check to make sure that

```r
left_join(lf, lf, sql_on = "...") %>% 
  left_join(lf, sql_on = "...")
```

works. Of course, we could simply start a new query but in the end the checks won't be much simpler. Also, at least for now it feels like more work to change it.

hadley: Ok, lets leave it as is then.

mgirlich:
@hadley Is there anything else you want to have changed or can we merge this? 😄

hadley:
Looks good to me!

hadley:
Woo hooo!

PR #992, opened 2022-08-31
Add default value for collapse arg for str_flatten

fh-afrachioni:
Adds `collapse = ''` as the default for `str_flatten` for postgres, snowflake, redshift, and mysql.  This is the behavior of `stringr::str_flatten`.


mgirlich:
Seems reasonable, thanks! Can you add a NEWS bullet?

fh-afrachioni:
Added!

PR #1011, opened 2022-09-20
Use DATETIME2 for casting to date-times

krlmlr:
TIMESTAMP is the wrong type, https://stackoverflow.com/a/7105129/946850.


mgirlich:
Thanks!

PR #1017, opened 2022-10-05
Support `.default` arg in `case_when()`

mgirlich:


R/translate-sql-conditional.R
```diff
@@ -54,8 +60,11 @@ sql_case_when <- function(...) {
 
   clauses <- purrr::map2_chr(query, value, ~ paste0("WHEN ", .x, " THEN ", .y))
   # if a formula like TRUE ~ "other" is at the end of a sequence, use ELSE statement
+  # TRUE has precedence over `.default`
   if (is_true(formulas[[n]][[2]])) {
     clauses[[n]] <- paste0("ELSE ", value[[n]])
+  } else if (!is_null(.default)) {
+    clauses[[n + 1]] <- paste0("ELSE ", .default)
   }
```
mgirlich, line 68: This feels a bit hacky. But this should give the same behaviour as dplyr.

R/translate-sql-conditional.R
```diff
@@ -31,8 +31,14 @@ sql_if <- function(cond, if_true, if_false = quo(NULL), missing = quo(NULL)) {
   sql(paste0(out, " END"))
 }
 
-sql_case_when <- function(...) {
+sql_case_when <- function(..., .default = NULL, .ptype = NULL, .size = NULL) {
   # TODO: switch to dplyr::case_when_prepare when available
+  if (!is_null(.ptype)) {
```
hadley, line 36: Pass along `error_call` so the error message reports the error is coming from `case_when()`?  `mssql_case_when()` also needs similar treatment, although that's really getting into the weeds.

R/translate-sql-conditional.R
```diff
@@ -31,8 +31,14 @@ sql_if <- function(cond, if_true, if_false = quo(NULL), missing = quo(NULL)) {
...
+    cli_abort("{.arg .ptype} is not supported in SQL translations.")
+  }
+  if (!is_null(.size)) {
+    cli_abort("{.arg .size} is not supported in SQL translations.")
```
hadley, line 40: Might be worth extracting a `check_not_supplied()` function?

```R
check_not_supplied <- function(arg, call = caller_env()) {
  arg <- caller_arg(arg)
  cli_abort("{.arg {arg}} is not supported in SQL translations.", call = call)  
}
```

hadley (approved):

R/translate-sql-conditional.R
mgirlich: `mssql_case_when()` uses `sql_case_when()`, so this works directly.

PR #1018, opened 2022-10-05
Deprecate passing `...` to `across()`

mgirlich:


R/tidyeval-across.R
```diff
@@ -24,6 +28,28 @@ partial_eval_if <- function(call, data, env, reduce = "&", error_call = caller_e
...
+      "  # Now",
+      "  across(a:b, ~mean(.x, na.rm = TRUE))"),
+      collapse = "\n"
+    )
+    lifecycle::deprecate_warn(
+      when = "2.3.0",
+      what = "across(...)",
+      details = details,
+      env = env_parent(error_call),
+      user_env = env_parent(error_call, n = 4L),
```
mgirlich, line 48: This is quite a hack and I don't really understand it. But this was the first combination of environment that:

* didn't say that this is probably an issue in the dbplyr package and people should open an issue
* the warnings didn't appear every time

hadley: Did you try using `env` from `partial_eval_if` etc? I haven't looked at it closely, but my intuition tells me it's likely more closely related to that environment/

mgirlich: This looks less magical now 😄

hadley (approved):

PR #1034, opened 2022-10-26
Fix `round()` and `supports_window_clause()` for Redshift

### LINKED ISSUE #1033: Redshift `round()` can't return decimal places
Redshift's `numeric` (== `decimal`) type has a default scale (number of decimal places) of 0, so naively casting values to `numeric` can lead to surprising results.

The Redshift backend already guards against this, by instead casting to `float`:
https://github.com/tidyverse/dbplyr/blob/f4cef2360088c39620bd85f474ec2e888a520dcc/R/backend-redshift.R#L40

However the Postgres backend, which the Redshift backend inherits from, uses a [custom function](https://github.com/tidyverse/dbplyr/blob/ec2374f22debcd98a05cd2905273aa1f63cf6809/R/backend-postgres.R#L62) for its translation of `round()`, which contains an explicit cast to `numeric`:
```r
postgres_round <- function(x, digits = 0L) {
  digits <- as.integer(digits)
  sql_expr(round(((!!x)) %::% numeric, !!digits))
}
```

This means we end up with the following SQL translation:
```r
translate_sql(round(1.234, 1), con = simulate_redshift())
#> <SQL> ROUND((1.234) :: numeric, 1)
```

This _always:_
* Implicitly rounds to 0 decimal places, via the cast to `numeric`
* Then, explicitly rounds the result to the specified number of decimal places, meaning we end up with at best 0dp

Hopefully an easy fix by overriding `round` in the Redshift translator with a small `round_redshift()` function - I'll add a pull request shortly!

### LINKED ISSUE #1035: Redshift shouldn't support window clause
Currently, `supports_window_clause()` returns `TRUE` for Redshift connections, because the Redshift backend inherits from the Postgres backend, and Postgres _does_ support window clauses.

However, Redshift does _not_ currently support window clauses - at least, they are not mentioned anywhere in the AWS Redshift docs, and attempting to use a named window in a Redshift query in the same way as in a Postgres query, results in a syntax error.

Fixable by adding `supports_window_clause()` S3 methods for Redshift connections in `backend-redshift.R`.

owenjonesuob:
Closes #1033, closes #1035.


mgirlich:
Thanks! 😄 
Next time please also add a NEWS entry mentioning yourself and the issue(s) the PR closes. I added them myself this time so that this PR could be merged.

Issue #488, opened 2020-07-28
Overwrite in compute 


gregleleu:
I use compute to save non-temporary tables to remote databases. When I re-run some analysis I get an error as the table already exists. It would be nice if `compute` had an overwrite option like `copy_to`.

It feels like the philosophy in dbplyr is more around creating temporary tables, but it such cases are a natural extension I believe. 

```r
library(tidyverse)
library(magrittr)
library(DBI)

con <- dbConnect(RSQLite::SQLite(), ":memory:")

mtcars_rm <- con %>% copy_to(mtcars)

mtcars_rm %>% 
  summarise(mpg_avg = mean(mpg)) %>% 
  compute(name = "mtcars_avg", temporary = FALSE)

mtcars_rm %>% 
  summarise(mpg_avg = mean(mpg)) %>% 
  compute(name = "mtcars_avg", temporary = FALSE, overwrite = TRUE)
```

hadley:
All the `overwrite` argument in `copy_to` is deletes the table if it already exists, so I think I'd suggest you just do that — adding an `overwrite` argument to `compute()` feels a bit too "in the weeds" to me. Even though I'm closing this issue, I really appreciate the feedback, and hope you'll continue to contribute in the future 😄

PR #1037, opened 2022-10-31
Let additional arguments pass from db_compute.DBIConnection() to dbplyr_save_query()

rsund:
Tailored backends may want to support additional arguments in `compute()`. Unfortunately `db_compute.DBIConnection` is not passing all arguments to `dbplyr_save_query()` and therefore additional arguments never reach the `sql_query_save()` -method.

For example, `overwrite` argument is not supported in dbplyr (#488). Adding it to tailored backend is more complicated than necessary as additional arguments are not passed through the whole chain from `compute()` to `sql_query_save`-method (see duckdb/duckdb#5115).

Fix seems to be easy and non harmful. If that is not the case and/or there is a reason for not letting arguments pass through, please close this PR.


hadley:
OTOH there's no other place for `...` to go here, and adding it seems generally harmless.

PR #1060, opened 2022-12-01
`fill()` supports `"updown"`/`"downup"` and ordering by non-numeric columns

### LINKED ISSUE #1057: Several related issues with `fill()`
`fill()` doesn't work as expected, see reprex.

``` r
library(tidyverse)
mydb <- DBI::dbConnect(RSQLite::SQLite(), "")
mydata <- tibble(id = 1:3, val = c(NA, "hello", NA))
DBI::dbWriteTable(mydb, "mydata", mydata)

# The message is a bit cryptic, it's not obvious that `arrange()` will do some
# magic since id is already sorted
# also, typo "determinstic"
dplyr::tbl(mydb, "mydata") %>% 
  tidyr::fill(val, .direction = "down")
#> Error in `tidyr::fill()`:
#> ✖ `.data` does not have explicit order.
#> ℹ Please use `arrange()` or `window_order()` to make determinstic.

# If I do use arrange, I get a warning that doesn't make sense for a R user
dplyr::tbl(mydb, "mydata") %>% 
  arrange(id) %>% 
  tidyr::fill(val, .direction = "down")
#> # Source:     SQL [3 x 2]
#> # Database:   sqlite 3.39.4 []
#> # Ordered by: id
#>      id val  
#>   <int> <chr>
#> 1     1 <NA> 
#> 2     2 hello
#> 3     3 hello

# window order doesn't exist, it's a {dbplyr} function, not obvious for {dplyr} user who never use {dbplyr} explicitly
dplyr::tbl(mydb, "mydata") %>% 
  window_order(id) %>% 
  tidyr::fill(val, .direction = "up")
#> Error in window_order(., id): could not find function "window_order"

# this works perfectly
dplyr::tbl(mydb, "mydata") %>% 
  dbplyr::window_order(id) %>% 
  tidyr::fill(val, .direction = "down") 
#> # Source:     SQL [3 x 2]
#> # Database:   sqlite 3.39.4 []
#> # Ordered by: id
#>      id val  
#>   <int> <chr>
#> 1     1 <NA> 
#> 2     2 hello
#> 3     3 hello

# however if I fill "up" the order is reversed
dplyr::tbl(mydb, "mydata") %>% 
  dbplyr::window_order(id) %>% 
  tidyr::fill(val, .direction = "up") 
#> # Source:     SQL [3 x 2]
#> # Database:   sqlite 3.39.4 []
#> # Ordered by: id
#>      id val  
#>   <int> <chr>
#> 1     3 <NA> 
#> 2     2 hello
#> 3     1 hello

# and updown and downup are not supported
dplyr::tbl(mydb, "mydata") %>% 
  dbplyr::window_order(id) %>% 
  tidyr::fill(val, .direction = "updown") 
#> Error in `tidyr::fill()`:
#> ! `.direction` must be one of "down" or "up", not "updown".
#> ℹ Did you mean "down"?

dplyr::tbl(mydb, "mydata") %>% 
  dbplyr::window_order(id) %>% 
  tidyr::fill(val, .direction = "downup") 
#> Error in `tidyr::fill()`:
#> ! `.direction` must be one of "down" or "up", not "downup".
#> ℹ Did you mean "down"?

# though it is achievable
dplyr::tbl(mydb, "mydata") %>% 
  dbplyr::window_order(id) %>% 
  tidyr::fill(val, .direction = "up") %>% 
  tidyr::fill(val, .direction = "down") 
#> # Source:     SQL [3 x 2]
#> # Database:   sqlite 3.39.4 []
#> # Ordered by: id
#>      id val  
#>   <int> <chr>
#> 1     1 hello
#> 2     2 hello
#> 3     3 hello
```

<sup>Created on 2022-11-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I think we need : 

* A better error message : ditch the reference to `arrange()`, something like : 

> Please use call `dbplyr::window_order()` before `fill()` to set an explicit row order.

* Fix row order after `.direction = "up"`
* Support "downup" and "updown" 
* A note in the tidyr doc, or a link to a help file for the tbl_lazy method, documenting the row order issue

mgirlich:
Closes #1057

### DIFF
--- a/R/verb-fill.R
+++ b/R/verb-fill.R
@@ -32,11 +32,11 @@
 #'     n_squirrels,
 #'     n_squirrels2,
 #'   )
-fill.tbl_lazy <- function(.data, ..., .direction = c("down", "up")) {
+fill.tbl_lazy <- function(.data, ..., .direction = c("down", "up", "updown", "downup")) {
   cols_to_fill <- tidyselect::eval_select(expr(c(...)), .data)
   cols_to_fill <- syms(names(cols_to_fill))
   order_by_cols <- op_sort(.data)
-  .direction <- arg_match0(.direction, c("down", "up"))
+  .direction <- arg_match0(.direction, c("down", "up", "updown", "downup"))
 
   if (is_empty(order_by_cols)) {
     cli_abort(
@@ -47,11 +47,15 @@ fill.tbl_lazy <- function(.data, ..., .direction = c("down", "up")) {
     )
   }
 
+  if (.direction == "updown") {
+    .data <- tidyr::fill(.data, !!!cols_to_fill, .direction = "up")
+    return(tidyr::fill(.data, !!!cols_to_fill, .direction = "down"))
+  } else if (.direction == "downup") {
+    .data <- tidyr::fill(.data, !!!cols_to_fill, .direction = "down")
+    return(tidyr::fill(.data, !!!cols_to_fill, .direction = "up"))
+  }
+
   if (.direction == "up") {
-    # `desc()` cannot be used here because one gets invalid SQL if a variable
-    # is already wrapped in `desc()`
-    # i.e. `desc(desc(x))` produces `x DESC DESC`
-    # but this implies that "up" does not work for sorting non-numeric columns!
     order_by_cols <- purrr::map(order_by_cols, swap_order_direction)
   }
 
--- a/R/verb-fill.R
+++ b/R/verb-fill.R
@@ -52,7 +52,7 @@ fill.tbl_lazy <- function(.data, ..., .direction = c("down", "up")) {
     # is already wrapped in `desc()`
     # i.e. `desc(desc(x))` produces `x DESC DESC`
     # but this implies that "up" does not work for sorting non-numeric columns!
-    order_by_cols <- purrr::map(order_by_cols, ~ quo(-!!.x))
+    order_by_cols <- purrr::map(order_by_cols, swap_order_direction)
   }
 
   dbplyr_fill0(
@@ -62,6 +62,18 @@ fill.tbl_lazy <- function(.data, ..., .direction = c("down", "up")) {
     order_by_cols = order_by_cols,
     .direction = .direction
   )
+}
+
+swap_order_direction <- function(x) {
+  if (is_quosure(x)) {
+    x <- quo_get_expr(x)
+  }
+
+  if (is_call(x, "desc")) {
+    call_args(x)[[1]]
+  } else {
+    expr(desc(!!x))
+  }
 }
 
 dbplyr_fill0 <- function(.con, .data, cols_to_fill, order_by_cols, .direction) {
--- a/R/verb-fill.R
+++ b/R/verb-fill.R
@@ -42,7 +42,7 @@ fill.tbl_lazy <- function(.data, ..., .direction = c("down", "up")) {
     cli_abort(
       c(
         x = "{.arg .data} does not have explicit order.",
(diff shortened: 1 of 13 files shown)


mgirlich:
@hadley It seems that our CI is broken

> unable to load shared object '/home/runner/work/_temp/Library/stringi/libs/stringi.so':
  libicui18n.so.66: cannot open shared object file: No such file or directory

There is a [stringi issue](https://github.com/gagolews/stringi/issues/486) for this problem which suggests to use `options(configure.args='--disable-pkg-config')`.

* should we configure this for our CI?
* do you know how to do this?

gaborcsardi:
The stringi binary on RSPM links to a newer ICU library now:
```
root@05d71cb5d9d1:~# ldd /root/R/x86_64-pc-linux-gnu-library/4.2/stringi/libs/stringi.so
	libicui18n.so.70 => /lib/x86_64-linux-gnu/libicui18n.so.70 (0x00000040018d2000)
```

You probably need to bump the cache. 

I'll see if we can detect and automate this somehow in `r-lib/actions`, unfortunately the system requirements don't really support this currently, so maybe not.

R/verb-fill.R
```diff
@@ -32,27 +32,31 @@
...
         x = "{.arg .data} does not have explicit order.",
-        i = "Please use {.fun arrange} or {.fun window_order} to make determinstic."
+        i = "Please use {.fun dbplyr::window_order} to make order explicit."
       )
     )
   }
 
+  if (.direction == "updown") {
+    .data <- tidyr::fill(.data, !!!cols_to_fill, .direction = "up")
+    return(tidyr::fill(.data, !!!cols_to_fill, .direction = "down"))
```
hadley, line 52: I think this would be slightly easier to read if you did

```R
.data <- tidyr::fill(...)
.data <- tidyr::fill(...)
return(.data)
```

R/verb-fill.R
```diff
@@ -64,6 +68,18 @@ fill.tbl_lazy <- function(.data, ..., .direction = c("down", "up")) {
   )
 }
 
+swap_order_direction <- function(x) {
+  if (is_quosure(x)) {
+    x <- quo_get_expr(x)
+  }
+
+  if (is_call(x, "desc")) {
```
hadley, line 76:
```suggestion
  if (is_call(x, "desc", n = 1)) {
```

Not very important, but protects against a weirder error if you have `desc()`.

man/complete.tbl_lazy.Rd
```diff
@@ -4,7 +4,7 @@
 \alias{complete.tbl_lazy}
 \title{Complete a SQL table with missing combinations of data}
 \usage{
-complete.tbl_lazy(data, ..., fill = list())
```
hadley, line 7: I think something idiosyncratic about your setup made this change. To make it work persistently (and for me) you need to use a hack:

```
Roxygen: {library(tidyr); list(markdown = TRUE)}
```

hadley (approved):

hadley:
@gaborcsardi I bumped the cache but it didn't help 😞

gaborcsardi:
Oh, actually Ubuntu did not update the icu version within the same Ubuntu version, so that's good. You just need to remove this line:
https://github.com/tidyverse/dbplyr/actions/runs/3609657832/workflow#L29
because with that line you are installing packages for an older Ubuntu version.

hadley:
@gaborcsardi oh good catch, thanks!

mgirlich:
@gaborcsardi Thanks a lot for your help here 😄

PR #1065, opened 2022-12-06
fix `fill()` for Presto

### LINKED ISSUE #1026: fill() does not work for Presto
Similar to #651, Presto/Trino expects `IGNORE NULLS` to be outside the closing parenthesis of the window function, e.g., `LAST_VALUE()`. Currently `fill()` throws an error like

```r
conn <- dbConnect(Presto(), catalog = "foo", schema = "default", user = "bar")
my_tbl <- tbl(conn, "my_tbl")
my_tbl |>
  window_order(x) |>
  fill(y)
#> Error in stop.with.error.message(content) : 
#>  Query abc123 failed: line 5:6: mismatched input 'IGNORE'. Expecting: ')', ',', 'ORDER'
```

A similar solution to #651 works here, too (with RPresto providing the driver).

```r
last_value_sql.PrestoConnection <- function(con, x) {
  build_sql("LAST_VALUE(", ident(as.character(x)), ") IGNORE NULLS", con = con)
}
```

Happy to submit a PR if helpful.

copernican:
Fixes #1026.

I tested this using a live Presto connection. I'm not sure if there's a good way to add an automated test without adding a lot of code to simulate such a connection, but happy to go that route if you prefer.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dbplyr (development version)
 
+* `fill()` now works with Presto (@copernican, #1065).
+
 * `last()` is now correctly translated when no window frame is specified
   (@mgirlich, #1063).
 


hadley:
Can you please file this in https://github.com/prestodb/RPresto?

PR #1073, opened 2022-12-09
Always use 1 and 0 as literals for logicals in SQL Server

### LINKED ISSUE #934: No TRUE literal in SQL Server, should `escape()` default to `list(clause = "SELECT")` ?
MSSQL has no `TRUE` literal, but it's returned from translations: https://stackoverflow.com/a/7171264/946850.

dm uses a home-grown routine to insert records into MS SQL Server, via `dbplyr::escape()` . This breaks with dbplyr >= 2.2.0.

The reprex illustrates the problem. What's a good solution?

``` r
library(dbplyr)

con <- simulate_mssql()

# Buggy, no TRUE literal: https://stackoverflow.com/a/7171264/946850
escape(TRUE, con = con)
#> <SQL> TRUE

# This is what dm needs:
escape(c(TRUE, FALSE), parens = FALSE, collapse = NULL, con = con)
#> <SQL> TRUE
#> <SQL> FALSE

# Still buggy:
translate_sql_(list(TRUE), con = con)
#> <SQL> TRUE

# Correct, but how to achieve for a vector?
translate_sql_(list(TRUE), con = con, context = list(clause = "SELECT"))
#> <SQL> 1

# Useless for dm:
translate_sql_(list(c(TRUE, FALSE)), con = con, context = list(clause = "SELECT"))
#> <SQL> (1, 0)
```

<sup>Created on 2022-07-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

CC @TSchiefer.

krlmlr:
Closes #934.


tests/testthat/_snaps/backend-mssql.md
```diff
@@ -160,7 +160,7 @@
       <SQL>
```
hadley, line 160: Can you update this test please? (since escaping no longer depends on context)

PR #1074, opened 2022-12-09
Support `join_by()`

mgirlich:
Support for the awesome new `join_by()` clause 😄


R/join-by-compat.R
```diff
@@ -0,0 +1,96 @@
+new_join_by <- function(exprs, condition, filter, cross, x, y) {
+  out <- list(
+    exprs = exprs,
+    condition = condition,
+    filter = filter,
+    cross = cross,
```
DavisVaughan, line 6: Note that we removed `cross` from `join_by()` altogether in favor of people using `cross_join()` instead

mgirlich: Thanks for the info!
I strongly prefer `cross_join()` over the `join_by()` variant 😄

R/backend-sqlite.R
```diff
@@ -155,10 +155,15 @@ sql_query_join.SQLiteConnection <- function(con, x, y, vars, type = "inner", by
     x = by$y,
     y = by$x,
     x_as = by$y_as,
-    y_as = by$x_as
+    y_as = by$x_as,
+    condition = by$condition
   )
 
   if (type == "full") {
+    if (any(by$condition != "==")) {
```
hadley, line 163: Can remove after #1150.

R/join-cols-compat.R
```diff
@@ -0,0 +1,175 @@
...
+check_join_vars <- function(vars,
+                            names,
+                            condition,
+                            input,
+                            ...,
+                            error_call = caller_env()) {
+  check_dots_empty0(...)
+
+  if (!is.character(vars)) {
+    message <- glue("Join columns in `{input}` must be character vectors.")
```
hadley, line 81: When does this error arise? Is it an internal error?

R/join-cols-compat.R
```diff
@@ -0,0 +1,175 @@
...
+    abort(message, call = error_call)
+  }
+
+  na <- is.na(vars)
+  if (any(na)) {
+    bullets <- c(
```
hadley, line 87: Should these be `cli::cli_abort()`?

R/lazy-join-query.R
```diff
@@ -65,7 +65,7 @@ join_check_vars <- function(vars, call) {
 }
 
 join_check_by <- function(by, call) {
-  if (!vctrs::vec_is_list(by)) {
+  if (!vctrs::vec_is_list(by) && !inherits(by, "dplyr_join_by")) {
     # TODO use `cli_abort()` after https://github.com/r-lib/rlang/issues/1386
     # is fixed
     abort("`by` must be a list", .internal = TRUE)
```
hadley, line 71: I think since you're touching the line above you could switch to `cli_abort()` since the bug is fixed.

R/verb-joins.R
```diff
@@ -252,36 +288,63 @@ anti_join.tbl_lazy <- function(x,
...
                      sql_on = NULL,
                      auto_index = FALSE,
                      x_as = NULL,
                      y_as = NULL,
                      call = caller_env()) {
+  x_names <- tbl_vars(x)
+  y_names <- tbl_vars(y)
+  if (identical(by, character()) && is.null(sql_on)) {
+    if (type != "cross") {
+      lifecycle::deprecate_soft(
```
hadley, line 308: Should specifically mention this in the NEWS

R/verb-joins.R
```diff
@@ -639,25 +711,15 @@ join_two_table_alias <- function(names, from) {
...
   vctrs::vec_assert(x, character(), size = 2, arg = "suffix", call = call)
   list(x = x[1], y = x[2])
 }
 
-add_suffixes <- function(x, y, suffix) {
-  if (identical(suffix, "")) {
-    return(x)
+check_join_by_supported <- function(by, call = caller_env()) {
+  if (any(by$filter != "none")) {
+    cli_abort("Rolling joins aren't supported on database backends.", call = call)
```
hadley, line 723: It seems likely we'll need to relax this on a per backend basis at some point, but I don't think we need to do it prospectively.

tests/testthat/test-verb-joins.R
```diff
@@ -76,13 +76,13 @@ test_that("joins with non by variables gives cross join", {
   df1 <- memdb_frame(x = 1:5)
   df2 <- memdb_frame(y = 1:5)
 
-  out <- collect(inner_join(df1, df2, by = character()))
-  expect_equal(nrow(out), 25)
+  expect_snapshot({
+    (expect_warning(out_inner <- collect(inner_join(df1, df2, by = character()))))
```
hadley, line 80: IMO it's not worth adding the `expect_warning()` here

tests/testthat/test-verb-joins.R
```diff
@@ -76,13 +76,13 @@ test_that("joins with non by variables gives cross join", {
   df1 <- memdb_frame(x = 1:5)
```
hadley, line 76: Test name needs an update maybe should be moved to live next to the other cross join tests?

R/query-join.R
```diff
@@ -284,7 +284,8 @@ sql_join_tbls <- function(con, by, na_matches = "never") {
         sql_expr_matches(sql(lhs[[i]]), sql(rhs[[i]]), con = con)
       })
     } else {
-      compare <- paste0(lhs, " = ", rhs)
+      by$condition[by$condition == "=="] <- "="
```
hadley, line 287: This is the only real change to the SQL generation right? Pretty much everything else is improved input validation and new handling of output columns?

hadley (approved):

R/verb-joins.R
DavisVaughan: I don't think there are many backends that support ASOF style joins though. None of the common ones like Postgres, SQLite, MS SQL Server, etc

R/join-cols-compat.R
mgirlich: Probably, it is an internal error. I basically only copied over the file from dplyr https://github.com/tidyverse/dplyr/blob/main/R/join-cols.R.

R/verb-joins.R
mgirlich: But at least there are usually some workarounds to get the same results. In case I need a rolling join in the database I might just implement the workaround in dbplyr 😄

PR #1086, opened 2023-01-02
Remove field `last_op`

mgirlich:
The field `last_op` was introduced when switching to the lazy query structure but it was only meant as a temporary helper for optimisations. With some minor modifications it isn't needed anymore. It is a breaking change but I don't expect other packages to actually break.

`revdepcheck()` confirms that there are no new problems.

### DIFF
--- a/tests/testthat/_snaps/verb-arrange.md
+++ b/tests/testthat/_snaps/verb-arrange.md
@@ -37,10 +37,6 @@
       ORDER BY `a`
     Code
       lf %>% arrange(a) %>% select(-a) %>% arrange(b)
-    Condition
-      Warning:
-      ORDER BY is ignored in subqueries without LIMIT
-      i Do you need to move arrange() later in the pipeline or use window_order() instead?
     Output
       <SQL>
       SELECT `b`
@@ -59,7 +55,6 @@
       <SQL>
       SELECT `b`
       FROM `df`
-      ORDER BY `a`
     Code
       # use order
       lf %>% arrange(a) %>% select(-a) %>% mutate(c = lag(b))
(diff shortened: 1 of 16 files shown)


mgirlich:
@hadley It is potentially a breaking change but it doesn't break anything according to `revdepcheck()`, so we could still merge this. What do you think?

tests/testthat/_snaps/verb-arrange.md
```diff
@@ -37,10 +37,6 @@
       ORDER BY `a`
     Code
       lf %>% arrange(a) %>% select(-a) %>% arrange(b)
-    Condition
```
hadley, line 40: Why does this warning go away? Because there is an arrange that's eventually applied?

hadley (approved):
I think it's fine to make this change; I'd consider it change to the internals that's fine to make at this point since it doesn't break any revdeps.

tests/testthat/_snaps/verb-arrange.md
mgirlich: It goes away because there is no subquery anymore. It is still shown when there actually is a subquery, e.g. 

``` r
lf <- lazy_frame(a = 1, b = 1)
lf |> 
  arrange(a) |> 
  mutate(c = cumsum(b)) |> 
  filter(c == 1) |> 
  arrange(b)
#> <SQL>
#> SELECT *
#> FROM (
#>   SELECT *, SUM(`b`) OVER (ORDER BY `a` ROWS UNBOUNDED PRECEDING) AS `c`
#>   FROM `df`
#> ) `q01`
#> WHERE (`c` = 1.0)
#> ORDER BY `b`
```

<sup>Created on 2023-01-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #1097, opened 2023-01-12
Restrict table alias length to avoid truncation

### LINKED ISSUE #1096: Restrict table alias length
Databases have limits on the length of a table alias (e.g [Postgres has 63](https://stackoverflow.com/questions/27865770/how-long-can-postgresql-table-names-be)). The table alias should therefore be shortened to not exceed this length but still be unique.

``` r
library(dplyr)

con <- DBI::dbConnect(RPostgres::Postgres())

nm1 <- strrep("a", 63)
DBI::dbWriteTable(con, nm1, tibble(x = 1))
df1 <- tbl(con, nm1)

nm2 <- strrep("b", 63)
DBI::dbWriteTable(con, nm2, tibble(x = 1))
df2 <- tbl(con, nm2)

left_join(df1, df1) |> show_query()
#> Joining, by = "x"
#> <SQL>
#> SELECT "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_LHS"."x" AS "x"
#> FROM "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" AS "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_LHS"
#> LEFT JOIN "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" AS "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_RHS"
#>   ON ("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_LHS"."x" = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_RHS"."x")
left_join(df1, df1)
#> Joining, by = "x"
#> NOTICE:  identifier "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_LHS" will be truncated to "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
#> 
#> NOTICE:  identifier "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_LHS" will be truncated to "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
#> NOTICE:  identifier "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_RHS" will be truncated to "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
#> NOTICE:  identifier "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_LHS" will be truncated to "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
#> NOTICE:  identifier "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_RHS" will be truncated to "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
#> Error in `collect()`:
#> Caused by error:
#> ! Failed to prepare query: ERROR:  table name "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" specified more than once
```

<sup>Created on 2023-01-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

fh-mthomson:
Fixes #1096 

It was my first time working with snapshot testing (very cool!) but my approach was to review / accept the proposed changes per `devtools::test()` messaging only but noting that it updated some unrelated sections (e.g., `can explain`) in case that's undesired.


fh-mthomson:
@mgirlich I've opened #1097 with a proposal to address this. 

It was my first time working with snapshot testing (very cool!) but my approach was to review / accept the proposed changes per `devtools::test()` messaging only but noting that it updated some unrelated sections (e.g., `can explain`) in case that's undesired.

mgirlich:
Thanks for your PR!
Do you actually have to use such long table names that you stumbled over this issue? 😱 

So far, this only solves parts of the issue. In case of more than two joins there could still be issues with the length of the table names, e.g.

``` r

lf <- lazy_frame(x = 1, .name = "who_uses_such_long_table_names_as_they_are_hard_to_write")
lf2 <- lazy_frame(x = 1, .name = "some_other_table")

left_join(lf, lf, by = "x") %>% 
  left_join(lf2, by = "x")
#> <SQL>
#> SELECT `who_uses_such_long_table_names_as_they_are_hard_to_write...1`.`x` AS `x`
#> FROM `who_uses_such_long_table_names_as_they_are_hard_to_write` AS `who_uses_such_long_table_names_as_they_are_hard_to_write...1`
#> LEFT JOIN `who_uses_such_long_table_names_as_they_are_hard_to_write` AS `who_uses_such_long_table_names_as_they_are_hard_to_write...2`
#>   ON (`who_uses_such_long_table_names_as_they_are_hard_to_write...1`.`x` = `who_uses_such_long_table_names_as_they_are_hard_to_write...2`.`x`)
#> LEFT JOIN `some_other_table`
#>   ON (`who_uses_such_long_table_names_as_they_are_hard_to_write...1`.`x` = `some_other_table`.`x`)
```

<sup>Created on 2023-01-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

If you want you can try to fix this (probably in `generate_join_table_names()`).

tests/testthat/_snaps/backend-mysql.md
```diff
@@ -75,16 +75,3 @@
       SET `df_x`.`c` = COALESCE(`df_x`.`c`, `...y`.`c`), `df_x`.`d` = `...y`.`d`
       RETURNING `df_x`.`a`, `df_x`.`b` AS `b2`
 
-# can explain
```
mgirlich, line 78: The snapshots are missing because you don't have a live database. Can you please add them back?
There is this nice git feature that lets you commit only part of a file which is very helpful for such cases.

R/verb-joins.R
```diff
@@ -614,10 +614,7 @@ join_two_table_alias <- function(names, from) {
   if (from[1] != "as" && from[2] != "as") {
     # self join of named table
     if (!is.na(names[1]) && !is.na(names[2]) && identical(names[1], names[2])) {
-      out <- c(
```
mgirlich, line 617: I think I decided to use this pattern because it is slightly more expressive than the generic LHS and RHS.
If we want to keep this alternative possibilities would be:

* abbreviate the table names e.g. by using `abbreviate()` (though I'm not quite convinced about its results...)
* use `substr()` to shorten the string and then append LHS and RHS

R/lazy-join-query.R
```diff
@@ -196,15 +196,26 @@ sql_build.lazy_multi_join_query <- function(op, con, ...) {
...
+    table_names_prepared <- table_names$name
+  } else{
+    table_names_prepared <- join_two_table_alias(table_names$name, table_names$from)
   }
 
-  join_two_table_alias(table_names$name, table_names$from)
+  # avoid database aliases exceeding the database-specific maximum length
+  table_names_prepared %>%
+    abbreviate(
+      # arbitrarily floor at 63 (Postgres limit) to avoid unnecessarily truncating reasonable lengths
```
fh-mthomson, line 213: Not ideal, but I couldn't think of a better way to _not_ do use an arbitrary floor.

For example, without, then: `name_LHS` and `name_RHS` were getting truncated to something like `nLHS` and `nRHS` which isn't informative / necessary.

R/verb-joins.R
fh-mthomson: Thank you for the context and suggestions; I agree. I went with `abbreviate()` based on it seeing like the clearest path forward. Hopefully the options I specified add a bit of robustness, while recognizing we don't expect this case to be hit very often.

fh-mthomson:
@mgirlich thank you for the review! I've pushed what I think is a more robust approach based on your great feedback. let me know what you think!

> Do you actually have to use such long table names that you stumbled over this issue? 😱

Unfortunately, I'm dealing with tables already materialized by another team as `{long_project_id_[1, 2, 3, ...}_{very_descriptive_table_name}`. Not ideal, but it works! 😅

mgirlich:
We could consider creating shorter names by abbreviate but I think this should be done in a different PR. I also noticed that `abbreviate()` loves underscores a bit too much

``` r
abbreviate("abbreviate_does_like_underscores_too_much")
#> abbreviate_does_like_underscores_too_much 
#>                                    "a___"
```

<sup>Created on 2023-02-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/lazy-join-query.R
```diff
@@ -199,15 +199,26 @@ sql_build.lazy_multi_join_query <- function(op, con, ...) {
 }
 
 generate_join_table_names <- function(table_names) {
+  table_name_length_max <- dplyr::coalesce(max(nchar(table_names$name)), 0)
```
hadley, line 202: Why would `max(nchar(table_names$name))` be NA?

R/lazy-join-query.R
```diff
@@ -199,15 +199,26 @@ sql_build.lazy_multi_join_query <- function(op, con, ...) {
...
+    table_names_prepared <- table_names$name
+  } else{
+    table_names_prepared <- join_two_table_alias(table_names$name, table_names$from)
   }
 
-  join_two_table_alias(table_names$name, table_names$from)
+  # avoid database aliases exceeding the database-specific maximum length
+  abbreviate(
+    table_names_prepared,
+    # arbitrarily floor at 63 (Postgres limit) to avoid unnecessarily truncating reasonable lengths
```
hadley, line 216: Can please you add a link to the postgres docs for this in the comment?

R/lazy-join-query.R
```diff
@@ -199,15 +199,26 @@ sql_build.lazy_multi_join_query <- function(op, con, ...) {
...
+    minlength = max(table_name_length_max, 63),
+    # (explicit, default) err on the side of unique table names instead of exact length
```
hadley, line 218: I don't understand what this comment is saying. Can you clarify?

tests/testthat/_snaps/verb-joins.md
```diff
@@ -40,6 +40,37 @@
...
+          `a012345678901234567890123456789012345678901234567890123456789012_L`.`y` = `a012345678901234567890123456789012345678901234567890123456789012_R`.`y`
+        )
+
+---
+
+    Code
+      self_join3 %>% remote_query()
+    Output
+      <SQL> SELECT
+        `a012345678901234567890123456789012345678901234567890123456789012...1`.`x` AS `x`,
```
hadley, line 62: That table name is 68 characters long?

tests/testthat/test-verb-joins.R
```diff
@@ -72,6 +72,42 @@ test_that("join works with in_schema", {
...
+  mf2 <- tbl(con, nm2)
+
+  # 2 tables
+  self_join2 <- left_join(mf1, mf1, by = c("x", "y"))
+
+  expect_equal(
+    self_join2 %>% collect(),
+    tibble(x = 1:3, y = "a")
+  )
+  expect_snapshot(
```
hadley, line 94: Given that you want to make specific assertions about the length of the generate table names, is it possible to avoid the snapshot tests in favour of something more specific?

tests/testthat/test-verb-joins.R
```diff
@@ -72,6 +72,42 @@ test_that("join works with in_schema", {
...
+  mf1 <- tbl(con, nm1)
+
+  nm2 <- paste0("b", paste0(0:62 %% 10, collapse = ""))
+  DBI::dbWriteTable(con, nm2, tibble(x = 2:3, y = "b"))
+  mf2 <- tbl(con, nm2)
+
+  # 2 tables
+  self_join2 <- left_join(mf1, mf1, by = c("x", "y"))
+
+  expect_equal(
```
hadley, line 90: Did this fail before? If so, the comment at top of test should refer to SQLite, not postgres

tests/testthat/test-verb-joins.R
```diff
@@ -72,6 +72,42 @@ test_that("join works with in_schema", {
   )
 })
 
+test_that("alias truncates long table names at database limit", {
+  # Postgres has max table name length of 63; ensure it's not exceeded when generating table alias
+  withr::local_db_connection(con <- dbConnect(RSQLite::SQLite(), ":memory:"))
```
hadley, line 77:
```suggestion
  con <- withr::local_db_connection(dbConnect(RSQLite::SQLite(), ":memory:"))
```

R/lazy-join-query.R
fh-mthomson: Currently, there can be cases where `NA` values are found in `table_names$name`. One such example (see below) is when joining an already-joined lazy query with a net new table.

I'm not familiar enough with the SQL composition to know whether each value of `name` _should_ be non-`NA`, but I suspect the current behavior is expected. @mgirlich any thoughts?

Example: after setting `table_name_length_max <- max(nchar(table_names$name))`:
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(dbplyr)
#> 
#> 
#>     ident, sql
lf <- dbplyr::lazy_frame(x = 1, a = 1, .name = "df1")
lf2 <- dbplyr::lazy_frame(x = 1, b = 2, .name = "df2")
lf3 <- dbplyr::lazy_frame(x = 1, b = 2, .name = "df3")

df_join_1 <- left_join(lf, lf2, by = "x")

right_join(df_join_1, lf3, by = "x")
#> Error in if (minlength <= 0L) {: missing value where TRUE/FALSE needed
```

<sup>Created on 2023-02-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Stepping inside `generate_join_table_names()`:

```
Browse[1]> table_names
# A tibble: 2 × 2
  name  from  
  <chr> <chr> 
1 NA    ""    
2 df3   "name"
```

R/lazy-join-query.R
fh-mthomson: added!

R/lazy-join-query.R
fh-mthomson: There is seemingly a fundamental limitation to `abbreviate()` which does not allow for both (a) max length to be binding and (b) names to be guarenteed unique.

From the documentation:
```
strict | logical: should minlength be observed strictly? Note that setting strict = TRUE may return non-unique strings.
```

I've opted for `strict = FALSE` to preserve uniqueness (since `dbplyr` will require it any), but there is a non-zero risk that aliases still exceed the database limit. As one mitigation, I've updated the `method` to `both.sides` which tends to add a bit more scrambling (slightly less obvious alias names) but the intended length is more likely to be aligned with the requested length.

I've spent a decent amount of time playing with combinations of args (including discussing with ChatGPT) but couldn't come up with a fully-error proof solution using `abbreviate` _or_ an alternate function.

tests/testthat/_snaps/verb-joins.md
fh-mthomson: After using `method = "both.sides"` in `abbreviate()`, this case is no longer an issue. This is a good example where `method = "left.keep"` preserves too much of the LHS at the expense of having unnecessary length.

tests/testthat/test-verb-joins.R
fh-mthomson: Good thinking! Added some to look at the actual values from `generate_join_table_names()`; this catches the error observed above.

tests/testthat/test-verb-joins.R
fh-mthomson: The motivating example were it was failing was on Postgres; updated the testing backend to Postgres instead.

tests/testthat/test-verb-joins.R
fh-mthomson: This was carried forward from `join works with in_schema` test. Not applicable now using Postgres backend, so will revert

hadley (approved):
Thanks for working on this!

R/lazy-join-query.R
mgirlich: In the current implementation this is expected but it's probably better to get rid of `NA` in a separate PR.

PR #1106, opened 2023-01-20
`select()` updates `order_var`

### LINKED ISSUE #1103: window_order crashing when deleting the variable that I am ordering for
Before when using window_order I could do something for example:

```r
x <- table_x %>% window_order(variable1) %>% .... %>% select(-variable1)
```

Now when the variable that you ordered for is not present it crashes.

Example with dbplyr 2.0.0:
working

Now with the new version dbplyr 2.3.0:
not_working

As far as I know there is no way to delete the ordering so I am stuck and I have to use the previous version. 
Is this something on purpose? Then how we can solve this error?

mgirlich:
Fixes #1103.
I had to slightly adapt my initial proposal to allow symbols and `desc(<symbol>)`. Note that this is only enforced in `window_order()` but not in `add_order()`. Without this the query of `slice_sample()` would have been quite a bit more complicated. It is a bit of a hack but as `add_order()` is an internal function this should be fine.


tests/testthat/_snaps/verb-arrange.md
```diff
@@ -60,7 +60,7 @@
       lf %>% arrange(a) %>% select(-a) %>% mutate(c = lag(b))
     Output
       <SQL>
-      SELECT `b`, LAG(`b`, 1, NULL) OVER (ORDER BY `a`) AS `c`
+      SELECT `b`, LAG(`b`, 1, NULL) OVER () AS `c`
```
mgirlich, line 63: This is a logical consequence of removing `a`. I think this makes more sense than the old behaviour.

tests/testthat/test-verb-select.R
```diff
@@ -71,6 +71,24 @@ test_that("select preserves grouping vars", {
...
+  # can drop order vars
+  expect_equal(lf %>% window_order(y) %>% select(-y) %>% op_sort(), list())
+  expect_equal(lf %>% window_order(desc(y)) %>% select(-y) %>% op_sort(), list())
+  # can rename order vars
+  expect_equal(lf %>% window_order(y) %>% select(y2 = y) %>% op_sort(), list(expr(y2)))
+  expect_equal(
+    lf %>% window_order(desc(y)) %>% select(y2 = y) %>% op_sort(),
+    list(expr(desc(y2)))
+  )
+  # keeps sort order
```
mgirlich, line 85: I think the order of variables shouldn't be touched by `select()`

tests/testthat/_snaps/verb-arrange.md
hadley: In the following code, I'm surprised we lose the `ORDER BY` in both cases. I think it makes sense to be able order a window function by a variable that doesn't appear in the final output. It's particularly surprising that we lose the `ORDER BY` in the window function, but not in the whole query.

``` r
library(dbplyr)

lf <- lazy_frame(a = 1, b = 2, c = 3)

lf %>% window_order(a) %>% select(-a) %>% mutate(c = lag(b))
#> <SQL>
#> SELECT `b`, LAG(`b`, 1, NULL) OVER () AS `c`
#> FROM `df`
lf %>% arrange(a) %>% select(-a) %>% mutate(c = lag(b))
#> <SQL>
#> SELECT `b`, LAG(`b`, 1, NULL) OVER () AS `c`
#> FROM `df`
#> ORDER BY `a`
```

<sup>Created on 2023-01-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

mgirlich: The first query works by switching `select()` and `mutate()`

```r
lf %>% window_order(a) %>% mutate(c = lag(b)) %>% select(-a)
#> <SQL>
#> SELECT `b`, LAG(`b`, 1, NULL) OVER (ORDER BY `a`) AS `c`
#> FROM `df`
```

So, it is possible to order by a variable not in the final output.

How would you handle the following query?

```r
lf %>% window_order(a) %>% select(-a) %>% mutate(c = lag(b)) %>% mutate(c2 = lag(c))
```

This now produces:

```sql
<SQL>
SELECT *, LAG(`c`, 1, NULL) OVER () AS `c2`
FROM (
  SELECT `b`, LAG(`b`, 1, NULL) OVER () AS `c`
  FROM `df`
) `q01`
```

If we keep the window order for the inner `LAG()` we should keep it for the outer `LAG()`. This would imply adding `a` to the inner `SELECT` and then removing it in the outer `SELECT`. It should be doable but definitely complicates things.
Therefore, I would argue that you should use `mutate() %>% select(-a)` instead of `select(-a) %>% mutate()`.

mgirlich: > It's particularly surprising that we lose the ORDER BY in the window function, but not in the whole query.

It is a bit surprising but it also makes sense because we "finish" the `arrange()` operation before removing `a`.

mgirlich:
@hadley You're now happy with the PR and only forgot to approve? Or do you need to take a closer look? 😉

tests/testthat/test-verb-select.R
hadley: Agreed

R/verb-select.R
```diff
@@ -166,6 +167,22 @@ rename_groups <- function(lazy_query, vars) {
   lazy_query
 }
 
+rename_order <- function(lazy_query, vars) {
+  old2new <- set_names(names(vars), vars)
+  order <- op_sort(lazy_query)
+
+  desc <- purrr::map_lgl(order, ~ is_call(.x, "desc", n = 1L))
```
hadley, line 174: Calling this `is_desc` would be slightly more descriptive

R/verb-slice.R
```diff
@@ -131,9 +131,10 @@ slice_by <- function(.data, order_by, size, .by, with_ties = FALSE) {
     )
   }
 
-  out <- .data %>%
-    window_order({{order_by}}) %>%
-    filter(!!window_fun) %>%
+  # must use `add_order()` as `window_order()` only allows variables
+  .data$lazy_query <- add_order(.data, quos({{order_by}}))
```
hadley, line 135: Does any existing test cover this? What happens if there's a select after this step?

hadley (approved):

R/verb-select.R
```diff
@@ -166,6 +167,22 @@ rename_groups <- function(lazy_query, vars) {
...
+  order <- purrr::map_if(order, desc, ~ call_args(.x)[[1L]])
+  order <- purrr::map_chr(order, as_name)
```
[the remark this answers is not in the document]
mgirlich: I don't quite follow. `check_window_order_dots()` ensures that each element of `order` is either a symbol or `desc(<symbol>)`.  And these two steps extract the underlying symbol and convert it to a string so that it can be mapped more easily.

R/verb-slice.R
mgirlich: The previous window order is restored below and there is a test for this. A `select()` after this step would drop the window order in case of `slice_sample()` and keep the window order for the others (unless the window order variables are de-selected).

R/verb-select.R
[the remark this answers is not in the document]
hadley: When is an element of `order` not already going to be a name?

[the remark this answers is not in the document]
mgirlich: After line 175 every element of `order` is a symbol. Then `as_name()` converts the symbols to strings.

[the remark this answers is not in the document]
mgirlich: I just looked into the documentation

> `rlang::as_name()` is the _opposite_ of `base::as.name()`

which is indeed quite mean 😄

PR #1112, opened 2023-01-30
Refactor checks

mgirlich:
This PR

* adds the rlang compat type checks
* adds some custom type checks powered by rlang's `stop_input_type()`
* adds `check_unsupported_arg()`
* replaces most `stopifnot()`
* removes dependency on assertthat package


mgirlich:
@hadley I think it is sufficient if you skim through the error messages, in particular over `check_unsupported_arg()`.

R/translate-sql-string.R
```diff
@@ -5,19 +5,17 @@
...
     sql_call2(f, x, start, length)
   }
 }
 
-check_integer <- function(x, arg) {
-  if (length(x) != 1 || !is.numeric(x)) {
-    cli_abort("{.arg {arg}} must be a single number")
-  }
-  as.integer(x)
+check_integer <- function(x, arg = caller_arg(x), call = caller_env()) {
```
hadley, line 16: I think you should give this a different prefix since it also does coercion (which we're generally moving away from in `check_` functions)

hadley (approved):
Looks like a great improvement! Thanks for working on it 😄

PR #1126, opened 2023-02-05
fix `compute` for Teradata

### LINKED ISSUE #147: Compute does not work with teradata
calling `compute` on a teradata connection does not work and throws the following error:

```
  nanodbc/nanodbc.cpp:1344: 42000: [Teradata][ODBC Teradata Driver][Teradata Database](-3707)Syntax error, expected something like a 'METHOD' keyword between the 'CREATE' keyword and the 'TEMPORARY' keyword. 
```

I think you need to use the keyword `VOLATILE ` instead of `TEMPORARY` for Teradata right?

But2ene:
This PR was created to share some fix for Teradata used at work. It should fix #147

Indexes must be passed at table creation.

Happy to correct code style / design choices / etc...

### DIFF
--- a/R/backend-teradata.R
+++ b/R/backend-teradata.R
@@ -30,6 +30,95 @@ dbplyr_edition.Teradata <- function(con) {
 }
 
 #' @export
+sql_table_index.Teradata <- function(con, table, columns, name = NULL,
+                                        unique = FALSE, ...) {
+
+  message("sql_table_index.Teradata do nothing.")
+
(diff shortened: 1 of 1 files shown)


mgirlich:
@hadley Any idea how you want to handle this? Returning `"select 1"` in `sql_table_index.Teradata()` doesn't feel right. It should probably error as it doesn't work anyway.

hadley:
@But2ene if this is still troubling you, I'd recommend starting by filing an issue with reprex so we can look into the underlying cause in more detail.

PR #1195, opened 2023-03-14
Fix MySQL & MariaDB types (numeric => DOUBLE, POSIXct => DATETIME)

### LINKED ISSUE #1189: TIMESTAMP is incorrect CAST type for MySQL & MariaDB
Similar to: https://github.com/tidyverse/dbplyr/pull/1011

DATETIME is the correct cast type: 
https://mariadb.com/kb/en/cast/
https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#function_cast

avsdev-cw:
Fixes #1189


mgirlich:
Thanks! I added a NEWS entry that mentions the fix and your name 😄

PR #1196, opened 2023-03-14
Add dots to generics

### LINKED ISSUE #919: Add ... to all generics
Adding dots after methods have been implemented in other packages is a source of a major pain, learned in pillar/tibble. Those methods are not exported for now, but it's easy to forget. Alternatively, a comment to remind us to add dots would help too.

https://github.com/tidyverse/dbplyr/blob/8798b00cc8ba103c3779a8f0e652f16e9cc0a81f/R/verb-copy-to.R#L308-L311

mgirlich:
Closes #919.

This PR adds `...` to these generics:

* `db_connection_describe()`
* `db_table_temporary()`
* `sql_expr_matches()`

and checks for empty dots in most methods with dots.

### DIFF
--- a/R/backend-hana.R
+++ b/R/backend-hana.R
@@ -80,5 +80,6 @@ db_table_temporary.HDB <- function(con, table, temporary, ...) {
 
 #' @export
 sql_values_subquery.HDB <- function(con, df, types, lvl = 0, ...) {
+  check_dots_empty0(...)
   sql_values_subquery_union(con, df, types = types, lvl = lvl, from = "DUMMY")
 }
--- a/R/backend-hana.R
+++ b/R/backend-hana.R
@@ -74,6 +74,8 @@ db_table_temporary.HDB <- function(con, table, temporary, ...) {
 `sql_table_analyze.HDB` <- function(con, table, ...) {
   # CREATE STATISTICS doesn't work for temporary tables, so
   # don't do anything at all
+
+  check_dots_empty0(...)
 }
 
 #' @export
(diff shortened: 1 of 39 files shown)


mgirlich:
@hadley I wasn't sure whether you disagree with adding `...` to the three functions or whether you might see an issue in checking for empty (or in a few places used) dots.

R/backend-hana.R
```diff
@@ -57,7 +57,8 @@ sql_translation.HDB <- function(con) {
 
 # nocov start
 #' @export
-db_table_temporary.HDB <- function(con, table, temporary) {
+db_table_temporary.HDB <- function(con, table, temporary, ...) {
+  check_dots_empty0(...)
```
hadley, line 61: FWIW I think it's technically incorrect to check for empty dots in an S3 method because it means that a subclass that calls a parent method with `NextMethod()` can't add any additional arguments, somewhat defeating the point of using `...` in the first place.

That's why we have `check_dots_used()`, which must be called from the generic.

mgirlich: I now remember there was a discussion about `check_dots_used()` vs `check_dots_empty()`, e.g. https://github.com/tidyverse/dplyr/issues/6599.
But the reason to use `check_dots_empty()` instead of `check_dots_used()` was mostly due to tidyselect. I guess for most generics in this PR it wouldn't make sense to add a tidyselect argument. So should I switch to `check_dots_used()` in the generic instead?

hadley: Yeah, I think that’s a bit safer.

mgirlich:
Closed in favour of #1222.

PR #1215, opened 2023-03-17
Add argument checks in some basic SQL functions

mgirlich:
They are pretty basic functions that so far didn't check their arguments at all

* `sql_table_index()`
* `sql_query_fields()`
* `sql_query_save()`
* `sql_query_wrap()`
* `sql_query_rows()`
* `sql_query_explain()`

* `db_copy_to()`
* `db_compute()`
* `db_write_table()`

### DIFF
--- a/R/db-sql.R
+++ b/R/db-sql.R
@@ -196,6 +196,8 @@ sql_query_explain.DBIConnection <- function(con, sql, ...) {
 #' @rdname db-sql
 #' @export
 sql_query_fields <- function(con, sql, ...) {
+  check_table_ident(sql, sql = TRUE)
+
   UseMethod("sql_query_fields")
 }
 #' @export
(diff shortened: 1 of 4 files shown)


R/db-sql.R
```diff
@@ -150,7 +150,18 @@ sql_table_analyze.DBIConnection <- function(con, table, ...) {
...
 #' @export
-sql_table_index <- function(con, table, columns, name = NULL, unique = FALSE, ...) {
+sql_table_index <- function(con,
+                            table,
+                            columns,
+                            name = NULL,
+                            unique = FALSE,
+                            ...,
+                            call = caller_env()) {
+  check_table_ident(table, call = call)
```
mgirlich, line 160: A table ident can be anything like `ident()`, `ident_q()`, `in_schema()`, `in_catalog()`, `DBI::Id()` or a string.

R/db-sql.R
```diff
@@ -185,6 +186,7 @@ sql_table_index.DBIConnection <- function(con,
 #' @rdname db-sql
 #' @export
 sql_query_explain <- function(con, sql, ...) {
+  check_scalar_sql(sql)
```
mgirlich, line 189: Only scalar SQL or a string make sense here.

Issue #1219, opened 2023-03-20
Access uses `<>` for inequality, not !=


erikvona:
Hi all,

As said in the title, Access uses `<>` for inequality, and throws a syntax error for `!=`. This should be added to the translation.

``` r
library(odbc)
library(DBI)
library(dplyr, warn.conflicts = FALSE)
library(dbplyr, warn.conflicts = FALSE)

file_path <- "J:\\ExampleDb.accdb"
con <- dbConnect(odbc(), driver = "Microsoft Access Driver (*.mdb, *.accdb)", dbq = file_path)

t1 <- tbl(con, "Table1") %>%
  filter(ID != 1) %>%
  collect()
#> Error in `collect()`:
#> ! Failed to collect lazy table.
#> Caused by error:
#> ! nanodbc/nanodbc.cpp:1509: 42000: [Microsoft][ODBC Microsoft Access Driver] Syntax error (missing operator) in query expression '(`ID` != 1.0)'. 
#> Backtrace:
#>     ▆
#>  1. ├─tbl(con, "Table1") %>% filter(ID != 1) %>% collect()
#>  2. ├─dplyr::collect(.)
#>  3. └─dbplyr:::collect.tbl_sql(.)
#>  4.   └─base::tryCatch(...)
#>  5.     └─base (local) tryCatchList(expr, classes, parentenv, handlers)
#>  6.       └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
#>  7.         └─value[[3L]](cond)
#>  8.           └─cli::cli_abort("Failed to collect lazy table.", parent = cnd)
#>  9.             └─rlang::abort(...)
```

<sup>Created on 2023-03-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #1220, opened 2023-03-20
Add != to <> translation for Access (#1219)

erikvona:
As noted in the issue and title, add != to <> translation for Access since that's what Access uses


mgirlich:
Thanks a lot! 😄

PR #1222, opened 2023-03-21
Check dots used

### LINKED ISSUE #919: Add ... to all generics

mgirlich:
Closes #919.

### DIFF
--- a/R/tbl-sql.R
+++ b/R/tbl-sql.R
@@ -13,6 +13,7 @@
 #'   Mainly useful for better performance when creating
 #'   multiple `tbl` objects.
 tbl_sql <- function(subclass, src, from, ..., vars = NULL) {
+  check_dots_used()
   # If not literal sql, must be a table identifier
   from_sql <- as.sql(from, con = src$con)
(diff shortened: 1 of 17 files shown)


mgirlich:
This PR doesn't but #1223 should.

R/tbl-sql.R
```diff
@@ -13,6 +13,7 @@
 #'   Mainly useful for better performance when creating
 #'   multiple `tbl` objects.
 tbl_sql <- function(subclass, src, from, ..., vars = NULL) {
+  check_dots_used()
```
krlmlr, line 16: This is not a generic, is this correct?

I'm now seeing failures with duckdb here.

PR #1223, opened 2023-03-21
Fix method consistency for `sql_select.DBIConnection()`

mgirlich:
To fix the CRAN check results.

### DIFF
--- a/R/db-sql.R
+++ b/R/db-sql.R
@@ -365,9 +365,9 @@ sql_select.DBIConnection <- function(con,
                                      order_by = NULL,
                                      limit = NULL,
                                      distinct = FALSE,
-                                     ...,
-                                     subquery = FALSE) {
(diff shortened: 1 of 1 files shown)


mgirlich:
@hadley this should fix the CRAN check results.

hadley:
Thanks! Ok to do a patch release later today?

mgirlich:
Yes, but there are two breaking changes:

* `sql_random()` is now deprecated. I used `lifecycle::deprecate_warn()` but this might need to change to `lifecycle::deprecate_soft()` then.
* The `vars` argument of `translate_sql()` has been removed after it threw an error for the last 7 years.

mgirlich:
This is probably a bit easier. Thanks!

hadley:
Now let me remember how to do that 😆

Issue #1205, opened 2023-03-15
Improve on how to create table identifiers


mgirlich:
Overall, they severely lack vector properties

* the class is lost when subsetting

``` r
ident("a", "b")[1]
#> [1] "a"
```

* combining changes the class to sql and quotes with backticks

``` r
c(ident("a"), ident_q("b"))
#> <SQL> `a`
#> <SQL> b
```

* `in_schema()` and `in_catalog()` can be coerced to a length 2 resp. 3 character which already caused a couple of issues

``` r
paste0("Can't write to table: ", in_schema("schema", "table"))
#> [1] "Can't write to table: schema" "Can't write to table: table"
paste0("Can't write to table: ", in_catalog("catalog", "schema", "table"))
#> [1] "Can't write to table: schema"  "Can't write to table: table"  
#> [3] "Can't write to table: catalog"
```

* wrapping doesn't work nicely. It simply quotes (and always uses backticks as quotes) and it isn't idempotent e.g. 

``` r
ident(ident(ident("a")))
#> <IDENT> ```a```
```

Then there is `as.sql()` which is used to:

* escape a table name, e.g. `as.sql(table, con = con)` (this is the most common use)
* in `in_schema()` and `in_catalog()`
* in `win_over()` to prepare `partition` and `order`. This doesn't even feel like the right use.

Overall, the handling of table identifiers is too fiddly and difficult to get right which caused a couple of issues like repeated quoting or issues with incorrect handling of `in_schema()` or `in_catalog()`.
The table identifiers should be build upon vctrs and the issues above be fixed this way.

Open questions:

* Is it necessary to accept arbitrary SQL?
* Is support for quoted identifiers needed? And how exactly? Can e.g. the schema be quoted and the table not? Or should it be possible to pass in a fully qualified name `schema.table`?

Challenges:

* Combining e.g. `in_schema()` and `ident()` doesn't make much sense with the current class system. It would make more sense to have a single class for table identifiers instead. This would obviously be a big breaking change.
* A common class should be build upon record objects. Unfortunately, they don't support names so a workaround is needed.

PR #1228, opened 2023-03-23
Use custom `glue_sql()`

mgirlich:
This PR adds a customised form of `glue_sql()` which is inspired by the cli syntax. It supports these inline prefixes:

* `tbl`: a table identifier; a string, <ident>, <ident_q>, schema, catalog, or <Id>
* `col`: a string, <ident>, <ident_q>. We could/should also add a qualified column class.
* `name`: a table or column alias, must be a string.
* `from`: a table identifier or a subquery.
* `sql`: SQL code
* `kw`: an SQL keyword to style.

Overall, this makes the code more concise, explains the purpose of interpolated elements better and makes it a bit easier to consistently support schema, catalog and Id objects.


R/backend-mssql.R
```diff
@@ -475,10 +475,10 @@ mssql_version <- function(con) {
...
   # https://stackoverflow.com/q/16683758/946850
-  build_sql(
-    "SELECT * INTO ", as.sql(name, con), " ",
-    "FROM (\n  ", sql, "\n) AS temp",
-    con = con
+  glue_sql2(
+    "SELECT * INTO {.tbl name} ",
+    "FROM (\n  {.sql sql}\n) AS temp",
+    .con = con
   )
```
hadley, line 482:
```suggestion
  glue_sql2("SELECT * INTO {.tbl name} FROM (\n  {.sql sql}\n) AS temp", .con = con)
```
?

R/build-sql.R
```diff
@@ -46,3 +46,101 @@ build_sql <- function(..., .env = parent.frame(), con = sql_current_con()) {
...
+      value <- text
+      type <- "raw"
+    }
+    value <- eval(parse(text = value, keep.source = FALSE), envir)
+    glue_check_collapse(type, should_collapse)
+
+    if (type == "tbl") {
+      if (is_bare_character(value)) {
+        value <- ident(value)
+      }
```
hadley, line 102: I think it's a bit weird to have multiple transformers that do the same thing. Would it be better to just use `name` for all of them?

Do we already have a `as.ident()` helper?

R/build-sql.R
```diff
@@ -46,3 +46,101 @@ build_sql <- function(..., .env = parent.frame(), con = sql_current_con()) {
   pieces <- purrr::map_chr(enexprs(...), escape_expr, con = con)
   sql(paste0(pieces, collapse = ""))
 }
+
+glue_sql2 <- function(...,
```
hadley, line 50: Can we delete `build_sql()` now?

R/db-sql.R
```diff
@@ -217,10 +216,11 @@ sql_query_save <- function(con, sql, name, temporary = TRUE, ...) {
 }
 #' @export
 sql_query_save.DBIConnection <- function(con, sql, name, temporary = TRUE, ...) {
-  build_sql(
+  glue_sql2(
```
hadley, line 219: Fit on one line now?

R/translate-sql-conditional.R
```diff
@@ -81,7 +81,7 @@ sql_case_match_clause <- function(f, x, con) {
 
 
 sql_if <- function(cond, if_true, if_false = quo(NULL), missing = quo(NULL)) {
-  out <- build_sql("CASE WHEN ", enpar(cond), " THEN ", enpar(if_true))
+  out <- "CASE WHEN {enpar(cond)} THEN {enpar(if_true)}"
```
hadley, line 84: Do all the `enpar()`ing up front?

R/backend-mssql.R
```diff
@@ -250,7 +250,7 @@ simulate_mssql <- function(version = "15.0") {
 
       `[` = function(x, i) {
         i <- with_mssql_bool(i)
-        build_sql("CASE WHEN (", i, ") THEN (", x, ") END")
+        glue_sql2("CASE WHEN ({i}) THEN ({x}) END", .con = sql_current_con())
```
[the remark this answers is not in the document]
mgirlich: I considered defaulting to `sql_current_con()` but I was against it for a couple of reasons:

* defaulting later to `sql_current_con()` felt much less work than the other way round
* it is a bit magic and can easily lead to subtle bugs
* I'd rather not do it in case we decide to later export `glue_sql2()`

But having it as the first argument sounds reasonable. Especially, because you rarely ever would like to pipe into it.

R/build-sql.R
mgirlich: In the current implementation this is indeed pretty weird. But the idea for this PR came from my attempt to improve on the handling of table identifiers (see #1205 for some more background). During the implementation of table identifiers build on top of vctrs, I had some issues figuring out where to use the new table identifier and in general what type an argument to e.g. `build_sql()` should be.
This motivated me to be much more explicit about types in `glue_sql2()`. It makes the intent more clear and allows to add some checks on `value` in a future PR.

PR #1229, opened 2023-03-24
Remove arg `src` from `tbl_lazy()`

### LINKED ISSUE #1208: Remove `src` argument of `tbl_lazy()
It was deprecated a long time ago.

mgirlich:
Closes #1208.
@hadley Or would you prefer first going to `deprecate_stop()`? I would have guessed `src` wasn't used much anymore that's why I removed it directly.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dbplyr (development version)
 
+* Removed argument `src` of `tbl_lazy()` after it has been deprecated for years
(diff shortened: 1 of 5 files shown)


hadley:
I think it's fine to remove. But I'd keep a test in, just to demonstrate that you still get an informative error.

mgirlich:
Are you fine with the generic empty dots error message or would you prefer a specific error message for `src`?

Issue #1146, opened 2023-02-10
Add `use_star` and `use_table_name` in `show_query()`


mgirlich:
* `use_star`: should query use `*` in the `SELECT` statement or not? `TRUE` by default
* `use_table_name`: should query prefix every column by the origin table in joins? `FALSE` by default

also add these arguments to `remote_query()`.

These arguments are useful when using dbplyr to create SQL.

PR #1266, opened 2023-04-28
Generate `SELECT` sql for joins in `sql_build()`

mgirlich:
Some time ago we added the feature to use `*` for column selection where possible. For select queries this was done in `sql_build()`, for join queries in `sql_render()`. This is inconsistent and I think it makes more sense to do this in `sql_build()`. This also has the advantage that it makes `sql_render()` a bit simpler for joins.
In the end this makes it easier to implement #1146 (this allows the user to control whether to use `*` or write out all columns).

### DIFF
--- a/R/db-sql.R
+++ b/R/db-sql.R
@@ -410,12 +410,11 @@ sql_select.DBIConnection <- function(con,
 sql_query_join <- function(con,
                            x,
                            y,
-                           vars,
+                           select,
                            type = "inner",
                            by = NULL,
                            na_matches = FALSE,
                            ...,
-                           select = NULL,
                            lvl = 0) {
   check_dots_used()
   UseMethod("sql_query_join")
@@ -424,12 +423,11 @@ sql_query_join <- function(con,
 sql_query_join.DBIConnection <- function(con,
                                          x,
                                          y,
-                                         vars,
+                                         select,
                                          type = "inner",
                                          by = NULL,
                                          na_matches = FALSE,
                                          ...,
-                                         select = NULL,
                                          lvl = 0) {
   JOIN <- switch(
     type,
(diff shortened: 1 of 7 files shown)


R/lazy-join-query.R
```diff
@@ -245,10 +245,19 @@ sql_build.lazy_rf_join_query <- function(op, con, ...) {
...
+    vars = vars_classic,
+    x_as = by$x_as,
+    y_as = by$y_as
+  )
+
   join_query(
     sql_optimise(sql_build(op$x, con), con),
     sql_optimise(sql_build(op$y, con), con),
     vars = vars_classic,
+    select = select,
```
mgirlich, line 260: Theoretically, we could remove `vars` but this would be a breaking change. But it might also be quite hard to properly deprecate `vars`. Should we just accept the breaking change? I don't expect many methods for `sql_render.join_query()` apart from the ones in dbplyr.

R/db-sql.R
```diff
@@ -442,7 +444,9 @@ sql_query_join.DBIConnection <- function(con,
   x <- dbplyr_sql_subquery(con, x, name = by$x_as, lvl = lvl)
   y <- dbplyr_sql_subquery(con, y, name = by$y_as, lvl = lvl)
 
-  select <- sql_rf_join_vars(con, type = type, vars, x_as = by$x_as, y_as = by$y_as)
+  if (is.null(select)) {
```
mgirlich, line 447: I added this hack because removing `vars` is a breaking change.

R/lazy-join-query.R
hadley: I think it's probably fine to just break it, but you might want to double check affected packages with this search: https://github.com/search?type=code&auto_enroll=true&q=org%3Acran+sql_render+path%3ANAMESPACE.

hadley (approved):

PR #1283, opened 2023-05-11
Clean up methods

mgirlich:
This PR removes methods that shouldn't have been added I think, e.g. `sql_build.ident()` feels wrong to me. Also this forced us to also having to add methods for <dbplyr_schema> and <dbplyr_catalog> which makes the handling really error prone. Instead, I added a class <base_query> which makes things much simpler.


MichaelChirico:
We have this in a test, what are we doing wrong?

```r
con <- dbplyr::simulate_dbi("MyDBIConnection")

sql <- dbplyr::sql("select * from testtable ;")
formatted_sql <- dbplyr::db_sql_render(con = con, sql)
```

It now breaks due to missing `sql_render.sql`, but I don't see anything in the documentation to steer me back to shore.

MichaelChirico:
Maybe more context is useful, this is part of a "sanity test" in the test suite to make sure basic SQL rendering is WAI:

```r
test_that("db_sql_render formats sql queries", {
  con <- dbplyr::simulate_dbi("MyDBIConnection")

  sql <- dbplyr::sql("select * from testtable ;")
  formatted_sql <- dbplyr::db_sql_render(con = con, sql)

  # We check here if the formatter returns a specific string. This might
  # break if the formatter will change the output format in the future. However,
  # we hope that the query is simple ennough so that its formatted version will
  # never change.
  expected_sql <- dbplyr::sql("SELECT * FROM testtable;")
  expect_identical(formatted_sql, expected_sql)
})
```

This is done on the raw `sql()` object because this `con` is a virtual class that does not have a `dbSendQuery()` method (so trying to render the simple `tbl(con, sql)` object is not possible).

We could distribute this test into the downstream package which actually has a `dbSendQuery()` method, but this is a more central place for such a sanity check.

PR #1284, opened 2023-05-12
Add <dbplyr_table_ident>

### LINKED ISSUE #1139: Check handling of schemas
There were some bugs with schemas e.g. #1133 or #938, so it might be worth looking for other issues with schemas.

### LINKED ISSUE #1204: Remove `dbi_quote()`
It is only used in `db_write_table.DBIConnection()`.

### LINKED ISSUE #1205: Improve on how to create table identifiers

### LINKED ISSUE #1261: Pass `DBI::Id()` to `DBI::dbWriteTable()`
Extracted from #1099.

For odbc connections, dbplyr currently calls `dplyr::db_write_table` (effectively `DBI::dbWriteTable` / `odbc:::odbc_write_table`) with a plain string as the `name` argument (`dbplyr_003`, etc).  On backends with many schemas and tables, this leads to very bad performance (~15 minutes of hang for me), as iteration over all schemas and tables is needed to find a match (via `odbcConnectionColumns`).

When an `Id` is supplied as the `name` argument to `odbc::odbc_write_table` instead, its catalog, schema, and table names are [passed to `odbc::odbcConnectionColumns`, and ultimately `SQLColumns` on the driver](https://github.com/r-dbi/odbc/blob/111cedfe2aacb1ca4f87016814fb7d295f7dfb3f/R/Connection.R#L145), enabling much faster execution.

I think it's important, then, that an `Id` encapsulating the schema and table names is passed to `db_write_table` in `db_copy_to.DBIConnection`.

### LINKED ISSUE #1264: Don't use unnecessary alias when using tables in schema
``` r

lf1 <- lazy_frame(tibble(x = 1), .name = in_schema("sch", "tab1"))
lf2 <- lazy_frame(tibble(x = 1), .name = in_schema("sch", "tab2"))
left_join(lf1, lf2, by = "x")
#> <SQL>
#> SELECT `tab1`.`x` AS `x`
#> FROM `sch`.`tab1` AS `tab1`
#> LEFT JOIN `sch`.`tab2` AS `tab2`
#>   ON (`tab1`.`x` = `tab2`.`x`)
```

<sup>Created on 2023-04-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

mgirlich:
Currently, there are multiple classes that can be used to create a sql tbl:

* `<ident>`
* `<dbplyr_schema>`
* `<dbplyr_catalog>`
* `<Id> (from DBI)`
* `<sql>`
* `<SQL>`

This is quite messy, needs lots of testing and lots of extra code to support every case. Further, these classes have some annoying properties (e.g. see #1205). This PR improves the situation by adding a new class <dbplyr_table_ident> which replaces these classes (at least internally for now).

Its using `vctrs::rcrd()` under the hood. In most cases we don't really need (or want) anything but a single table but it might still be nice to have the vector properties.

There are a couple of questions though:

1. `in_schema()` and `in_catalog()` support <ident_q>, e.g. `in_schema(ident_q("a b"), "table")` gives `` <SCHEMA> a b.`table` ``. I don't think it is used often and I don't really see the need for it. Supporting this would make the implementation more complicated.
2. Should we deprecate `<dbplyr_schema>` and `<dbplyr_catalog>`? We use it internally anyway, so package authors might need to adapt to it anyway. Also, they would likely face the same issues we had with the schema and catalog classes.
3. I would love to get rid of the `sql` field in the table ident class. Unfortunately, it is needed to support `<ident_q>`. Could we deprecate `<ident_q>` for table identifiers? I don't really see a good use case for them. I think they rather make problems than provide actual benefit.

Closes #1204. Closes #1205. Closes #1261. Closes #1139. Closes #1264.


mgirlich:
@hadley What's your opinion here?

hadley:
Basic idea seems reasonable to me. Answer to questions below:

1. As long as there's someway to opt-out of quoting, no need to keep `ident_q()`
2. I'm fine with it; I've never found the idea of giving special names to various levels of namespacing terribly compelling.
3. No strong feelings and no recollection of why it was set up this was in the first place, so as long as (1) is handled, then I'm happy.

mgirlich:
> As long as there's someway to opt-out of quoting, no need to keep `ident_q()`

At which detail level to you want to be able to opt out of quoting?

1. Either quote everything (i.e. catalog, schema, and table) or nothing, e.g.

```
new_table_ident("catalog", "schema", "table", quote = TRUE)
#> "catalog"."schema"."table"
new_table_ident("catalog", "schema", "table", quote = FALSE)
#> catalog.schema.table
```

2. Can toggle quoting for catalog, schema, and table separately, e.g.

```
new_table_ident("catalog", "schema", "table")
#> "catalog"."schema"."table"
new_table_ident("catalog", "schema", "table", quote_schema = FALSE)
#> "catalog".schema."table"
```

3. Should it be possible to provide a qualified table name, e.g.

```
new_table_ident(sql("catalog.schema.table"))
```

The 3rd case doesn't really need to be implemented separately as it can be covered in both cases, but I think it might affect the user interface.

By the way, what's your use case to provide already quoted identifiers?

hadley:
All I want is someway to provide a string that dbplyr will leave as is — this is just a useful escape hatch if we hit some weird database or weird table name where our normal approach doesn't work. So in terms of the user interface, option 3 is what I was thinking of (although I'm not tied to `sql()`).

mgirlich:
@hadley You can skim over this PR to get a rough idea on how `as_table_ident()` affects the code.
My main questions are:

* do we want to export the `<dbplyr_table_ident>` class? (together with `new/as_table_ident()`)
* It would be nice to get rid of the classes `<dbplyr_catalog>`, `<dbplyr_schema()>` and `<ident_q>` in the future. But its not easy to get rid of a class. Maybe we could do that by deprecating `escape.dbplyr_schema/dbplyr_catalog/ident_q()`. What do you think?

R/db-io.R
```diff
@@ -161,15 +161,15 @@ db_write_table.DBIConnection <- function(con,
...
+  table <- as_table_ident(table)
   check_character(types, allow_null = TRUE)
   check_named(types)
   check_bool(temporary)
 
   tryCatch(
     dbWriteTable(
       con,
-      name = dbi_quote(table, con),
+      name = table_ident_to_id(table),
```
mgirlich, line 172: Before we passed a quoted string of class `<SQL>`. It has benefits to instead pass an `<Id>` object (see #1261). As I had to adapt this anyway, I thought I might just change it directly.

R/db-io.R
```diff
@@ -187,11 +187,6 @@ db_write_table.DBIConnection <- function(con,
 
 # Utility functions ------------------------------------------------------------
 
-dbi_quote <- function(x, con) UseMethod("dbi_quote")
```
mgirlich, line 190: `dbi_quote()` basically wasn't used anywhere.

R/table-ident.R
```diff
@@ -0,0 +1,239 @@
...
+                            quoted = FALSE,
+                            error_call = caller_env()) {
+  check_dots_empty(call = error_call)
+
+  check_character(table, call = error_call)
+  check_character(schema, call = error_call)
+  check_character(catalog, call = error_call)
+  check_logical(quoted, call = error_call)
+
+  n <- vctrs::vec_size_common(
```
mgirlich, line 14: It is not really necessary to support vectors but it might be handy at some point.

R/table-ident.R
```diff
@@ -0,0 +1,244 @@
...
+
+inform_unwanted_table_ident <- function(f) {
+  cli::cli_inform(
+    c(
+      i = "Using {.fn {f}} for a table identifier is intended as fallback in case of bugs.",
+      i = "If you need it to work around a bug please open an issue {.url https://github.com/tidyverse/dbplyr/issues}."
+    ),
+    .frequency = "regularly",
+    .frequency_id = paste0("dbplyr_", f)
+  )
```
mgirlich, line 102: This should hopefully push people to avoid using `ident_q()` and `sql()` unless necessary.

R/tbl-sql.R
```diff
@@ -15,13 +15,9 @@
...
   check_character(vars, allow_null = TRUE)
-  from_sql <- as.sql(from, con = src$con)
-  vars <- vars %||% dbplyr_query_fields(src$con, from_sql)
 
-  # If not literal sql, must be a table identifier
-  if (!(is.ident(from) || is.sql(from) || is_schema(from) || inherits(from, "Id"))) {
-    from <- as.sql(from, con = src$con)
-  }
+  from <- as_from(from)
+  vars <- vars %||% dbplyr_query_fields(src$con, from)
```
mgirlich, line 20: This is one of the examples that show how complicated the handling was before.

hadley:
Sorry, I lost the notification. I'll add it to my to do list and try and take a look this week.

R/table-ident.R
```diff
@@ -0,0 +1,239 @@
...
+  names <- vctrs::field(x, "name")
+  if (all(is.na(names))) {
+    return(NULL)
+  }
+
+  names
+}
+
+#' @export
+`names<-.dbplyr_table_ident` <- function(x, value) {
```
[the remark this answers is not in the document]
krlmlr: I like the "alias" idea. Using names seems awkward.

tests/testthat/_snaps/verb-compute.md
```diff
@@ -21,7 +21,7 @@
       df %>% compute(name = in_schema("main", "db1"), temporary = FALSE)
     Condition
       Error in `db_save_query.DBIConnection()`:
-      ! Can't save query to table `main`.`db1`.
+      ! Can't save query to table main.db1.
```
krlmlr, line 24: Is this intended?

krlmlr (commented):
Great idea! I'll watch this space to make the necessary changes in dm. Is there a fallback that works in CRAN and dev dplyr?

tests/testthat/_snaps/verb-compute.md
mgirlich: Yes, this is on purpose:

* this might help to disencourage people to use `format()` when they should actually use `escape()`
* the backticks aren't the correct way to escape names in many databases. So, the error message might confuse people that dbplyr didn't escape the table identifier correctly.

R/tbl-sql.R
hadley: `as_from()` doesn't need the connection because escaping is now handled later?

R/verb-joins.R
```diff
@@ -666,8 +666,8 @@ add_semi_join <- function(x,
     c(x_alias$name, y_alias$name),
     c(x_alias$from, y_alias$from)
   )
-  by$x_as <- ident(by$x_as)
-  by$y_as <- ident(by$y_as)
+  by$x_as <- by$x_as
+  by$y_as <- by$y_as
```
hadley, line 670:
```suggestion
```

tests/testthat/_snaps/verb-compute.md
hadley: We do generally prefer to use some indicator of the start and end of the string because it's not uncommon to accidentally add a leading or trailing space, which is then really hard to discover.

R/backend-postgres-old.R
```diff
@@ -15,17 +15,19 @@ db_write_table.PostgreSQLConnection <- function(con,
...
-
+  table <- as_table_ident(table)
   if (!isFALSE(temporary)) {
     cli_abort(c(
       "RPostgreSQL backend does not support creation of temporary tables",
       i = "Either set {.code temporary = FALSE} or switch to {.pkg RPostgres}"
     ))
   }
 
+  # RPostgreSQL doesn't handle `Id()` or `SQL()` correctly, so we can only pass
```
hadley, line 26: Do we want to warn about this? Or do we think so few people still use RPostgreSQL that it's not worth it?

R/build-sql.R
```diff
@@ -97,14 +97,10 @@ sql_quote_transformer <- function(connection) {
     glue_check_collapse(type, should_collapse)
 
     if (type == "tbl") {
-      if (is_bare_character(value)) {
-        value <- ident(value)
-      }
+      value <- as_table_ident(value)
     } else if (type == "from") {
       # TODO maybe this could call dbplyr_sql_subquery()
```
hadley, line 102: Remove this todo?

R/db-sql.R
```diff
@@ -245,30 +248,20 @@ sql_query_wrap.DBIConnection <- function(con, from, name = NULL, ..., lvl = 0) {
...
-}
-
-as_subquery_name <- function(x, default = ident(unique_subquery_name())) {
-  if (is.ident(x)) {
-    x
-  } else if (is.null(x)) {
-    default
-  } else {
-    ident(x)
+  if (identical(table_ident_name(from), name)) {
```
hadley, line 261: What case does this catch?

R/table-ident.R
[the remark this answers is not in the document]
hadley: Yeah, I like the alias idea too.

hadley (approved):

R/db-sql.R
mgirlich: This was to prevent an unnecessary table alias. I now moved this into `escape()` where it probably makes a bit more sense.

R/tbl-sql.R
mgirlich: 💯

PR #1288, opened 2023-05-26
Filtering for MS SQL Consistent with dplyr

### LINKED ISSUE #1233: MSSQL Incorrect Translation of Boolean and row_number while Filtering Zero Rows
I'm trying to filter all rows from a table to handle certain edge cases where I want to pass a table forward for use in joins but just have it be empty if necessary. Typically, a `filter(FALSE)` is all that I need to do, but when connected to an MS SQL Server via `odbc` and the `freetds` driver, there's some translation hiccups that prevent the two most common patterns for this from happening. 

My guess is that, `TRUE` and `FALSE` are being incorrectly translated to numerics instead of booleans? And the `row_number()` translation could probably rely on some sort of sensible default windowing scheme

Reprex:
``` r
suppressPackageStartupMessages(library(dplyr))
suppressPackageStartupMessages(library(dbplyr))

df <- tibble(a = 1:10) 

db_rs <- tbl_lazy(df = df, con = simulate_redshift())
db_sq <- tbl_lazy(df = df, con = simulate_sqlite())
db_ms <- tbl_lazy(df = df, con = simulate_mssql())

# Successful in their respective DBs
db_rs %>% filter(FALSE)
#> <SQL>
#> SELECT *
#> FROM `df`
#> WHERE (FALSE)
db_sq %>% filter(FALSE)
#> <SQL>
#> SELECT *
#> FROM `df`
#> WHERE (0)

# Fails in Microsft SQL Server 
db_ms %>% filter(FALSE)
#> <SQL>
#> SELECT *
#> FROM `df`
#> WHERE (0)
db_ms %>% filter(row_number() == -1)
#> <SQL>
#> SELECT `a`
#> FROM (
#>   SELECT *, ROW_NUMBER() OVER () AS `q01`
#>   FROM `df`
#> ) `q01`
#> WHERE (`q01` = -1.0)

# Workarounds
db_ms %>% filter(row_number(a) == -1)
#> <SQL>
#> SELECT `a`
#> FROM (
#>   SELECT
#>     *,
#>     CASE
#> WHEN (NOT((`a` IS NULL))) THEN ROW_NUMBER() OVER (PARTITION BY (CASE WHEN ((`a` IS NULL)) THEN 1 ELSE 0 END) ORDER BY `a`)
#> END AS `q02`
#>   FROM `df`
#> ) `q01`
#> WHERE (`q02` = -1.0)
db_ms %>% filter(1 != 1)
#> <SQL>
#> SELECT *
#> FROM `df`
#> WHERE (1.0 != 1.0)
```

<sup>Created on 2023-03-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Specifically, the two errors for the failing filter queries in my environment are: 
```
Error in `collect()`:
! Failed to collect lazy table.
Caused by error:
! nanodbc/nanodbc.cpp:1752: 00000: [FreeTDS][SQL Server]Statement(s) could not be prepared.  [FreeTDS][SQL Server]An expression of non-boolean type specified in a context where a condition is expected, near ')'. 
<SQL> 'SELECT TOP 11 *
FROM "dbo"."test_df"
WHERE (0)'
```
and 
```
Error in `collect()`:
! Failed to collect lazy table.
Caused by error:
! nanodbc/nanodbc.cpp:1752: 00000: [FreeTDS][SQL Server]Statement(s) could not be prepared.  [FreeTDS][SQL Server]The function 'ROW_NUMBER' must have an OVER clause with ORDER BY. 
<SQL> 'SELECT TOP 11
  "year",
  "month",
  "day",
  "dep_time",
  "sched_dep_time",
  "dep_delay",
  "arr_time",
  "sched_arr_time",
  "arr_delay",
  "carrier",
  "flight",
  "tailnum",
  "origin",
  "dest",
  "air_time",
  "distance",
  "hour",
  "minute",
  "time_hour"
FROM (
  SELECT *, ROW_NUMBER() OVER () AS "q01"
  FROM "dbo"."test_df"
) "q01"
WHERE ("q01" = -1.0)'
```

ejneer:
Fixes #1233 boolean translation issue

This PR would make MS SQL filtering behavior more consistent with dplyr/other databases. Currently, filter expressions with a symbol or atomic value that aren't a part of a logical expression will fail.

Both of these queries will fail because the WHERE clauses are invalid in T-SQL. The values must be a part of a comparison expression (e.g. 1 == 1).
``` r
tbl(con, "mtcars") |> filter(TRUE) |> show_query()
#> <SQL>
#> SELECT "mtcars".*
#> FROM "mtcars"
#> WHERE (1)

tbl(con, "mtcars") |> mutate(x = mpg < 15) |> filter(x) |> show_query()
#> <SQL>
#> SELECT "q01".*
#> FROM (
#>   SELECT "mtcars".*, CAST(IIF("mpg" < 15.0, 1, 0) AS BIT) AS "x"
#>   FROM "mtcars"
#> ) "q01"
#> WHERE ("x")
```

The general strategy of the PR is to examine all of the WHERE clauses of a query and look for:

+ Top level atoms/symbols (as above)
+ Atoms/symbols involved in an expression that would require a logical (e.g. `|`)

When either of these are found the values are explicitly cast and made part of an equality expression. All other expressions remain as-is.

``` r
tbl(con, "mtcars") |> filter(TRUE) |> show_query()
#> <SQL>
#> SELECT "mtcars".*
#> FROM "mtcars"
#> WHERE (cast(1 AS "BIT") = 1)

tbl(con, "mtcars") |> mutate(x = mpg < 15) |> filter(x) |> show_query()
#> <SQL>
#> SELECT "q01".*
#> FROM (
#>   SELECT "mtcars".*, CAST(IIF("mpg" < 15.0, 1, 0) AS BIT) AS "x"
#>   FROM "mtcars"
#> ) "q01"
#> WHERE (cast("x" AS "BIT") = 1)

tbl(con, "mtcars") |> filter(FALSE | TRUE | TRUE == TRUE) |> show_query()
#> <SQL>
#> SELECT "mtcars".*
#> FROM "mtcars"
#> WHERE (cast(0 AS "BIT") = 1 OR cast(1 AS "BIT") = 1 OR 1 = 1)
```


mgirlich:
Thanks for the PR. Do you think it would be possible to not make this part of `db_sql_render()` but part of the translation by using `sql_current_context()`? I think this could give a cleaner solution.

ejneer:
I don't think this can happen at the translation level, at least not how I'm accomplishing it here.  When `TRUE`/`FALSE` are escaped to 1/0 `sql_current_context()` can indicate that I'm in a where clause, but I have no information about the expressions that the `TRUE`/`FALSE` values may be a part of.  e.g. when escaping I can't tell the difference between the `TRUE`s in `filter(TRUE)` and `filter(TRUE = TRUE)`. Only the first case requires modification to work.

Modifying information available at escape time would require changing `translate_sql_` which would make the scope of this change much larger.

I'm happy to have another go at this issue though if this solution is too out of place however.

mgirlich:
Okay, now I understand the issue better. This reminds me a little of the hack I used in the translation of `-` and `+` in order to avoid unnecessary parentheses. You might want to have a look into this: https://github.com/tidyverse/dbplyr/blob/main/R/backend-.R#L30.

ejneer:
I see the similarity -- I could modify the translation of all the functions I'm handling [here](https://github.com/tidyverse/dbplyr/blob/3bad13f8b0cceda4779b746958a5f74419b7a945/R/backend-mssql.R#L600-L610)  to accomplish this PR.

Where I'm stuck are the base cases, `filter(TRUE)` and `filter(x)` -- These aren't part of any call and just get escaped unless I'm missing something.  Any ideas on where I can intercept these to modify them aside from rendering?

hadley:
@mgirlich given the general weirdness of SQL server bit/boolean semantics I'm ok with this approach and I think it's worth merging. Are you ok with that?

mgirlich:
Yes, sounds reasonable to me.

PR #1292, opened 2023-05-30
Retrocompatibility with Oracle prior to version 12

JeremyPasco:
Hi, sadly french public hospitals are full of very old Oracle DB instances (11.2 or even lower).
The current implementation of Oracle backend makes use of `FETCH` to limit the number of retrieved rows. `FETCH` has been introduced since v12.

The older way to limit rows was a simple `WHERE ROWNUM <= {limit}` which still works with recent Oracle versions. IMO, the main reason to use `FETCH` was to allow an offset, which is out of scope here. Then, could you turn back to the `ROWNUM` way please?


mgirlich:
Thanks for the PR. Usually it doesn't make much sense to support such old versions but this seemed like a small enough change.

hadley:
Unfortunately that SQL is not equivalent to the previous translation, as in #1436, so I think we might need to revert this change.

Issue #1305, opened 2023-06-15
Support `relationship` parameter in join verbs


bairdj:
dplyr added a `relationship` parameter to join verbs in v1.1.1, which throws an error if provided to dbplyr.

The issue I am experiencing is for some unit tests that run across both local and lazy tables. dplyr throws a warning if there is a many-to-many relationship, which can be silenced by providing the `relationship` parameter, but this then causes the lazy table tests to fail.

I don't think dbplyr necessarily needs to support the check (would probably have to be an extra query to check if it is violated), but it would be good if it would accept the parameter.

``` r
library(dplyr, warn.conflicts = FALSE)
#> Warning: package 'dplyr' was built under R version 4.2.3
library(dbplyr, warn.conflicts = FALSE)
#> Warning: package 'dbplyr' was built under R version 4.2.3

join_df <- tibble(
  Species = c("setosa", "versicolor", "virginica"),
  x = 1:3
)

# dplyr join
iris %>%
  inner_join(
    join_df,
    by = join_by(Species),
    relationship = "many-to-one"
  ) %>%
  slice_head(n = 5)
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species x
#> 1          5.1         3.5          1.4         0.2  setosa 1
#> 2          4.9         3.0          1.4         0.2  setosa 1
#> 3          4.7         3.2          1.3         0.2  setosa 1
#> 4          4.6         3.1          1.5         0.2  setosa 1
#> 5          5.0         3.6          1.4         0.2  setosa 1

# dbplyr join
iris_dbplyr <- tbl_lazy(iris, con = simulate_mssql())

iris_dbplyr %>%
  inner_join(
    join_df,
    by = join_by(Species),
    relationship = "many-to-one"
  ) %>%
  slice_head(n = 5)
#> Error in `inner_join()`:
#> ! `...` must be empty.
#> ✖ Problematic argument:
#> • relationship = "many-to-one"
#> Backtrace:
#>     ▆
#>  1. ├─... %>% slice_head(n = 5)
#>  2. ├─dplyr::slice_head(., n = 5)
#>  3. ├─dplyr::inner_join(., join_df, by = join_by(Species), relationship = "many-to-one")
#>  4. └─dbplyr:::inner_join.tbl_lazy(., join_df, by = join_by(Species), relationship = "many-to-one")
#>  5.   └─rlang::check_dots_empty()
#>  6.     └─rlang:::action_dots(...)
[...]

PR #1307, opened 2023-06-15
Support relationship in joins

bairdj:
Fix for #1305 

This follows the same conventions as the `multiple` and `unmatched` parameters, in that it supports the database default behaviour (allow many-to-many).

If the user attempts to provide a value for `relationship` other than `NULL` or `many-to-many`, an error will be thrown.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # dbplyr (development version)
 
-* `*_join()` now allows providing a `relationship` argument (@bairdj, #1305).
+* `*_join()` now allows using specifying the relationship argument. It must be `NULL` or `"many-to-many"` (@bairdj, #1305).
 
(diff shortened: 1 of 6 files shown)


bairdj:
No problem. I've made the update to news

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # dbplyr (development version)
 
+* `*_join()` now allows providing a `relationship` argument (@bairdj, #1305).
```
mgirlich, line 3: Maybe be a bit more specific what is supported:

```suggestion
* `*_join()` now allows using specifying the relationship argument. It must be `NULL` or `"many-to-many"` (@bairdj, #1305).
```

bairdj: Sounds good

mgirlich:
Thanks for the PR! 😄

PR #1310, opened 2023-06-21
Add `sql_options` argument

### LINKED ISSUE #1146: Add `use_star` and `use_table_name` in `show_query()`

mgirlich:
Fixes #1146.
As we now have 3 arguments that control the generated SQL I guess it makes more sense to put this into a single options argument.


R/sql-build.R
```diff
@@ -60,46 +66,49 @@ sql_render <- function(query,
...
                                   con = NULL,
                                   ...,
-                                  use_star = TRUE,
                                   subquery = FALSE,
-                                  lvl = 0,
-                                  cte = FALSE) {
-  qry <- sql_build(query, con = con, ..., use_star = use_star)
-  qry <- sql_optimise(qry, con = con, ..., subquery = subquery)
+                                  lvl = 0) {
+  # FIXME the handling of `...` seems wrong overall.
```
mgirlich, line 86: I didn't want to add `sql_options` to `sql_render()` as the options aren't needed in any other `sql_render()` method but only the `sql_build()` methods. This feels like a hack overall but is due to the fact that the SQL generation works like:

1. Call `sql_render()` on <tbl_lazy> resp. <lazy_query>
2. Build the lazy query via `sql_build()` (here the `sql_options` are needed; in there `sql_render()` is never called).
3. Optimise the resulting query via `sql_optimise()`.
4. Render the query via `sql_render()` (here the `sql_options` aren't needed).

NEWS.md
```diff
@@ -34,8 +34,13 @@
...
 
-* `show_query()` and `remote_query()` get the argument `use_star` that controls
-  whether to use `SELECT *` to select every column (@mgirlich, #1146).
+* `show_query()` and `remote_query()` get the argument `sql_options` that allows
+  to control how the SQL is generated. The `cte` argument is deprecated.
+  Currently, the following options are provided:
+  * use common table expressions?
+  * use `SELECT *` or explicitly select every column?
+  * qualify all columns in a join or only the ambiguous ones?
+  (@mgirlich, #1146).
```
hadley, line 43: Include argument names?

R/db.R
```diff
@@ -55,12 +55,92 @@ sql_join_suffix.DBIConnection <- function(con, suffix, ...) {
...
+      what = "db_sql_render(cte)",
+      with = I("db_sql_render(sql_options = dbplyr_sql_options(cte = TRUE))")
+    )
+    sql_options <- dbplyr_sql_options(cte = TRUE)
+  }
+
+  if (is.null(sql_options)) {
+    sql_options <- dbplyr_sql_options()
+
+    out <- db_sql_render(con, sql, sql_options = sql_options)
```
hadley, line 72: I almost wonder if it's better to do something like:

```R
db_sql_render <- function(con, sql, ..., cte = FALSE, sql_options = NULL) {
   ...
   db_sql_render2(con, sql, ..., sql_options = sql_options)
}

db_sql_render2 <- function(con, sql, ..., cte = FALSE, sql_options = NULL) {
  UseMethod("db_sql_render")
}
```

I think we also need either a `as_sql_options()` or `check_sql_options()` to ensure that `sql_option` is actually an list of the correct type.

R/db.R
```diff
@@ -55,12 +55,92 @@ sql_join_suffix.DBIConnection <- function(con, suffix, ...) {
...
+#' lf1 <- lazy_frame(key = 1, a = 1, b = 2)
+#' lf2 <- lazy_frame(key = 1, a = 1, c = 3)
+#'
+#' result <- left_join(lf1, lf2, by = "key") %>%
+#'   filter(c >= 3)
+#'
+#' show_query(result)
+#' sql_options <- dbplyr_sql_options(cte = TRUE, qualify_all_columns = TRUE)
+#' show_query(result, sql_options = sql_options)
+dbplyr_sql_options <- function(cte = FALSE, use_star = TRUE, qualify_all_columns = FALSE) {
```
hadley, line 107: Why not just `sql_options()`?

R/sql-build.R
hadley: Even given that, I think you're still better off having an explicit `sql_options` argument rather than fishing out of `...`

hadley (approved):
Nice change!

PR #1311, opened 2023-06-22
Require `con` in `translate_sql()`

### LINKED ISSUE #1209: `translate_sql()` should be more careful before falling back to `simulate_dbi()`
In non-interactive mode it should error if `con %||% sql_current_con()` is `NULL` and in interactive mode it should probably inform about this.

mgirlich:
Closes #1209.
While this adds a bit of boilerplate in the tests and the vignette, I think it is much safer this way.
This is a breaking change so we could also use a lifecycle deprecation but as it is hardly every intended to use `translate_sql()` without a connection I'm not sure we need it.


hadley (approved):
Yeah, this is a good change. I was just too lazy to make it earlier.

PR #1321, opened 2023-06-30
Nudge users away from `names()` and `$` on <tbl_lazy>

### LINKED ISSUE #1056: FR: Nudge users via thin wrapper methods for tbl_lazy methods of names(), $
Successor to #997, #1010.

That issue/PR was about defining a `names.tbl_lazy()` method that behaves more like what new users might expect out of running `names()` on a lazy table, namely that it returns the column names.

But as discussed (https://github.com/tidyverse/dbplyr/pull/1010#issuecomment-1322274125), this is a bit fraught because other basic R functions like `str()` induce some implicit requirements on consistency among `names()` and related operators like `$`, `[`, and `[[`, so we decided against creating such a method and instead forcing users to call `colnames()`.

Nevertheless, I think it's still helpful to nudge users in the right direction if they're using `names()` on a `tbl_lazy` object, since the value will be confusing and very few users will actually want the results.

The same can be said for `$` -- even if it doesn't return a column, I would guess that's what the vast majority of people who wind up using `$` on a `tbl_lazy` object will have intended.

We can improve the UX here by turning these methods into simple wrappers that nudge the user the first time they use `names()`, but then fall silent to get out of their hair, e.g. (rough sketch)

```r
names.tbl_lazy = function(x) {
  if (!cache$warned) { cache$warned <- TRUE; message("Did you mean colnames()?") }
  NextMethod("names")
}
```

mgirlich:
Closes #1056.

For the `names()` method this uses the same check on the enviroment as `lifecycle::deprecate_soft()`.
For `$` this throws an error if anything other than `lazy_query` or `src` is accessed.


R/tbl-lazy.R
```diff
@@ -105,7 +105,60 @@ the$warned_on_tbl_lazy_names <- FALSE
...
+    ))
+  }
+  NextMethod("$")
+}
+
+# COPIED FROM lifecycle
+# https://github.com/r-lib/lifecycle/blob/9417eca8f5091f95b4569fb6c388e4394e2b2157/R/deprecate.R#L358
+is_direct <- function(env) {
+  env_inherits_global(env) || from_testthat(env)
+}
```
hadley, line 134: You can use `rlang::env_is_user_facing()`: https://rlang.r-lib.org/reference/env_is_user_facing.html 😄

R/tbl-lazy.R
```diff
@@ -105,7 +105,60 @@ the$warned_on_tbl_lazy_names <- FALSE
 
 #' @export
 names.tbl_lazy <- function(x) {
-  # FIXME find a way to nudge users away from using `names()` without RStudio
-  # getting on your nerves
+  if (is_direct(rlang::caller_env())) {
+    cli::cli_inform(c(
+      `!` = "The {.fn names} method of {.cls tbl_lazy} is not intended to be called.",
```
hadley, line 110:
```suggestion
      `!` = "The {.fn names} method of {.cls tbl_lazy} for internal use only.",
```

Maybe?

R/tbl-lazy.R
```diff
@@ -105,7 +105,60 @@ the$warned_on_tbl_lazy_names <- FALSE
...
+      i = "Did you mean {.fn colnames}?"
+    ))
+  }
   NextMethod("names")
 }
+
+#' @export
+`$.tbl_lazy` <- function(x, name) {
+  unexpected_name <- !name %in% c("lazy_query", "src")
```
hadley, line 119: Probably worth a snapshot test here (and above), just in case.

R/tbl-lazy.R
mgirlich: That makes is so much easier 😄

PR #1316, opened 2023-06-28
MS SQL Fix Translation of Unordered `row_number()`

ejneer:
Fixes `row_number` issue of #1233

PR #1331, opened 2023-07-08
Support unordered row_number() on Snowflake, MSSQL, and Teradata

fh-mthomson:
* Snowflake: Closes [TBD] 
* MSSQL: Expands / revises logic in #1316 (to closes `row_number()` portion of #1233 for MSSQL)
* Teradata: Removes Teradata-specific handling

Of note, from https://stackoverflow.com/questions/44105691/row-number-without-order-by per second comment, I opted to use `SELECT NULL` over `SELECT 1` (which also aligns with the implementation for Teradata).


tests/testthat/_snaps/backend-teradata.md
```diff
@@ -53,9 +53,21 @@
       mf %>% group_by(y) %>% mutate(rown = row_number())
     Output
       <SQL>
-      SELECT `df`.*, ROW_NUMBER() OVER (PARTITION BY `y` ORDER BY `y`) AS `rown`
+      SELECT
+        `df`.*,
+        ROW_NUMBER() OVER (PARTITION BY `y` ORDER BY (SELECT NULL)) AS `rown`
```
fh-mthomson, line 58: Interestingly, Teradata (via `win_rank_tdata()`) was previously defaulting to `ORDER BY` the result of `win_current_group()` (added in https://github.com/tidyverse/dbplyr/commit/4109051ff697446bc5a280d65aa708e568047f57#diff-c4f086980e5692d16e3337acd7bfefe3a9315135688133ee4f24ea2944ccba3bR174 by @overmar ).

By generalizing the behavior to  fall back to `SELECT NULL` for all back ends (incl. Snowflake / MSSQL) this now no longer defaults to ordering by the group. 

I'm likely not familiar enough with (1) PARTITION vs ORDER  and (2) Teradata to determine whether this would have adverse impacts, so would welcome input from more knowledgeable folks!

fh-mthomson:
@mgirlich would you be open to reviewing? thank you!

fh-mthomson:
@mgirlich @hadley would you be open to including this PR in the 2.4.0 release? 

Apologies for the direct pester, but it's last (known) Snowflake <> `dbplyr` blocker for us; would love to close out if the timing works for y'all!

R/translate-sql-window.R
```diff
@@ -147,6 +146,7 @@ win_rank <- function(f) {
       no_na_expr <- purrr::reduce(not_is_na_exprs, ~ call2("&", .x, .y))
     } else {
       order_over <- win_current_order()
+      if (use_default_order_null & is_empty(order_over)) order_over <- sql("(SELECT NULL)")
```
hadley, line 149:
```suggestion
      if (use_default_order_null & is_empty(order_over)) {
         order_over <- sql("(SELECT NULL)")
      }
```

R/backend-teradata.R
```diff
@@ -217,20 +217,3 @@ sql_table_analyze.Teradata <- function(con, table, ...) {
 }
 
 utils::globalVariables(c("ATAN2", "SUBSTR", "DECIMAL", "WEEKNUMBER_OF_YEAR", "SUM"))
-
-#' @export
-#' @rdname win_over
-win_rank_tdata <- function(f) {
```
hadley, line 223: Was this exported in a released dbplyr?

R/translate-sql-window.R
```diff
@@ -122,10 +122,9 @@ rows <- function(from = -Inf, to = 0) {
   }
 }
 
-
 #' @rdname win_over
 #' @export
-win_rank <- function(f) {
+win_rank <- function(f, use_default_order_null = FALSE) {
```
hadley, line 127: This argument name feels a bit long to me, and needs to be documented. Maybe `empty_order = TRUE` or  `order_by = c("required", "optional")`, or something along those lines?

hadley (commented):
Thanks for working on this!

@mgirlich I'm ok with allowing an argument-less `row_number()`; it doesn't really make a lot of sense for databases, but it's still sometimes useful.

R/backend-teradata.R
fh-mthomson: It was, as of 2.3.0 (added in https://github.com/tidyverse/dbplyr/commit/4109051ff697446bc5a280d65aa708e568047f57#diff-c4f086980e5692d16e3337acd7bfefe3a9315135688133ee4f24ea2944ccba3bR167). I'd guess it was primarily to enable `row_number()` for Teradata backend.

Open to any recommended patterns for deprecation!

R/translate-sql-window.R
fh-mthomson: Changed to `empty_order` (defaulting to `FALSE` to not apply to all backends) and added some docs!

mgirlich:
Thanks for the PR! 😄

PR #1350, opened 2023-08-10
add escaping for Snowflake interval functions

### LINKED ISSUE #1351: Regression: migrating from build_sql() to glue_sql2() lost escaping
# Problem summary
In https://github.com/tidyverse/dbplyr/pull/1228, most references to `build_sql()` where replaced with `glue_sql2()`.

This subtly broke some fundamental SQL translations, since `escape()` methods were no longer being applied.

# Reprex (using Snowflake as my motivating example)

## Previously:
```r
local_con(dbplyr::simulate_snowflake())
months = function(x) {
  build_sql("INTERVAL '", x, " month'")
}
months(1)
  <SQL> INTERVAL '1.0 month'
```

## current translation, which fails:
```r
months = function(x) {
  glue_sql2(sql_current_con(), "INTERVAL '{x} month'")
}
months(1)

Error in `.transformer()`:
! `value` must be a string or scalar SQL, not the number 1.
Run `rlang::last_trace()` to see where the error occurred.
```

# Solutions considered
1. [hacky] Manually add `escape(x)`  in applicable calls to `glue_sql2()` (e.g., #1350 to mitigate breaking Snowflake translations)
2. [robust] Migrate relevant `escape()` functionality from `build_sql()` to `gluesql2()` (https://github.com/tidyverse/dbplyr/blob/main/R/build-sql.R#L32-L44)

fh-mthomson:
Closes #1351 (Solution 1, just for Snowflake)


mgirlich:
Closed in favour of #1352.
I tried to find and fix all the new issues after switching to `glue_sql2()`. This also has the easier way of simply using `{.val x}` instead of manually escaping 😄

Issue #1080, opened 2022-12-15
function translations for DATEDIFF and DATEADD 


ablack3:
**Feature Request**

DATEDIFF and DATEADD are commonly used SQL functions with slightly different syntax on different database platforms. I would like to use dbplyr's translation approach to generate SQL statements DATEDIFF and DATEADD that are correct for various database platforms.

I can use sql passthrough but then the dplyr code won't be translated and work on multiple database platforms. I would like to write one dplyr expression to calculate the difference between dates (or addition of an interval to a date) that will be translated correctly on both sql server and postgres. Below is a reprex that generates correct SQL on postgres but not sql server.

``` r
library(tidyverse)
df <- tibble(date1 = as.Date("2020-02-02"), date2 = as.Date("2020-02-03"))
db <- dbplyr::lazy_frame(df, con = dbplyr::simulate_postgres())

# Date add
db %>% 
  mutate(date3 = date1 + lubridate::years(1)) %>% 
  show_query()
#> <SQL>
#> SELECT *, `date1` + CAST('1 years' AS INTERVAL) AS `date3`
#> FROM `df`

# Date diff
db %>% 
  mutate(days = date2 - date1) %>% 
  show_query()
#> <SQL>
#> SELECT *, `date2` - `date1` AS `days`
#> FROM `df`

db <- dbplyr::lazy_frame(df, con = dbplyr::simulate_mssql())

# Date add
db %>% 
  mutate(date3 = date1 + lubridate::years(1)) %>% 
  show_query()
#> <SQL>
#> SELECT *, `date1` + years(1.0) AS `date3`
#> FROM `df`

# Date diff
db %>% 
  mutate(days = date2 - date1) %>% 
  show_query()
#> <SQL>
#> SELECT *, `date2` - `date1` AS `days`
#> FROM `df`
```

<sup>Created on 2022-12-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

mgirlich:
dbplyr doesn't know the column types, so it is not possible to simply translate `+` and `-` correctly for dates. Therefore, a custom function like `date_add()` and `date_diff()` would be necessary. Unfortunately, this would probably feel a bit artificial as this isn't required in R. Also, this would be quite hard to find. So, I'm not sure it is worthing adding such a function.

@hadley What's your opinion on this?

ablack3:
What about translating the R function `difftime` to `DATEDIFF`?

hadley:
[...]

PR #1357, opened 2023-08-27
Add mssql translations for two clock functions

ablack3:
@hadley, @mgirlich. Here is what I'm thinking for two of the clock functions. Once I learn how to make well formed PRs I can add more but I wanted to see what you think about this one first. Can you review and discuss or make suggestions? Thanks!

### DIFF
--- a/R/backend-mssql.R
+++ b/R/backend-mssql.R
@@ -350,6 +350,16 @@ simulate_mssql <- function(version = "15.0") {
           sql_expr(DATEPART(QUARTER, !!x))
         }
       },
+
+      # clock ---------------------------------------------------------------
+      add_days = function(x, n, ...) {
+        rlang::check_dots_empty(...)
+        sql_expr(DATEADD(DAY, !!n, !!x))
+      },
+      add_years = function(x, n, ...) {
+        rlang::check_dots_empty(...)
+        sql_expr(DATEADD(YEAR, !!n, !!x))
+      }
     )
 
   if (mssql_version(con) >= "11.0") { # MSSQL 2012
--- a/tests/testthat/test-backend-mssql.R
+++ b/tests/testthat/test-backend-mssql.R
@@ -124,6 +124,19 @@ test_that("custom lubridate functions translated correctly", {
   expect_error(test_translate_sql(quarter(x, fiscal_start = 5)))
 })
 
+test_that("custom clock functions translated correctly", {
+  local_con(simulate_mssql())
+  expect_equal(test_translate_sql(add_years(x, 1L)), sql("DATEADD(YEAR, 1, `x`)"))
+  expect_equal(test_translate_sql(add_years(x, 1)), sql("DATEADD(YEAR, 1.0, `x`)"))
+  expect_equal(test_translate_sql(add_days(x, 1L)), sql("DATEADD(DAY, 1, `x`)"))
+  expect_equal(test_translate_sql(add_days(x, 1)), sql("DATEADD(DAY, 1.0, `x`)"))
+
+  expect_equal(test_translate_sql(clock::add_years(x, 1)), sql("DATEADD(YEAR, 1.0, `x`)"))
+  expect_equal(test_translate_sql(clock::add_days(x, 1)), sql("DATEADD(DAY, 1.0, `x`)"))
+
+  expect_error(test_translate_sql(add_days(x, 1, "dots must be empty")))
+})
+
 test_that("last_value_sql() translated correctly", {
   con <- simulate_mssql()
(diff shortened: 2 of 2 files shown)


hadley:
Yes, this what I was thinking 😄 

Tagging @DavisVaughan in case he has thoughts too,

DavisVaughan:
Seems useful to me!

R/backend-mssql.R
```diff
@@ -350,6 +350,16 @@ simulate_mssql <- function(version = "15.0") {
           sql_expr(DATEPART(QUARTER, !!x))
         }
       },
+
+      # clock ---------------------------------------------------------------
+      add_days = function(x, n, ...) {
+        rlang::check_dots_empty(...)
```
mgirlich, line 356: You can omit the rlang namespace. The rlang package is so commonly used that it is usually imported.
And you either do `check_dots_empty()` or `check_dots_empty0(...)` but you must not pass `...` to `check_dots_empty()`.

```suggestion
        check_dots_empty()
```

mgirlich:
@ablack3 Thanks for you PR. Please add a NEWS entry mentioning the issue #1080 and yourself. And please modify your first message and mention that this closes #1080. This way the issue will be automatically closed when this PR is merged.

mgirlich:
@ablack3 Do you want to finish the PR? 😄 I'm happy to merge afterwards.

ablack3:
> @ablack3 Do you want to finish the PR? 😄 I'm happy to merge afterwards.

Yes, I'll finish it this week.

hadley:
Yeah, I think `difftime()` is a reasonable translation.

hadley:
@ablack3 as you're working on this, it would be super useful if you'd take some notes so we could document the process of adding date-time wrappers for other backends (and then we could provide more expectations about how to express date time manipulations in R code for the user)

ablack3:
> @ablack3 as you're working on this, it would be super useful if you'd take some notes so we could document the process of adding date-time wrappers for other backends (and then we could provide more expectations about how to express date time manipulations in R code for the user)

Yes of course. I'll write some documentation and share it for feedback. It would be nice to have one set of date manipulation functions that translate to many backends.

ablack3:
Ok I think this PR is ready. I would expect that you'll have some recommendations so please let me know if you suggested edits to the PR. I will also add similar translations in the bigrquery package and duckdb package in separate PRs. Just wondering why the backends for bigquery and duckdb are not in the dbplyr package.

R/backend-postgres.R
```diff
@@ -235,6 +235,29 @@ sql_translation.PqConnection <- function(con) {
...
       },
+
+      # clock ---------------------------------------------------------------
+      add_days = function(x, n, ...) {
+        check_dots_empty()
+        sql_expr((!!x + !!n%*INTERVAL%'1 day'))
+      },
+      add_years = function(x, n, ...) {
+        check_dots_empty()
+        sql_expr((!!x + !!n%*INTERVAL%'1 year'))
```
ablack3, line 246: @mgirlich this works but is very awkward. Is there a better way to get the correct SQL here?

This is what I want to write but it is not a valid R expression
`sql_expr((!!x + !!n * INTERVAL'1 year'))`

The SQL should look like `\`column_name\` + 2 * INTERVAL'1 year'`

R/backend-postgres.R
```diff
@@ -235,6 +235,29 @@ sql_translation.PqConnection <- function(con) {
...
+      },
+
+      difftime = function(time1, time2, tz, units = "days") {
+
+        if (!missing(tz)) {
+          cli::cli_abort("The {.arg tz} argument is not supported for SQL backends.")
+        }
+
+        if (units[1] != "days") {
+          cli::cli_abort('The only supported value for {.arg units} on SQL backends is "days"')
```
ablack3, line 256: I thought we could just support the 'days' unit to start. Keep it simple.

R/backend-postgres.R
ablack3: @hadley is there a better way to write this expression? `!!x + !!n%*INTERVAL%'1 year'` seems strange but it's the only way I could come up with that worked.

hadley (approved):
This seems like a good place to start. Do you want to add a news bullet and we can merge?

ablack3:
> This seems like a good place to start. Do you want to add a news bullet and we can merge?

Ok I added a few more translations and added a news bullet. Thank you!

R/backend-postgres.R
ablack3: Ok I tried glue_sql2 instead. Hopefully I used it correctly.

PR #1374, opened 2023-10-06
Check duplicate names in sql_translator()

krlmlr:
Tests will fail because `round()` is defined twice for the base translation.


mgirlich:
Good catch. Thanks for the PR!

PR #1375, opened 2023-10-06
Work around missing CAST(x AS INTEGER) in MySQL

krlmlr:
Also introduces `simulate_mariadb()` to account for the differences between MySQL and MariaDB.


mgirlich (approved):

mgirlich:
Thanks for the PR!

PR #1402, opened 2023-11-09
add str_replace and str_replace_all for Oracle backend

### LINKED ISSUE #1393: backend-oracle.R implement str_replace functionality similar to MySQL.
Feature update for Oracle SQL. I noticed that str_replace_* does not have variant support for Oracle SQL.  I was wondering if we could update the backend for Oracle SQL to include this functionality. 

See the MSSQL issue here: <https://community.rstudio.com/t/str-replace-all-with-mssql/54700>. This is similar to what I would like implemented in backend-oracle.R as it does have the REGEXP_REPLACE function <https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/REGEXP_REPLACE.html#GUID-EA80A33C-441A-4692-A959-273B5A224490>

The connection is here:
image

The error the code produces is this:
image

This is similar to the original issue from the MSSQL, in this case though the Oracle SQL can accept the REGEXP_REPLACE functionality. 
<https://github.com/tidyverse/dbplyr/blob/785d106560cd5e1498f2c9309de71d6e8c6dd189/R/backend-mysql.R#L41>

Is this a feature that can be added to the backend-oracle.R?

thomashulst:
Fixes #1393


hadley:
Would you mind adding a couple of snapshot tests and a news bullet?

hadley:
Thanks again @thomashulst!

Issue #1405, opened 2023-11-10
Translate `str_starts()` to `REGEXP_INSTR` instead of `INSTR` for Snowflake


nathanhaigh:
Snowflake doesn't have a `INSTR` function. As such, I think the following `stringr` functions are affected:

- str_detect() when `fixed()` is used
- str_ends()
- str_starts()

```r
x <- c("XXX", "XXY", "YXXX")

#####
# str_detect()
#####
lazy_frame(TOKEN=x,con = simulate_snowflake()) %>%
  filter(
    stringr::str_detect(TOKEN, fixed("XXX"))
  )
<SQL>
SELECT `df`.*
FROM `df`
WHERE (INSTR(`TOKEN`, 'XXX') > 0)

#####
# str_starts()
#####
lazy_frame(TOKEN=x,con = simulate_snowflake()) %>%
  filter(
    stringr::str_starts(TOKEN, fixed("XXX"))
  )
<SQL>
SELECT `df`.*
FROM `df`
WHERE (INSTR(`TOKEN`, 'XXX') = 1)

#####
# str_ends()
#####
lazy_frame(TOKEN=x,con = simulate_snowflake()) %>%
  filter(
    stringr::str_ends(TOKEN, fixed("XXX"))
  )
<SQL>
SELECT `df`.*
FROM `df`
WHERE (INSTR(`TOKEN`, 'XXX') = ((LENGTH(`TOKEN`) - LENGTH('XXX')) + 1))
```

nathanhaigh:
I see from `R/backend-snowflake.R` the `str_detect()` function is hacking around the fact that `REGEXP` anchors the start and end by default. A simpler solution is to use `CONTAINS()` instead.

hadley:
Fixed in #1406

PR #1406, opened 2023-11-10
Better support for string matching in Snowflake

nathanhaigh:
* Added support for str_starts() and str_ends() by using Snowflake's REGEXP_INSTR() function
 * Refactored str_detect() to use Snowflake's CONTAINS() function instead of hacking with REGEXP() which anchors the start and end by default

Satisfies #1405

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # dbplyr (development version)
 
+* Snowflake (@nathanhaigh, #1406)
+  * Added support for `str_starts()` and `str_ends()` via `REGEXP_INSTR()`
+  * Refactored `str_detect()` to use `REGEXP_INSTR()` so now supports
+    regular expressions.
+  * Refactored `grepl()` to use `REGEXP_INSTR()` so now supports
+    case-insensitive matching through `grepl(..., ignore.case = TRUE)`
+
 * Functions qualified with the base namespace are now also translated, e.g.
   `base::paste0(x, "_1")` is now translated (@mgirlich, #1022).
 
--- a/R/backend-snowflake.R
+++ b/R/backend-snowflake.R
@@ -38,23 +38,32 @@ sql_translation.Snowflake <- function(con) {
       str_locate = function(string, pattern) {
         sql_expr(POSITION(!!pattern, !!string))
       },
-      # REGEXP on Snowflaake "implicitly anchors a pattern at both ends", which
-      # str_detect does not.  Left- and right-pad `pattern` with .* to get
-      # str_detect-like behavior
       str_detect = function(string, pattern, negate = FALSE) {
-        sql_str_pattern_switch(
-          string = string,
-          pattern = {{ pattern }},
-          negate = negate,
-          f_fixed = sql_str_detect_fixed_instr("detect"),
-          f_regex = function(string, pattern, negate = FALSE) {
-            if (isTRUE(negate)) {
-              sql_expr(!(((!!string)) %REGEXP% (".*" || (!!pattern) || ".*")))
-            } else {
-              sql_expr(((!!string)) %REGEXP% (".*" || (!!pattern) || ".*"))
-            }
-          }
-        )
+        con <- sql_current_con()
+
+        if (negate) {
+          translate_sql(!CONTAINS(!!string, !!pattern), con = con)
+        } else {
+          translate_sql(CONTAINS(!!string, !!pattern), con = con)
+        }
+      },
+      str_starts = function(string, pattern, negate = FALSE) {
+        con <- sql_current_con()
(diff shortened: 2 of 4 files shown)


fh-afrachioni:
Thank you for working on Snowflake support!

`CONTAINS` does not support regular expressions.  I would argue pattern matching is an expected and important behavior of `str_detect()`, worth the minor hack to get us there from `REGEXP`.

nathanhaigh:
> Thank you for working on Snowflake support!
> 
> `CONTAINS` does not support regular expressions. I would argue pattern matching is an expected and important behavior of `str_detect()`, worth the minor hack to get us there from `REGEXP`.

Agreed about pattern matching! Serves me right for working on this late a night and not reading the Snowflake document page for `CONTAINS` properly!

I wonder if swapping over to `REGEXP_INSTR` might be a less hacky alternative to `REGEXP`?

fh-afrachioni:
> I wonder if swapping over to [REGEXP_INSTR](https://docs.snowflake.com/en/sql-reference/functions/regexp_instr) might be a less hacky alternative to [REGEXP](https://docs.snowflake.com/en/sql-reference/functions/regexp)?

I like it!  I think the need to compare the result of `REGEXP_INSTR` with `0` is cleaner than the existing pattern-padding.

I _especially_ like it because I just realized `REGEXP_INSTR` takes an optional `<regexp_parameters>` argument, for which one option is
> `i`: case-insensitive.

This isn't important for `str_detect()`, but it does open the door to `grepl(..., ignore.case = TRUE)`, which has been an annoyance here for awhile.  How would you feel about updating `grepl()` as well?

nathanhaigh:
> This isn't important for `str_detect()`, but it does open the door to `grepl(..., ignore.case = TRUE)`, which has been an annoyance here for awhile. How would you feel about updating `grepl()` as well?

Sure - I can take a look at that.

nathanhaigh:
@fh-afrachioni Done - please take a look and let me know if there is anything further that might need adding.

fh-afrachioni:
Hi @nathanhaigh , this looks great!  I'm excited to finally have `ignore.case` working.

One observation: I think we might need another level of backslash escaping.  In order to use [`\D`](https://docs.snowflake.com/en/sql-reference/functions-regexp#label-regexp-general-usage-notes), for example, I apparently need to write ` grepl('\\\\D', col_name)` in my R IDE (using `\` as the R escape character, to encode two literal `\` in R code, which are sent to Snowflake without being escaped, and interpreted as a single, backslash-escaped, `\`).  [Snowflake needs the backslash character to be escaped in string literals](https://docs.snowflake.com/en/sql-reference/functions-regexp#label-regexp-escape-character-caveats).

nathanhaigh:
@fh-afrachioni Are you proposing that the user shouldn't need to know that the extra level of backslash escaping is required? If so, do you have any thoughts on how to achieve this? I'm happy to do some reading but would need a couple of pointers in the right direction as I'm not familiar with the dbplyr codebase.

nathanhaigh:
> Hi @nathanhaigh , this looks great! I'm excited to finally have `ignore.case` working.
> 
> One observation: I think we might need another level of backslash escaping. In order to use [`\D`](https://docs.snowflake.com/en/sql-reference/functions-regexp#label-regexp-general-usage-notes), for example, I apparently need to write ` grepl('\\\\D', col_name)` in my R IDE (using `\` as the R escape character, to encode two literal `\` in R code, which are sent to Snowflake without being escaped, and interpreted as a single, backslash-escaped, `\`). [Snowflake needs the backslash character to be escaped in string literals](https://docs.snowflake.com/en/sql-reference/functions-regexp#label-regexp-escape-character-caveats).

Good catch @fh-afrachioni - fixed in the latest commit

nathanhaigh:
Is there anything further I need to do to get this PR merged?

fh-afrachioni:
Thanks for making the update!  I'm not a maintainer, but maybe @mgirlich would be willing to consider this lightweight change.

hadley (approved):
LGTM.  Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

nathanhaigh:
@hadley done :)

hadley:
Thanks!

yuhenghuang:
Came all the way here for the same issue in #1405 .

I got confused that `main` branch (the source code I was checking) == `2.4.0` branch (the dbplyr I was using) and couldn't find why the code seems to work whereas the error's still there.

Now that I am relieved this will be fixed in the next stable release and all I should do now is wait.

Great thanks!

PR #1432, opened 2023-12-22
Translations for namespaced functions must exist

### LINKED ISSUE #1426: Make namespaced functions error if no translation?
i.e. shouldn't this be an error?

``` r
library(dbplyr)

lazy_frame(x = 1) |> mutate(x = dplyr::foo(x))
#> <SQL>
#> SELECT foo(`x`) AS `x`
#> FROM `df`
```

<sup>Created on 2023-12-21 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

hadley:
I'm not sure how this broke, but the code for handling this already exists in `translate_sql()`, so I'm not sure why we started doing some `::` handling in `partial_eval()`.

Fixes #1426

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dbplyr (development version)
 
+* Namespaced dplyr calls now error if the function doesn't exist, or 
+  a translation is not available (#1426).
+
 * SQL server: `filter()` does a better job of converting logical vectors 
   from bit to boolean (@ejneer, #1288). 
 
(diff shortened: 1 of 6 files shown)


mgirlich (approved):

mgirlich:
We might also simply remove the compound calls logic in `partial_eval_call()`. At least no test failed when I removed it.

hadley:
Let's tackle that in a separate PR. For example, `.env$bar()` currently errors, but maybe it's more consistent for it to be equivalent to `local(bar())`?

PR #1442, opened 2024-01-19
Support unnamed Id() for DBI >= 1.2.0

krlmlr:
There are more tests that need to be adapted, but this is necessary for dm.


hadley:
Closing since we're doing a bigger rewrite.

PR #1444, opened 2024-01-31
Fix bug in `rows_patch.tbl_lazy()` throwing an error when multiple columns need to be patched

### LINKED ISSUE #1443: `rows_patch.tbl_lazy` fails when patching more than one column
The `tbl_lazy` version of `rows_patch()` throws an error if more than one column is to be updated.
There's a simple bug in the code that builds the `coalesce` calls, PR is incoming!

``` r
library(dplyr)
library(dbplyr)

rows_patch(
  memdb_frame(x = 1:3, y = c(11, 12, NA), z = c(31, NA, 33)),
  memdb_frame(x = 2:3, y = 22:23, z = 42:43),
  by = "x",
  unmatched = "ignore",
  in_place = FALSE
)
#> Error in `sym()`:
#> ! Can't convert a character vector to a symbol.
```

<sup>Created on 2024-01-31 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

gorcha:
Close #1443

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # dbplyr (development version)
 
-* `rows_patch()` no longer throws an error when more than one column needs to be
+* `rows_patch(in_place = FALSE)` now works when more than one column should be
   patched (@gorcha, #1443).
 
 * Namespaced dplyr calls now error if the function doesn't exist, or 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dbplyr (development version)
 
+* `rows_patch()` no longer throws an error when more than one column needs to be
+  patched (@gorcha, #1443).
+
 * Namespaced dplyr calls now error if the function doesn't exist, or 
   a translation is not available (#1426).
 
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # dbplyr (development version)
 
+* `rows_patch()` no longer throws an error when more than one column needs to be
+  patched (@gorcha, #1443).
```
mgirlich, line 4: This should mention that this fix is for `in_place = FALSE` and I think it is a bit nicer to state it positively ("now works" instead of "no longer throws an error")

```suggestion
* `rows_patch(in_place = FALSE)` now works when more than one column should be
  patched (@gorcha, #1443).
```

mgirlich (approved):

mgirlich:
Thanks! For filing the issue and creating the PR to fix it. Just a minor change in the NEWS and then I'm happy to merge.

gorcha:
Thanks @mgirlich!

PR #1446, opened 2024-02-07
Refine `select` inline criteria to keep `arrange`d computed columns

### LINKED ISSUE #1437: `select()` fails after specific sequence of `dplyr` commands
It seems that when you sequence a `group_by()`, then a `mutate()`, and lastly an `arrange()`, `dbplyr` is not able to find the variable created by the `mutate()` when using `select()`. This works against a regular data frame. Reprex here:

``` r
suppressPackageStartupMessages(library(dplyr)) 
suppressPackageStartupMessages(library(dbplyr))
memdb_frame(
  Name = c("Alice", "Bob", "Alice", "Bob", "Charlie"),
  Subject = c("Math", "Math", "English", "English", "Math"),
  Score = c(90, 85, 88, 92, 78)
) %>% 
  group_by(Name) %>% 
  mutate(AvgScore = 1) %>% 
  ungroup() %>% 
  arrange(AvgScore, Name) %>% 
  select(Name)
#> Error in `collect()`:
#> Caused by error:
#> ! no such column: AvgScore
```

<sup>Created on 2024-01-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Initially reported as a possible bug in `sparklyr` here: https://github.com/sparklyr/sparklyr/issues/3413

ejneer:
Fixes #1437

This will stop `arrange`d computed columns from being inlined away from subqueries. Of course the subquery `ORDER BY` warning will be thrown, but I think that is expected behavior.

Before:

``` r
library(dbplyr)

db <- lazy_frame(x = 1, y = 2) 

db %>% 
  mutate(z = 1) %>% 
  arrange(x, z) %>%
  select(x)
#> <SQL>
#> SELECT `x`
#> FROM `df`
#> ORDER BY `x`, `z`
```

<sup>Created on 2024-02-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

After:

``` r
library(dbplyr)

db <- lazy_frame(x = 1, y = 2) 

db %>% 
  mutate(z = 1) %>% 
  arrange(x, z) %>%
  select(x)
#> <SQL>
#> SELECT `x`
#> FROM (
#>   SELECT `df`.*, 1.0 AS `z`
#>   FROM `df`
#> ) AS `q01`
```

<sup>Created on 2024-02-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/verb-select.R
+++ b/R/verb-select.R
@@ -134,11 +134,8 @@ add_select <- function(lazy_query, vars) {
 select_can_be_inlined <- function(lazy_query, vars) {
   vars_data <- op_vars(lazy_query)
 
-  computed_columns <- purrr::pmap(
-    lazy_query$select,
-    function(name, expr, ...) if (is_quosure(expr)) name
-  ) %>%
-    purrr::compact()
+  computed_flag <- purrr::map_lgl(lazy_query$select$expr, is_quosure)
+  computed_columns <- lazy_query$select$name[computed_flag]
 
   order_vars <- purrr::map_chr(lazy_query$order_by, as_label)
 
--- a/R/verb-select.R
+++ b/R/verb-select.R
@@ -134,9 +134,6 @@ add_select <- function(lazy_query, vars) {
 select_can_be_inlined <- function(lazy_query, vars) {
   vars_data <- op_vars(lazy_query)
 
-  bijective_or_not_distinct <- is_bijective_projection(vars, vars_data) ||
-      !is_true(lazy_query$distinct)
-
   computed_columns <- purrr::pmap(
     lazy_query$select,
     function(name, expr, ...) if (is_quosure(expr)) name
@@ -145,12 +142,10 @@ select_can_be_inlined <- function(lazy_query, vars) {
 
   order_vars <- purrr::map_chr(lazy_query$order_by, as_label)
 
-  all(
-    # computed columns used for ordering (if any) must also be
-    # selected to inline the query
-    intersect(computed_columns, order_vars) %in% vars,
-    bijective_or_not_distinct
-  )
+  # computed columns used for ordering (if any) must also be
+  # selected to inline the query
+  all(intersect(computed_columns, order_vars) %in% vars) &&
+      (is_bijective_projection(vars, vars_data) || !is_true(lazy_query$distinct))
 }
 
 is_bijective_projection <- function(vars, names_prev) {
--- a/R/verb-select.R
+++ b/R/verb-select.R
@@ -132,22 +132,19 @@ add_select <- function(lazy_query, vars) {
 }
 
 select_can_be_inlined <- function(lazy_query, vars) {
-  vars_data <- op_vars(lazy_query)
-
+  # all computed columns used for ordering (if any) must be present
   computed_flag <- purrr::map_lgl(lazy_query$select$expr, is_quosure)
   computed_columns <- lazy_query$select$name[computed_flag]
 
   order_vars <- purrr::map_chr(lazy_query$order_by, as_label)
+  ordered_present <- all(intersect(computed_columns, order_vars) %in% vars)
 
-  # computed columns used for ordering (if any) must also be
-  # selected to inline the query
-  all(intersect(computed_columns, order_vars) %in% vars) &&
-      (is_bijective_projection(vars, vars_data) || !is_true(lazy_query$distinct))
-}
+  # if the projection is distinct, it must be bijective
+  is_distinct <- is_true(lazy_query$distinct)
+  is_bijective_projection <- identical(sort(unname(vars)), op_vars(lazy_query))
+  distinct_is_bijective <- !is_distinct || is_bijective_projection
 
-is_bijective_projection <- function(vars, names_prev) {
-  vars <- unname(vars)
-  identical(sort(vars), names_prev)
+  ordered_present && distinct_is_bijective
 }
 
 rename_groups <- function(lazy_query, vars) {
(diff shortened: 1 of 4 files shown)


R/verb-select.R
```diff
@@ -131,6 +131,28 @@ add_select <- function(lazy_query, vars) {
...
+
+  computed_columns <- purrr::pmap(
+    lazy_query$select,
+    function(name, expr, ...) if (is_quosure(expr)) name
+  ) %>%
+    purrr::compact()
+
+  order_vars <- purrr::map_chr(lazy_query$order_by, as_label)
+
+  all(
```
hadley, line 148: Given there's only two clauses here, I think `&&` would be easier to understand.

R/verb-select.R
```diff
@@ -131,6 +131,28 @@ add_select <- function(lazy_query, vars) {
   )
 }
 
+select_can_be_inlined <- function(lazy_query, vars) {
+  vars_data <- op_vars(lazy_query)
+
+  bijective_or_not_distinct <- is_bijective_projection(vars, vars_data) ||
+      !is_true(lazy_query$distinct)
+
```
hadley, line 139: I'd probably not name this (since the variable name is rather long and only used in one place) and instead inline in the logical expression below.

hadley:
Thanks for taking a stab at this!

hadley:
@mgirlich could you please take a look at this PR too

R/verb-select.R
```diff
@@ -131,6 +131,23 @@ add_select <- function(lazy_query, vars) {
...
 }
 
+select_can_be_inlined <- function(lazy_query, vars) {
+  vars_data <- op_vars(lazy_query)
+
+  computed_columns <- purrr::pmap(
+    lazy_query$select,
+    function(name, expr, ...) if (is_quosure(expr)) name
+  ) %>%
+    purrr::compact()
```
mgirlich, line 141:
```suggestion
  computed_flag <- purrr::map_lgl(lazy_query$select$expr, is_quosure)
  computed_columns <- lazy_query$select$name[computed_flag]
```

This is short and a bit more clear. Also this makes sure that `computed_columns` is actually a character vector, not a list.

tests/testthat/_snaps/verb-select.md
```diff
@@ -181,6 +181,22 @@
...
+      Warning:
+      ORDER BY is ignored in subqueries without LIMIT
+      i Do you need to move arrange() later in the pipeline or use window_order() instead?
+    Output
+      <SQL>
+      SELECT `x`
+      FROM (
+        SELECT `df`.*, 1.0 AS `z`
+        FROM `df`
+      ) AS `q01`
```
mgirlich, line 198: To me the snapshot doesn't look like it fits to the title of the test. Can you come up with a test that tests the intended behaviour more directly?

mgirlich (approved):

tests/testthat/_snaps/verb-select.md
ejneer: This is just supposed to show that the subquery hasn't been removed. Do you think it should not be a snapshot test, or that should have a more descriptive title. If the latter, how about "Computed columns used to arrange are not removed by subsequent selects"?

mgirlich: I think if I would come back to this test in a year and see the snapshot I wouldn't understand if the snapshot is correct or not. So, probably it shouldn't be a snapshot.

You could test the optimisation more directly, e.g.

```r
  out <- lf %>% mutate(z = 1) %>% arrange(x, z) %>% select(x)
  
  # 
  inner_query <- out$lazy_query$x
  expect_s3_class(inner_query, "lazy_select_query")
  expect_equal(
    inner_query$order_by,
    list(quo(x), quo(z)),
    ignore_formula_env = TRUE
  )

  expect_equal(op_vars(inner_query), c("x", "y", "z"))
  expect_equal(op_vars(out$lazy_query), "x")
```

and maybe also compare this to `lf %>% mutate(z = 1) %>% arrange(x, z) %>% select(x, z)`, i.e. a case where the inline optimisation was applied.

ejneer: appreciate the pointers!

R/verb-select.R
```diff
@@ -131,9 +131,20 @@ add_select <- function(lazy_query, vars) {
   )
 }
 
-is_bijective_projection <- function(vars, names_prev) {
-  vars <- unname(vars)
-  identical(sort(vars), names_prev)
+select_can_be_inlined <- function(lazy_query, vars) {
```
hadley, line 134: I made a few tweaks here to hopefully make the logic a bit clearer.

tests/testthat/test-verb-select.R
```diff
@@ -35,6 +35,31 @@ test_that("select after distinct produces subquery", {
   expect_true(lq$x$distinct)
 })
 
+test_that("select after arrange produces subquery, if needed", {
+  lf <- lazy_frame(x = 1)
+
+  # shouldn't inline
+  out <- lf %>% mutate(z = 2) %>% arrange(x, z) %>% select(x)
```
hadley, line 42: I'm not sure this is testing the right thing because when I run it, I see:

```
SELECT `x`
FROM (
  SELECT `df`.*, 2.0 AS `z`
  FROM `df`
) AS `q01`
Warning message:
ℹ Do you need to move arrange() later in the pipeline or use window_order() instead? 
```

hadley: Or maybe that is the expected behaviour? But definitely worth a snapshot test, I think.

ejneer: Yeah that is expected. I've re-added the snapshot I removed in my prior commit.

hadley:
Thanks for your work on this @ejneer — it's much appreciated!

PR #1458, opened 2024-02-20
Export table_path infrastructure

### LINKED ISSUE #1300: Export <dbplyr_table_ident> tools
At least for the dm package this would be nice.
After we did this we might be able to get rid of the classes <dbplyr_catalog>, <dbplyr_schema()> and <ident_q> in the future. But its not easy to get rid of a class. Maybe we could do that by deprecating escape.dbplyr_schema/dbplyr_catalog/ident_q().

hadley:
Fixes #1300

Includes some changes to make the interface a little nicer and to fix some issues I noticed while I was looking at this code with fresh eyes:

* `table_name()` -> `table_path_name()`
* `is_table_id()` requires characters and table_paths to be length 1
* Vectorise `table_path_components()` instead of `table_path_name()`
* Make `table_path_components()` an S3 generic (as an escape hatch if needed)

---

@mgirlich this is a first pass based on what I need for bigrquery. Is there anything obviously missing?


tests/testthat/test-table-name.R
```diff
@@ -88,3 +92,31 @@ test_that("as_table_path warns when using sql", {
...
+  )
+  expect_equal(
+    table_path_components(table_path("`x.y`.z"), con),
+    list(c("x.y", "z"))
+  )
+  expect_equal(
+    table_path_components(table_path("`x.y`.`y.z`"), con),
+    list(c("x.y", "y.z"))
+  )
+})
```
mgirlich, line 113: maybe add an expliciti test that `table_path_components()` is vectorised.

mgirlich (approved):

mgirlich:
I don't see anything obvious missing.

PR #1430, opened 2023-12-22
Improve $ handling

hadley:
Fixes #1368

PR #1466, opened 2024-02-22
Soften rules for interpolating values

### LINKED ISSUE #1465: Research stricter interpolation revdep failures
* arrow — https://github.com/apache/arrow/pull/40197
* bcdata — https://github.com/bcgov/bcdata/pull/338
* CDMConnector — https://github.com/darwin-eu/CDMConnector/pull/16 (also affects CohortSurvival, DrugUtilisation, IncidencePrevalence, PatientProfiles)

hadley:
Partially reverts #1430. See https://github.com/bcgov/bcdata/pull/338 for details. If we do this again in the future, I think we have to do argument checking in individual translations. That's a lot of work but allows a whole suite of potentially useful translations for more complex R objects.

Fixes #1465


hadley:
Doesn't help enough with bcdata because the data frame case still causes errors might need to rethink this through again as there definitely seem to be some advantages to the escape based approach.

hadley:
Ok, now fully reverted to the previous approach, but provided a better error in the default case, and ensured that `$` works with both local and database variables.

mgirlich (approved):

mgirlich:
Seems reasonable to roll back this change. It would be nice to be a bit stricter but considering the breakages of code it's probably not really worth it for now.

PR #1467, opened 2024-02-26
Switch methods to default, not DBIConnection

### LINKED ISSUE #1453: Check `table_path()` approach with backends
And figure out what fixes are needed.

hadley:
This makes dbplyr work with pool, and I think it's the right decision as these methods are primarily wrappers around DBI methods, so we should fall back to its defaults.

Closes #1453


hadley:
@mgirlich once you've had a chance to review these last 3 PRs, I'll start the release process.

PR #1475, opened 2024-03-07
Keep filtered aggregates with subsequent select in semi_join

### LINKED ISSUE #1474: Snowflake translation error: dropped filter with `anti_join`
Greetings, `dbplyr` friends.  I'd like to report an issue we're experiencing with Snowflake translations, related to `filter()` criteria dropped from lazy tables when `semi_join()`ed.  This seems specifically limited to `filter()` applied to columns which result from `summarize()`, that are not selected for inclusion in the result.

Here's a small example, reproducing the behavior:
```r
sim_1 <- dbplyr::lazy_frame(a = 5, b = 5, con = dbplyr::simulate_snowflake())
sim_2 <- dbplyr::lazy_frame(a = 5, b = 5, con = dbplyr::simulate_snowflake())

sim_2_transformed <- sim_2 %>% 
  group_by(a) %>% 
  summarize(group_count = n()) %>% 
  filter(group_count > 3) %>% 
  select(a)

join_result <- semi_join(sim_1, sim_2_transformed, by = 'a')
```

If we inspect the query produced for `sim_2_transformed`, it looks correct:
```
> sim_2_transformed
<SQL>
SELECT `a`
FROM `df`
GROUP BY `a`
HAVING (COUNT(*) > 3.0)
```

However, the join drops the `HAVING` criterion leading to an incorrect result; `join_result` is
```
> join_result
<SQL>
SELECT `df_LHS`.*
FROM `df` AS `df_LHS`
WHERE EXISTS (
  SELECT 1 FROM `df` AS `df_RHS`
  WHERE (`df_LHS`.`a` = `df_RHS`.`a`)
)
```

Notably, if I remove `select(a)` from the definition of `sim_2_transformed`, the `HAVING` clause is included as expected:
```
> join_result_without_final_select
<SQL>
SELECT `df`.*
FROM `df`
WHERE EXISTS (
  SELECT 1 FROM (
  SELECT `a`, COUNT(*) AS `group_count`
  FROM `df`
  GROUP BY `a`
  HAVING (COUNT(*) > 3.0)
) AS `RHS`
  WHERE (`df`.`a` = `RHS`.`a`)
)
```

Thanks for the continuing support of Snowflake backends; I'm happy to help with any testing that might be valuable.

Release version of dbplyr, 2.4.0.  cc @fh-mthomson

ejneer:
Fixes #1474

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dbplyr (development version)
 
+* `semi_join()` will no longer inline away an aggregate filter (i.e. `HAVING`
+  clause) that was followed by a `select()` (@ejneer, #1474)
+
 * Refined the `select()` inlining criteria to keep computed columns used to
   `arrange()` subqueries that are eliminated by a subsequent select (@ejneer,
   #1437).
--- a/R/lazy-select-query.R
+++ b/R/lazy-select-query.R
@@ -123,6 +123,9 @@ is_lazy_select_query_simple <- function(x,
   if (!is_empty(x$limit)) {
     return(FALSE)
   }
+  if (!is_empty(x$having)) {
+    return(FALSE)
+  }
 
   TRUE
 }
--- a/tests/testthat/_snaps/verb-joins.md
+++ b/tests/testthat/_snaps/verb-joins.md
@@ -200,6 +200,24 @@
           (`df_RHS`.`b` = 2)
       )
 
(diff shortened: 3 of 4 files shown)


hadley:
This seems like a low risk change (since we're choosing not to do a simplification in one case) and I'm about to release, so I've merged it.  Thanks for the fix!

PR #1501, opened 2024-05-15
Sql server count big

edward-burn:
Use COUNT_BIG instead of COUNT for sql server so as to avoid overflow error for very large tables

### DIFF
--- a/R/backend-mssql.R
+++ b/R/backend-mssql.R
@@ -416,7 +416,7 @@ simulate_mssql <- function(version = "15.0") {
   sql_variant(
     mssql_scalar,
     sql_translator(.parent = base_odbc_agg,
-      n          = function() sql("BIG_COUNT(*)"),
+      n          = function() sql("COUNT_BIG(*)"),
       sd            = sql_aggregate("STDEV", "sd"),
       var           = sql_aggregate("VAR", "var"),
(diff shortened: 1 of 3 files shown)


R/backend-mssql.R
```diff
@@ -425,6 +425,7 @@ simulate_mssql <- function(version = "15.0") {
   sql_variant(
     mssql_scalar,
     sql_translator(.parent = base_odbc_agg,
+      n           = function() sql("COUNT_BIG(*)"),
```
hadley, line 428: Could you please include a comment with a link to the docs?

tests/testthat/test-backend-mssql.R
```diff
@@ -400,6 +400,12 @@ test_that("row_number() with and without group_by() and arrange(): unordered def
   expect_snapshot(mf %>% arrange(y) %>% mutate(rown = row_number()))
 })
 
+test_that("count_big", {
+  mf <- lazy_frame(x = c(1:5), y = c(rep("A", 5)), con = simulate_mssql())
+  expect_snapshot(count(mf))
+  expect_snapshot(tally(mf))
```
hadley, line 406: I think one snapshot test is adequate here

edward-burn: Makes sense. Have removed the second one

PR #1550, opened 2024-10-30
various small translate-sql vignette corrections

bbolker:
I hope all these changes are correct and acceptable.

One change I didn't make because I wasn't sure: on the current line 52, the comment says "In SQLite variable names are escaped by double quotes:", but the printed output shows single back-quotes instead (`` rather than "") ... ??


simonpcouch (approved):
Thanks so much, @bbolker. These changes are spot-on; I'll open a separate issue about the SQLite variable name escaping after merging.

Issue #1546, opened 2024-10-21
`backend-*` spring cleaning for erroring code


simonpcouch:
Will start out with just one backend as a demo, but imagining one PR per checkbox here (applying to all backends, rather than one PR per backend with all of the following complete in each. Happy to adjust here.):

* [x] #1552
	* ...to snapshot when testing the error message
	* ...to classed error when testing an error from another package or an error thrown from a widely-used helper
	* ...to `expect_no_error()` (or no test) when `regex = NA`
	* Analogous story for `expect_warning()` and `expect_message()`.
* [x] #1555
* [ ] Transition erroring code to cli
	* Update snapshots
* [ ] Test untested erroring code

PR #1554, opened 2024-10-31
use type checkers in `backend-snowflake.R`

simonpcouch:
Related to #1546 but doesn't close. Speccing out the changes in `backend-snowflake.R` that would be applied for each backend to address the "use type checkers" bullet in the linked issue.

My sense here is that we ought not to be checking arguments that are injected via `!!` straight into `translate_sql()` or `sql_expr()`?

### DIFF
--- a/R/backend-snowflake.R
+++ b/R/backend-snowflake.R
@@ -27,6 +27,7 @@ sql_translation.Snowflake <- function(con) {
       paste = snowflake_paste(" "),
       paste0 = snowflake_paste(""),
       str_c = function(..., sep = "", collapse = NULL) {
+        check_string(sep)
         if (!is.null(collapse)) {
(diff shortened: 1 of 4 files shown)


R/backend-snowflake.R
```diff
@@ -111,6 +115,8 @@ sql_translation.Snowflake <- function(con) {
         sql_expr(EXTRACT(DAY %FROM% !!x))
       },
       wday = function(x, label = FALSE, abbr = TRUE, week_start = NULL) {
+        check_bool(label)
+        check_bool(abbr)
```
hadley, line 119: `check_number_whole(week_start, allow_null = TRUE)`?

hadley (approved):
Looks great!

PR #1556, opened 2024-11-05
use type checkers in `backend-*.R` files

### LINKED ISSUE #1555: Use type checkers

simonpcouch:
Closes #1555. As in #1554, didn't apply checks to arguments that are passed directly to `sql_*()` functions.

### DIFF
--- a/R/backend-mssql.R
+++ b/R/backend-mssql.R
@@ -634,8 +634,6 @@ mssql_bit_int_bit <- function(f) {
 
 #' @export
 `db_sql_render.Microsoft SQL Server` <- function(con, sql, ..., cte = FALSE, use_star = TRUE) {
-  check_unsupported_arg(cte, allowed = FALSE)
-  check_unsupported_arg(use_star, allowed = TRUE)
   # Post-process WHERE to cast logicals from BIT to BOOLEAN
   sql$lazy_query <- purrr::modify_tree(
     sql$lazy_query,
--- a/R/backend-mssql.R
+++ b/R/backend-mssql.R
@@ -107,11 +107,14 @@ simulate_mssql <- function(version = "15.0") {
                                                     conflict = c("error", "ignore"),
                                                     returning_cols = NULL,
                                                     method = NULL) {
+  # https://stackoverflow.com/questions/25969/insert-into-values-select-from
+  conflict <- rows_check_conflict(conflict)
+
+  check_character(returning_cols, allow_null = TRUE)
+
   check_string(method, allow_null = TRUE)
   method <- method %||% "where_not_exists"
   arg_match(method, "where_not_exists", error_arg = "method")
-  # https://stackoverflow.com/questions/25969/insert-into-values-select-from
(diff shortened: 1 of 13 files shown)


simonpcouch:
Just confirmed that the `postgres` GHA failure [can be replicated on `main`](https://github.com/tidyverse/dbplyr/actions/runs/11630332240)—will go ahead and request review.

R/backend-mssql.R
```diff
@@ -107,6 +107,7 @@ simulate_mssql <- function(version = "15.0") {
                                                     conflict = c("error", "ignore"),
                                                     returning_cols = NULL,
                                                     method = NULL) {
+  check_string(method, allow_null = TRUE)
```
hadley, line 110: IMO the order of argument checks should match the order of the arguments.

Also can we say anything about the type of `returning_cols`?

R/backend-mssql.R
```diff
@@ -333,6 +335,7 @@ simulate_mssql <- function(version = "15.0") {
       second = function(x) sql_expr(DATEPART(SECOND, !!x)),
 
       month = function(x, label = FALSE, abbr = TRUE) {
+        check_bool(label)
```
hadley, line 338: I'd be tempted to move the check for `check_unsupported_arg()` next to this check.

R/backend-mssql.R
```diff
@@ -373,27 +378,16 @@ simulate_mssql <- function(version = "15.0") {
...
       date_count_between = function(start, end, precision, ..., n = 1L){
-
         check_dots_empty()
-        if (precision != "day") {
-          cli_abort("{.arg precision} must be {.val day} on SQL backends.")
-        }
-        if (n != 1) {
-          cli_abort("{.arg n} must be {.val 1} on SQL backends.")
-        }
+        check_unsupported_arg(precision, allowed = "day")
```
hadley, line 382: Nice!

R/backend-mssql.R
```diff
@@ -637,6 +631,8 @@ mssql_bit_int_bit <- function(f) {
 
 #' @export
 `db_sql_render.Microsoft SQL Server` <- function(con, sql, ..., cte = FALSE, use_star = TRUE) {
+  check_unsupported_arg(cte, allowed = FALSE)
```
hadley, line 634: I think you shouldn't check those here because we're still calling the next method (which might support them)?

hadley (approved):
This is going to be a very nice improvement to the user experience.

R/backend-mssql.R
simonpcouch: Oh, interesting. I moved that `check_unsupported_arg()` call to be evaluated unconditionally before realizing that the only allowed value for `abbr` is `FALSE` and the default is `TRUE`. Bringing the check out of `if (label)` would mean that `month(x)` would fail by default on SQL server. I opted not to make your suggested change here and will leave this as unresolved for a day or two to give you a chance to reply before merging.

PR #1558, opened 2024-11-21
Fix issue when using CTE and UNION, refs #1557

### LINKED ISSUE #1557: dbplyr::sql_render generates error when using CTEs and dplyr::union
I ran into an error when attempting union some tables with the CTE option turned on.

``` r
tib <- tibble::tibble(a = c(1))
con <- dbplyr::simulate_dbi()
t <- dbplyr::tbl_lazy(tib, con = con)
tu <- dplyr::union(t, t)
dbplyr::sql_render(tu, sql_options = dbplyr::sql_options(cte = TRUE))
#> Error in qry$unions$table[[i]]: subscript out of bounds
```

<sup>Created on 2024-11-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I doubt it is relevant to the issue, but I'm using a combination of simulated DBI backends and lazy tables run through dplyr verbs to build up SQL queries.  I then use `dbplyr::sql_render` to produce SQL statements.  I was experimenting with using CTEs and when I set cte = TRUE, I got this error.

aguynamedryan:
Fixes #1557

### DIFF
--- a/R/query-set-op.R
+++ b/R/query-set-op.R
@@ -110,7 +110,7 @@ flatten_query.union_query <- function(qry, query_list, con) {
   query_list_new <- flatten_query(x, query_list, con)
   qry$x <- get_subquery_name(x, query_list_new)
 
-  for (i in vctrs::vec_seq_along(qry$unions)) {
+  for (i in vctrs::vec_seq_along(qry$unions$table)) {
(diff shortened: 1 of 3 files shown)


hadley:
That doesn't look like a correct fix to me.

PR #1561, opened 2024-11-29
Better cross-package links

hadley:
With dev roxygen2


hadley:
Looks like I fixed this elsewhere.

PR #1567, opened 2024-12-29
update bit64::as.integer64 use in test

### LINKED ISSUE #1564: snap change for postgres
Seeing on GHA re: column casts with PostgreSQL:

```
── Failure (test-backend-postgres.R:341:3): casts `y` column for local df ──────
... %>% collect() (`actual`) not equal to `out` (`expected`).

actual vs expected
                val
  actual[1, ]    10
- actual[2, ]    20
+ expected[2, ]  10

`actual$val`:   "10" "20"
`expected$val`: "10" "10"

── Failure (test-backend-postgres.R:367:3): casts `y` column for local df ──────
tbl(con, "df_x") %>% collect() (`actual`) not equal to `out` (`expected`).

actual vs expected
                val
  actual[1, ]    10
- actual[2, ]    20
+ expected[2, ]  10

`actual$val`:   "10" "20"
`expected$val`: "10" "10"
```

Here's the last run where that test [passed](https://github.com/tidyverse/dbplyr/actions/runs/11620396513/job/32362083292) on 10/31 and the first where it [failed](https://github.com/tidyverse/dbplyr/actions/runs/11630332240) on 11/06. Both have PostgreSQL 17.0 (Debian 17.0-1.pgdg120+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit, on Ubuntu 22.04.

I was able to replicate on Linux with `pak::local_install_dev_deps()`, then mirroring the GHA service with:

```sh
sudo docker run --name postgres-dev \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_PASSWORD=password \
  -e POSTGRES_DB=test \
  -p 5432:5432 \
  -d postgres

sudo docker exec postgres-dev pg_isready -U postgres
```

and:

```r
con <- DBI::dbConnect(RPostgres::Postgres(),
  dbname = "test",
  host = "localhost",
  port = 5432,
  user = "postgres",
  password = "password"
)
```

Related to #965--I don't think the fix here is just a snap update.

edward-burn:
fixes #1564 

@simonpcouch @hadley the check is failing for other reasons I believe - it seems that the error "vector memory limit of 16.0 Gb reached, see mem.maxVSize()" on macos came between https://github.com/tidyverse/dbplyr/actions/runs/11802007517/job/32876673073 and https://github.com/tidyverse/dbplyr/actions/runs/12089965184/job/33716160658

### DIFF
--- a/.github/workflows/test-coverage.yaml
+++ b/.github/workflows/test-coverage.yaml
@@ -6,7 +6,9 @@ on:
   pull_request:
     branches: [main, master]
 
-name: test-coverage
+name: test-coverage.yaml
+
+permissions: read-all
 
 jobs:
   test-coverage:
@@ -15,7 +17,7 @@ jobs:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - uses: r-lib/actions/setup-r@v2
         with:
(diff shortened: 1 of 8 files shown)


simonpcouch (approved):
Ha, nice catch—just a bad vector recycle. Thumbs up from me, though I'll defer to Hadley on merging.

edward-burn:
@hadley @simonpcouch I took another look at this and added some additional commits to try and get everything passing
- update another snapshot test - this one is for pillar (default seems now to be 20 rows for printing so changed test to reflect this)
- update the test coverage file

PR #1580, opened 2025-02-07
Fix copy_to overwrite flag

### LINKED ISSUE #1535: copy_to(overwrite=TRUE) not working in duckdb and dbplyr
It works fine for a dataframe, but not for a lazy_tbl.

``` r
library(duckdb)
#> Loading required package: DBI
library(dplyr)
#> Warning: package 'dplyr' was built under R version 4.2.3
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
con <- dbConnect(duckdb())

copy_to(con, mtcars, "mtcars", temporary = FALSE)
copy_to(con, mtcars, "mtcars", overwrite = TRUE, temporary = FALSE)

mtcars_sql <- tbl(con, "mtcars")

copy_to(con, mtcars_sql, "mtcars2", temporary = FALSE)
copy_to(con, mtcars_sql, "mtcars2", overwrite = TRUE, temporary = FALSE)
#> Error in `db_compute()`:
#> ! Can't copy query to table mtcars2.
#> Caused by error in `db_save_query.DBIConnection()`:
#> ! Can't save query to table mtcars2.
#> ℹ Using SQL: CREATE TABLE mtcars2 AS SELECT * FROM mtcars
#> Caused by error:
#> ! rapi_execute: Failed to run query
#> Error: Catalog Error: Table with name "mtcars2" already exists!

dbDisconnect(con)
```

<sup>Created on 2024-08-14 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

liudvikasakelis:
Fixes #1535. 
While the report mentions duckdb, I chose to write test in SQLite due to there being no other duckdb tests yet. Issue seems backend agnostic anyway. 
Will be grateful for reviews :)


liudvikasakelis:
After some test wrangling, I think this is ready for review. In summary: 
* proposed test case I think captures the gist of the bug, even if in SQLite. This case fails without the fix and passes with it.  
* reprex provided in the issue #1535 succeeds without error with provided fix (tested manually)
* errors seen in GH automated checks:
  * test-tbl-sql.R:33:3 - already present in commit 57269309 (tested locally). It's a simple fix but I'm unsure if this PR is the best place :shrug: 
  * test-backend-postgres.R:341:3 - already present in commit 57269309 (tested locally)
  * test-backend-postgres.R:367:3 - already present in commit 57269309 (tested locally)
  * test-coverage 'has been automatically failed' due to deprecated version

i.e. the fix addresses the issue and doesn't introduce more bugs, to the best of my knowledge

PR #1582, opened 2025-02-10
don't call `all_calls()` and `all_names()` recursively

EmilHvitfeldt:
Since `all_calls()` and `all_names()` are run recursively, they complain about longer expressions. This PR rewrites them with the base functions `all.names()` and `all.vars()` to avoid this problem

Before This PR

Kapture 2025-02-10 at 10 46 11

With this PR

``` r
long <- rlang::parse_quo(paste(rep(letters, 50), collapse = " + "), env = globalenv())

dbplyr:::all_names(long)
#>  [1] "a" "b" "c" "d" "e" "f" "g" "h" "i" "j" "k" "l" "m" "n" "o" "p" "q" "r" "s"
#> [20] "t" "u" "v" "w" "x" "y" "z"
dbplyr:::all_calls(long)
#> [1] "+"
```

### DIFF
--- a/R/translate-sql.R
+++ b/R/translate-sql.R
@@ -255,8 +255,7 @@ all_calls <- function(x) {
   if (is_quosure(x)) return(all_calls(quo_get_expr(x)))
   if (!is.call(x)) return(NULL)
 
-  fname <- as.character(x[[1]])
-  unique(c(fname, unlist(lapply(x[-1], all_calls), use.names = FALSE)))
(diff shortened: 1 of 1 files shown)


hadley:
Hmmm, this appears to break a test for `escape()`, i.e. it's not giving the same results. If you're still interested in a fix, I suspect claude can probably convert the existing implementation from recursion to iteration and preserve the existing behaviour.

PR #1608, opened 2025-05-13
Infix `$` translation for nested values in Snowflake

jsowder:
Nested values in Snowflake are accessed via `:` rather than `.`.

https://docs.snowflake.com/en/user-guide/querying-semistructured#traversing-semi-structured-data


hadley:
Thank you!

Issue #1583, opened 2025-02-19
DatabaseConnector support


ablack3:
DatabaseConnector is DBI driver backend that supports multiple database systems using jdbc.

We would like to make sure dplyr sql translation dispatches correctly when using DatabaseConnector connections with dbplyr. 

The approach we are trying is to define a translation environment for these connection in DatabaseConnector as follows.

```
#' @export
#' @importFrom dbplyr sql_translation 
sql_translation.DatabaseConnectorJdbcConnection <- function(con) {
  
  switch(dbms(con),
     "oracle" = utils::getFromNamespace("sql_translation.Oracle", "dbplyr")(con),
     "postgresql" = utils::getFromNamespace("sql_translation.PqConnection", "dbplyr")(con),
     "redshift" = utils::getFromNamespace("sql_translation.RedshiftConnection", "dbplyr")(con),
     "sql server" = utils::getFromNamespace("sql_translation.Microsoft SQL Server", "dbplyr")(con),
     "bigquery" = utils::getFromNamespace("sql_translation.BigQueryConnection", "bigrquery")(con),
     "spark" = utils::getFromNamespace("sql_translation.Spark SQL", "dbplyr")(con),
     "snowflake" = utils::getFromNamespace("sql_translation.Snowflake", "dbplyr")(con),
     "synapse" = utils::getFromNamespace("sql_translation.Microsoft SQL Server", "dbplyr")(con),
     rlang::abort("Sql dialect is not supported!")) 
}
```

Is this the best way to add dplyr support or should we add something to dbplyr to dispatch the correct translation environment. 

We also found we needed to write the method sql_query_select.DatabaseConnectorJdbcConnection as well.

hadley:
See #1624

PR #1616, opened 2025-08-15
Add DatabaseConnector backend #1583

ablack3:
This PR adds DatabaseConnector Support to dbplyr. 
DatabaseConnector is a DBI-compliant database backend that supports multiple database systems. It uses JDBC drivers to establish connections, and the connection object includes a `dbms` slot that specifies the database type. This backend enables dplyr's SQL translations to work across all database supported platforms by reusing translations from other backends.


hadley:
Lets tackle this again after #1624 since it will be much much simpler.

PR #1620, opened 2025-10-01
add support for `DISTINCT ON`

lschneiderbauer:
This PR is a first draft to address https://github.com/duckdb/duckdb-r/issues/384, and add support for the usage of "DISTINCT ON" when using `distinct(..., .keep_all = TRUE)` which is a SQL variant supported by PostgreSQL, and DuckDB, see e.g. https://duckdb.org/docs/stable/sql/query_syntax/select.html#distinct-on-clause. Currently, `.keep_all = TRUE` is implemented using window functions. Using `DISTINCT ON` instead promises a performance boost of that operation.

I came to the conclusion that this cannot be addressed within an external dbplyr backend, but it requires a minor modification of the `lazy_select_query` data structure itself:
Currently, a `lazy_select_query` supports a `distinct` state, which can be either `TRUE` or `FALSE` (corresponding to a normal `SELECT` vs. a `SELECT DISTINCT`.
The basic idea of this PR is to add a third state to the `distinct` attribute which represents a list of columns that belong to the `SELECT DISTICT ON (...)` clause.
Dbplyr backends can opt in to make use of `DISTINCT ON` via implementing a method of the new generic `supports_distinct_on()` that returns `TRUE`.

### Open issues

1. A major issue is the handling of the order specification. `DISTINCT ON` uses `ORDER BY` to specify an ordering. That is also a reason why `ORDER BY` is allowed in subqueries in PostgreSQL and DuckDB. As far as I can see, dbplyr currently forbids the `ORDER BY` statement in subqueries. I did not investigate yet, if that can be modified easily, or even can be changed at all. In any case, from the user-perspective `window_order()` would probably still be the right verb to specify the order.
2. The syntax highlighting of the currently generated SQL code is incorrect, that's probably because `sql_clause()` is not used correctly.
3. Having one `distinct` attribute that holds either `TRUE`, `FALSE` or a column list representation leads to a lot of required case distinction checks in the code, which are rather unpleasent to read and complicate the code. There is probably a way to model this in a more streamlined fashion.

I would appreciate feedback to the open issues as well as the already existing code. It might not be the right approach after all.

### DIFF
--- a/R/lazy-select-query.R
+++ b/R/lazy-select-query.R
@@ -168,6 +168,9 @@ print.lazy_select_query <- function(x, ...) {
   if (length(select)) {
     cat_line("Select:   ", named_commas(select))
   }
+  if (is_lazy_sql_part(x$distinct)) {
+    cat_line("Dist on:  ", named_commas(x$distinct))
+  }
   if (length(x$where)) {
     cat_line("Where:    ", named_commas(x$where))
   }


lschneiderbauer:
> This seems like a reasonable approach to me. Are you still interested in working on this? If so, I can give you a detailed review. If not, I'm happy to finish it off.

Both options are fine with me. Whatever is more convenient to you.

R/lazy-select-query.R
```diff
@@ -19,7 +19,10 @@ lazy_select_query <- function(x,
   # stopifnot(is.character(group_by))
   stopifnot(is_lazy_sql_part(order_by))
   check_number_whole(limit, allow_infinite = TRUE, allow_null = TRUE)
-  check_bool(distinct)
+  # distinct = FALSE     -> no distinct
```
hadley: This is a great comment that helps me understand the intent of the rest of the code 😄 

But it does illustrates a problem with `lazy_select_query()` — it isn't properly documented and thus it's hard to tell what the types of the inputs are. You don't need to fix it here, but it might be a useful follow up PR.

R/lazy-select-query.R
```diff
@@ -19,7 +19,10 @@ lazy_select_query <- function(x,
...
+  # distinct = TRUE      -> distinct over all columns
+  # distinct = columns   -> DISTINCT ON (...)
+  stopifnot(is.logical(distinct) || is_lazy_sql_part(distinct))
```
hadley:
```suggestion
  stopifnot(is_bool(distinct) || is_lazy_sql_part(distinct))
```

Since `is.logical()` allows vectors and `NA`, neither of which we support.

R/lazy-select-query.R
```diff
@@ -166,8 +173,24 @@ sql_build.lazy_select_query <- function(op, con, ..., sql_options = NULL) {
 
   alias <- remote_name(op$x, null_if_local = FALSE) %||% unique_subquery_name()
   from <- sql_build(op$x, con, sql_options = sql_options)
+
+  selects <- op$select
+
+  if (!is.logical(op$distinct)) {
```
hadley: I think it would be more clear to say what `op$distinct` is, rather than what it's not.

hadley (commented):
Partial review; I'll complete after rebasing so it's styled with air.

R/sql-clause.R
```diff
@@ -53,10 +53,19 @@ sql_clause_select <- function(
     top <- as.integer(top)
   }
 
+  sql_distinct <-
+    if (isTRUE(distinct)) {
+      " DISTINCT"
```
hadley: Our usual style would be to put `sql_distinct <- ` in each branch or extract out a helper function. I'd probably lean towards a helper function since you could inline that in `clause` below.

R/verb-distinct.R
```diff
@@ -148,12 +160,25 @@ quo_is_variable_reference <- function(quo) {
 }
 
 
-add_distinct <- function(.data) {
+add_distinct <- function(.data, distinct) {
   lazy_query <- .data$lazy_query
 
+  distinct <-
```
hadley: Same fix here

R/verb-distinct.R
```diff
@@ -173,6 +198,11 @@ add_distinct <- function(.data) {
     return(out)
   }
 
-  lazy_query$distinct <- TRUE
+  if (!is.logical(distinct)) {
+    lazy_query$distinct <- new_lazy_select(distinct)
```
hadley: Echoing one of my earlier comments, I think things might be simpler if you did the conversion to `new_lazy_select()` later.

hadley (commented):
The code seems solid to me overall — just lots of small style questions. Thanks again for working on this!

lschneiderbauer:
> The code seems solid to me overall — just lots of small style questions. Thanks again for working on this!

Thanks for the review!

The ordering issue is still open (from the MR description):

> A major issue is the handling of the order specification. DISTINCT ON uses ORDER BY to specify an ordering. That is also a reason why ORDER BY is allowed in subqueries in PostgreSQL and DuckDB. As far as I can see, dbplyr currently forbids the ORDER BY statement in subqueries. I did not investigate yet, if that can be modified easily, or even can be changed at all. In any case, from the user-perspective window_order() would probably still be the right verb to specify the order.

Do you have thoughts on that?

PR #1644, opened 2025-12-01
add str_ilike support

### LINKED ISSUE #1628: Add `str_ilike()` translations

### LINKED ISSUE #1630: align str_like with stringr 1.6.0
Related to #1628, would be good I think to align str_like with stringr 1.6.0 (deprecate ignore_case = TRUE)

``` r
library(stringr)

str_like(c("a", "b", "c"), "a", ignore_case = TRUE)
#> Warning: The `ignore_case` argument of `str_like()` is deprecated as of stringr 1.6.0.
#> ℹ `str_like()` is always case sensitive.
#> ℹ Use `str_ilike()` for case insensitive string matching.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
#> [1]  TRUE FALSE FALSE
str_like(c("a", "b", "c"), "a", ignore_case = FALSE)
#> [1]  TRUE FALSE FALSE

options(dplyr.strict_sql = TRUE)
translate_sql(str_like(var, "a", ignore_case = TRUE), con = simulate_postgres())
#> <SQL> `var` ILIKE 'a'
translate_sql(str_like(var, "a", ignore_case = FALSE), con = simulate_postgres())
#> <SQL> `var` LIKE 'a'
```

<sup>Created on 2025-11-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

edward-burn:
@hadley this hopefully closes #1628 and closes #1630

### DIFF
--- a/tests/testthat/_snaps/backend-.md
+++ b/tests/testthat/_snaps/backend-.md
@@ -50,13 +50,6 @@
 # can only translate case sensitive str_like
 
     Code
-      test_translate_sql(str_like(x, "abc"))
-    Output
-      <SQL> `x` LIKE 'abc'
-
----
-
-    Code
       test_translate_sql(str_like(x, "abc", ignore_case = FALSE))
     Condition
       Warning:
@@ -65,9 +58,6 @@
       i Use `str_ilike()` for case insensitive string matching.
     Output
       <SQL> `x` LIKE 'abc'
-
----
-
     Code
       test_translate_sql(str_like(x, "abc", ignore_case = TRUE))
     Condition
@@ -78,9 +68,6 @@
       Error in `str_like()`:
       ! Backend does not support case insensitive `str_like()`.
       i Use `tolower()` on both arguments to achieve a case insensitive match.
-
----
-
     Code
       test_translate_sql(str_ilike(x, "abc"))
     Condition
--- a/tests/testthat/test-backend-.R
+++ b/tests/testthat/test-backend-.R
@@ -125,20 +125,13 @@ test_that("lead and lag translate n to integers", {
 
 test_that("can only translate case sensitive str_like", {
   local_con(simulate_dbi())
-  expect_snapshot(
-    test_translate_sql(str_like(x, "abc"))
-  )
-  expect_snapshot(
+  expect_equal(test_translate_sql(str_like(x, "abc")), sql("`x` LIKE 'abc'"))
+
+  expect_snapshot(error = TRUE, {
     test_translate_sql(str_like(x, "abc", ignore_case = FALSE))
-  )
-  expect_snapshot(
-    test_translate_sql(str_like(x, "abc", ignore_case = TRUE)),
-    error = TRUE
-  )
-  expect_snapshot(
-    test_translate_sql(str_ilike(x, "abc")),
-    error = TRUE
-  )
+    test_translate_sql(str_like(x, "abc", ignore_case = TRUE))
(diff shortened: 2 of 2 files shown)


R/backend-postgres.R
```diff
@@ -137,16 +137,26 @@ sql_translation.PqConnection <- function(con) {
...
-      str_like = function(string, pattern, ignore_case = TRUE) {
-        check_bool(ignore_case)
-        if (isTRUE(ignore_case)) {
-          sql_expr(!!string %ILIKE% !!pattern)
-        } else {
-          sql_expr(!!string %LIKE% !!pattern)
+      str_like = function(string, pattern, ignore_case = deprecated()) {
+
+        if (lifecycle::is_present(ignore_case)) {
+          lifecycle::deprecate_warn(
```
hadley: Given that you've had to copy this three times, it would make sense to extract it into a function.

hadley (commented):
Thanks for working on this! More comments below.

R/backend-.R
```diff
@@ -285,7 +285,19 @@ base_scalar <- sql_translator(
...
+        what = "str_like(ignore_case)",
+        details = c(
+          "`str_like()` is always case sensitive.",
+          "Use `str_ilike()` for case insensitive string matching."
+        )
+      )
+    }
+
     if (isTRUE(ignore_case)) {
       cli_abort(c(
```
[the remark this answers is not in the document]
edward-burn: @hadley this was just to match the current stringr behaviour (below), but happy to remove it if you think that's preferrable? (Previous str_like changes in dbplyr have only been in the dev version)

``` r
library("stringr")
packageVersion("stringr")
#> [1] '1.6.0'
str_like("A", "a", ignore_case = TRUE)
#> Warning: The `ignore_case` argument of `str_like()` is deprecated as of stringr 1.6.0.
#> ℹ `str_like()` is always case sensitive.
#> ℹ Use `str_ilike()` for case insensitive string matching.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
#> [1] TRUE
```

<sup>Created on 2025-12-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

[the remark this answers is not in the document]
hadley: Right but the dbplyr code errors currently.

R/backend-postgres.R
edward-burn: Sure - happy to add that if you are happy with this https://github.com/tidyverse/dbplyr/pull/1644#discussion_r2583847343 (if you prefer that it just errors for all backends when ignore_case = TRUE then this code wouldn't be needed)

tests/testthat/_snaps/backend-.md
```diff
@@ -52,11 +52,23 @@
     Code
       test_translate_sql(str_like(x, "abc", ignore_case = TRUE))
     Condition
+      Warning:
```
hadley: Concretely, this is what I'm concerned about — the function warns and then errors, which doesn't seem correct. I don't think you see that so obviously in your examples above because of the way deprecation warnings work.

It should either warn or error, not both.

hadley:
I simplified the tests a bit after extracting out a reusable function, since that made it more clear we don't need to test the individual implementations so heavily.

edward-burn:
Great, thanks so much!

PR #1652, opened 2025-12-03
redshift specific error

### LINKED ISSUE #1571: possible typo in error message for redshift translation of quantile() in mutate()
In the below redshift example, I get an error that the function is not supported for PostgreSQL, but I suppose it should intead say redshift?

``` r
lazy_frame(x = c(1,2), con = simulate_redshift()) %>% 
  mutate(
    q25 = quantile(x, 0.25, na.rm = TRUE),
    median = median(x, na.rm = TRUE),
    q75 = quantile(x, 0.75, na.rm = TRUE)) |> 
  show_query()
#> Error in `quantile()`:
#> ! Translation of `quantile()` in `mutate()` is not supported for
#>   PostgreSQL.
#> ℹ Use a combination of `summarise()` and `left_join()` instead:
#>   `df %>% left_join(summarise(<col> = quantile(x, 0.25, na.rm = TRUE)))`.
```

<sup>Created on 2025-01-06 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

edward-burn:
closes #1571

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # dbplyr (development version)
 
 * Corrected error message for `quantile()` and `median()` in `mutate()` on Redshift (@edward-burn, #1571).
+* All set operations now error if you pass extra arguments (instead of silently ignoring then) (#1585).
 * `str_flatten()` (#1540) and `n_distinct()` (#1579) now have an `na.rm` argument, which regularly warns when it's not `TRUE`.
 * `semi_join()` and `anti_join()` once again work with filtered windowed values (#1534, #1606).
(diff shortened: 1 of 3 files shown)


hadley:
You shouldn't need to add me to the repo, just check this box:

Screenshot 2025-12-03 at 11 14 12

edward-burn:
Ah, apparently I don't have this button because I have an organisation-owned fork (never knew that was a thing) - sorry to make this painful

hadley:
Ah that's a pain. But looks like I'm good to go now. Thanks!

PR #1656, opened 2025-12-03
Don't qualify `*` in single-table SELECTs

### LINKED ISSUE #1482: `head()` modifies `SELECT *`
When using `head()` on a tbl, the select all statement is modified. Example below. In the example this is not a problem itself, but when calling tables in a specific schema, the name of the table is added, but the name of the schema is not. (I'm not quite sure for which backends this exactly is a problem.)

``` r
suppressMessages(library(dbplyr))
suppressMessages(library(dplyr, warn.conflicts = FALSE))
df <- memdb_frame(x = runif(26), y = runif(26))
df |> show_query()
#> <SQL>
#> SELECT *
#> FROM `dbplyr_98Ik9vEYQa`
df |> head() |> show_query()
#> <SQL>
#> SELECT `dbplyr_98Ik9vEYQa`.*
#> FROM `dbplyr_98Ik9vEYQa`
#> LIMIT 6
```

<sup>Created on 2024-03-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### LINKED ISSUE #1485: `sql` in `in_catalog` breaks with `filter`
I am having trouble generating reproducible example because I don't know how to mock catalogs and table valued functions but effectively issue is:

```r
library(dbplyr)
library(dplyr)

con <- simulate_odbc()

# this works fine, we have select * with no prefix; datawarehouse and dbo will have quotes in query but MyTableValuedFunction won't
tbl(con, in_catalog("Datawarehouse", "DBO", sql("MyTableValuedFunction(Param1, Param2)"))) %>%
  show_query

# now that I have filter it will try to prefix * like "MyTableValuedFunction(Param1, Param2)".*  with quotes and it won't work it needs an alias
tbl(con, in_catalog("Datawarehouse", "DBO", sql("MyTableValuedFunction(Param1, Param2)"))) %>%
  filter(getdate() > 1) %>%
  show_query
```
This worked before 2.4.0 so it appears fixes in 2.5.0 haven't covered everything.

### LINKED ISSUE #1577: Adding filter creates a qualified select * which causes error
We use a DB2 flavored SQL database. Ever since {dbplyr} was updated past version 2.3.4, an error occurs in certain combinations of commands.

It appears this is due to a `SELECT` statement having a qualified name combined with `*`. If the quotes are removed or the entire qualifier with up to the catalog is provided, it works. However, I do not know how this can be enforced. Specifying specific columns also works as `*` is no longer used. 

I believe this is related in some capacity to this prior issue: https://github.com/tidyverse/dbplyr/issues/1480 

Perhaps there is a simple solution Ive missed to adjust quoting in column selection with `*`?

## Example 

Due to the nature of the system being DB2, it is hard to create a reprex. However, I have provided some pseudocode and example outputs to help elucidate the issue.

Setup Connection 
```r
library(dbplyr)
library(dplyr)
con <- DBI::dbConnect(odbc::odbc(),
               "DB2", # DSN for our DB2
                uid = Sys.getenv("myuid"),     
                pwd = Sys.getenv("mypwd"), timeout = 10)
```

**Works** - the select * does not have a quoted qualifier
```r

DBI::dbGetQuery(con, "SELECT MYCATALOG.MYSCHEMA.MYTABLE.*
FROM MYCATALOG.MYSCHEMA.MYTABLE.
WHERE MYCOLUMN != 'SOMEVALUE'"

```

**Fails** - due to a qualified select *
```r
mytbl <-tbl(con, in_catalog(sql("MYCATALOG"), sql("MYSCHEMA"), sql("MYTABLE")))
 mytb  %>%
  filter(MYCOLUMN != "SOMEVALUE") %>%
  show_query() # collect()
```

! ODBC failed with error 00000 from [IBM][CLI Driver][DB2/AIX64].
✖  SQL0204N "MYTABLE" is an undefined name.  SQLSTATE=42704

• <SQL> 'SELECT "MYTABLE".*
• FROM MYCATALOG.MYSCHEMA.MYTABLE
• WHERE ("MYCOLUMN " != 'SOMEVALUE''

hadley:
This seems like it should be safe, since there's no ambiguity when there's only a single table, and that's what `lazy_select_query` represents.

Fixes #1577. Fixes #1485. Fixes #1482.

@mgirlich do you have time to review a few dbplyr PRs?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 * `glue_sql2()` is now exported for building SQL strings with glue syntax and
   type markers. `build_sql()` is deprecated in favor of `glue_sql2()` (#1249).
+* `SELECT` statements that use a single table now use `*` rather than `TABLE.*` (#1485, #1577).
 * dbplyr 1e interfaces are now deprecated (#1197). Backend developers have had >2 years to update.
 * MySQL gains slightly better translation for `as.integer()` and `as.integer64()` (#1647).
 * Fixed snowflake translations that were being reported as unknown (@edward-burn, #1570). 
--- a/R/lazy-select-query.R
+++ b/R/lazy-select-query.R
@@ -199,7 +199,6 @@ sql_build.lazy_select_query <- function(op, con, ..., sql_options = NULL) {
     select = op$select,
     select_operation = op$select_operation,
     in_vars = op_vars(op$x),
-    table_alias = alias,
     con = con,
     use_star = sql_options$use_star
   )
@@ -227,7 +226,6 @@ get_select_sql <- function(
   select,
   select_operation,
   in_vars,
-  table_alias,
   con,
   use_star
 ) {
@@ -250,13 +248,13 @@ get_select_sql <- function(
 
   if (use_star && is_select_identity(select, in_vars)) {
     out <- list(
-      select_sql = sql_star(con, table_alias),
+      select_sql = sql_star(con),
       window_sql = character()
     )
     return(out)
   }
 
-  select <- select_use_star(select, in_vars, table_alias, con, use_star)
+  select <- select_use_star(select, in_vars, con, use_star)
 
   # translate once just to register windows
   win_register_activate()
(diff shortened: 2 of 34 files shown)


hadley:
New PR incoming shortly.

PR #1668, opened 2025-12-06
str_ilike for spark

edward-burn:
Also add str_ilike for spark


hadley:
@edward-burn FYI I've switched to using `str_glue()` everywhere to be consistent and easier to understand.

PR #1676, opened 2025-12-09
Move away from `sql_expr()` towards `sql_glue()`

hadley:
Gluing strings together is a better metaphor for this level of SQL translation. It also avoids calling functions that don't exist, hence eliminating the need for `globalVariables()`.

I had forgotten that `sql_expr()` automatically uppercases function names, so this also makes the R code closer to the generated SQL.

To do:

* [x] Better names than `glue_sql2()` (used in methods with `con` as the first argument) and `sql_glue()` (used in `sql_translation()`).

### DIFF
--- a/R/translate-sql-conditional.R
+++ b/R/translate-sql-conditional.R
@@ -110,15 +110,11 @@ sql_if <- function(cond, if_true, if_false = quo(NULL), missing = quo(NULL)) {
   #   WHEN NOT <cond> THEN `if_false`
   #   ELSE `missing`
   # END
-
   if (!quo_is_null(if_false) && !identical(if_false, missing)) {
-    false_sql <- " WHEN NOT {enpared_cond} THEN {enpared_if_false}"
-    out <- paste0(out, false_sql)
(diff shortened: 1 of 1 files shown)


hadley:
@mgirlich no need for a review here but I wanted to make sure you at least saw this. I think standardising on glue syntax everywhere is a win because it makes the code more consistent and I think pasting togethers string is a better framework for SQL generation at this level.

R/translate-sql-aggregate.R
```diff
@@ -155,15 +154,11 @@ sql_quantile <- function(f, style = c("infix", "ordered"), window = FALSE) {
...
-    sql <- switch(
-      style,
-      infix = sql_call2(f, x, probs),
-      ordered = glue_sql2(
-        sql_current_con(),
-        sql_call2(f, probs),
-        " WITHIN GROUP (ORDER BY {x})"
-      )
-    )
+    if (style == "infix") {
```
hadley: I find this new translation much easier to understand

PR #1678, opened 2025-12-10
Add new `.sql` pronoun

### LINKED ISSUE #1117: Export a mask akin to rlang::.data for custom SQL functions
A useful feature of dbplyr is automatically passing through unknown calls straight into SQL, e.g. some flavors support `ANY_VALUE()` to take a value from a column arbitrarily during aggregation:

```r
conn |>
  tbl("my_schema.table") |>
  group_by(grp) |>
  summarize(some_x = ANY_VALUE(x)) |>
  ungroup() |>
  show_query()
```

Will give something like

```sql
SELECT grp, ANY_VALUE(x) AS some_x
FROM my_schema.tbl
GROUP BY grp
```

However, within packages, this presents the classic issue with `R CMD check` notes that `ANY_VALUE()` is an undefined function, from the static R checker's point of view.

rlang offers `.data` and `.env` masks for this purpose:

https://rlang.r-lib.org/reference/dot-data.html

Could dbplyr offer something similar, e.g. an `.sql` mask we could use as:

```r
conn |>
  tbl("my_schema.table") |>
  group_by(grp) |>
  summarize(some_x = .sql$ANY_VALUE(x)) |>
  ungroup() |>
  show_query()
```

hadley:
Fixes #1117


hadley:
@MichaelChirico I'd love your thoughts on this if you have a few minutes 😄

vignettes/translation-function.Rmd
```diff
@@ -16,25 +16,26 @@ options(tibble.print_min = 6L, tibble.print_max = 6L, digits = 3)
...
 library(dbplyr)
-library(dplyr)
+library(dplyr, warn.conflicts = FALSE)
 
-con <- simulate_dbi()
+lf <- lazy_frame(x = 1, y = 2, g = "a")
 ```
 
-`dbplyr::translate_sql()` powers translation of individual function calls, and I'll use it extensively in this vignette to show what's happening. You shouldn't need to use it ordinary code as dbplyr takes care of the translation automatically.
+In this vignette, I'll use `lazy_frame()` to create a toy lazy table that allows us to see the translation without needing to connect to a real database:
```
MichaelChirico: nit: It looks like we use "I'll use" after it's already used, maybe put the `lf <- ...` in the next cell?

vignettes/translation-function.Rmd
```diff
@@ -93,85 +94,93 @@ dbplyr no longer translates `%/%` because there's no robust cross-database trans
...
 
-Or use `||` for string concatenation (although most backends will translate `paste()` and `paste0()` for you):
+You can also use `str_like()` for this common case:
 
 ```{r}
-translate_sql(x %||% y, con = con)
+lf |> filter(str_like(x, "%foo%"))
+```
+
+You could use `||` for string concatenation, but in most cases it's more R-like to use `paste()` or `paste0()`:
```
MichaelChirico:
```suggestion
You could use `%||%` for string concatenation, but in most cases it's more R-like to use `paste()` or `paste0()`:
```

tests/testthat/test-verb-mutate.R
```diff
@@ -211,6 +211,17 @@ test_that("empty mutate returns input", {
   expect_equal(mutate(gf, !!!list()), gf)
 })
 
+test_that("can use .sql pronoun", {
+  lf <- lazy_frame(x = 1)
+
+  # Provides similar capability to existing escaping mechanisms, but is more
+  # convenient for package usage since you only need to globalVariables(".sql")
```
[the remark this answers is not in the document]
hadley: I contemplated that too. So I think your suggestion means I should do it.

PR #1680, opened 2025-12-11
Refactor `tbl_sql()`/`tbl_lazy()` creation

hadley:
Make the process more direct, more consistent, and involve fewer jumps back and forth between dplyr and dbplyr.

Includes adding a new `con` field to `tbl_lazy` since that's what we use 99% of the time. We keep a `src` around for backward compatibility.

### DIFF
--- a/R/tbl-lazy.R
+++ b/R/tbl-lazy.R
@@ -109,7 +109,7 @@ names.tbl_lazy <- function(x) {
       i = "Did you mean {.fn colnames}?"
     ))
   }
-  names(unclass(x))
+  NextMethod()
 }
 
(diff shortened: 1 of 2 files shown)


R/tbl-lazy.R
```diff
@@ -114,12 +109,12 @@ names.tbl_lazy <- function(x) {
       i = "Did you mean {.fn colnames}?"
     ))
   }
-  NextMethod("names")
+  names(unclass(x))
```
hadley: Without this, `names.tbl_lazy` gets called twice. Presumably some `NextMethod()` weirdness.

PR #1715, opened 2025-12-18
Inline filter after a join

### LINKED ISSUE #722: Produce less nested SQL for joins

hadley:
@mgirlich could you please take a look at this? Probably easiest to see exactly what's happening in https://github.com/tidyverse/dbplyr/commit/94e3206ca971e4182cdb8eb4879b7096fc3a0269; the other commit is just some refactoring to make the changes a bit easier.

Fixes #722

### DIFF
--- a/R/lazy-join-query.R
+++ b/R/lazy-join-query.R
@@ -194,11 +194,24 @@ sql_build.lazy_multi_join_query <- function(op, con, ..., sql_options = NULL) {
     }
   )
 
-  # Column names in WHERE don't need to be qualified with table names/aliases
-  # because we know that the join has already generated aliases for ambiguous
-  # variables
+  join_vars <- sql_multi_join_vars(
+    con,
+    op$vars,
+    table_vars,
+    use_star = FALSE,
+    qualify_all_columns = sql_options$qualify_all_columns
+  )
+  # WHERE happens after SELECT, but columns names are disambiguated using
+  # SELECT expressions, so need to backtransform
+  where <- lapply(op$where, \(expr) {
+    replace_sym(
+      expr,
+      names(join_vars),
+      lapply(unname(join_vars), \(x) sql(x[[1]]))
+    )
+  })
   where_sql <- translate_sql_(
-    op$where,
+    where,
     con = con,
     context = list(clause = "WHERE")
   )
--- a/R/verb-filter.R
+++ b/R/verb-filter.R
@@ -74,9 +74,11 @@ add_filter <- function(.data, dots) {
     lazy_query
   } else if (filter_can_inline(dots, lazy_query, con)) {
     # WHERE happens before SELECT so can't refer to aliases
-    dots <- rename_aliases(dots, lazy_query$select)
-
     # might be either a lazy_select_query or a lazy_multi_join_query
+    if (is_lazy_select_query(lazy_query)) {
+      dots <- rename_aliases(dots, lazy_query$select)
+    }
+
     lazy_query$where <- c(lazy_query$where, dots)
     lazy_query
   } else {
--- a/tests/testthat/_snaps/backend-mssql.md
+++ b/tests/testthat/_snaps/backend-mssql.md
@@ -759,6 +759,8 @@
       EOF within quoted string
       Warning in `scan()`:
       EOF within quoted string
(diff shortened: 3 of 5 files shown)


R/lazy-join-query.R
```diff
@@ -190,11 +194,34 @@ sql_build.lazy_multi_join_query <- function(op, con, ..., sql_options = NULL) {
...
+    use_star = FALSE,
+    qualify_all_columns = sql_options$qualify_all_columns
+  )
+  # WHERE happens after SELECT, but columns names are disambiguated using
+  # SELECT expressions, so need to backtransform
+  where <- lapply(op$where, \(expr) {
+    replace_sym(
+      expr,
+      names(join_vars),
+      lapply(unname(join_vars), \(x) sql(x[[1]]))
```
mgirlich: 1. `x[[1]]` is a bit difficult to grasp. You need to know how `join_vars` looks like.
2. This line could be put outside of the apply to make this more readable (and make this a bit faster)

R/lazy-join-query.R
```diff
@@ -190,11 +194,34 @@ sql_build.lazy_multi_join_query <- function(op, con, ..., sql_options = NULL) {
...
+  join_vars <- sql_multi_join_vars(
+    con,
+    op$vars,
+    table_vars,
+    use_star = FALSE,
+    qualify_all_columns = sql_options$qualify_all_columns
+  )
+  # WHERE happens after SELECT, but columns names are disambiguated using
+  # SELECT expressions, so need to backtransform
+  where <- lapply(op$where, \(expr) {
```
mgirlich: Maybe refactor this into a function like `backtransform_where_table_vars()`

mgirlich (approved):

R/lazy-join-query.R
hadley: Now think this PR is likely to be the right basic approach. I'll need to wrap up my work on dbplyr early in the new year and this PR does the job.

PR #1741, opened 2025-12-24
Refactor `sql_clause()` and escaping

hadley:


hadley:
Too much here, so I'll extract out into smaller pieces.

PR #1761, opened 2025-12-30
Inline `filter()` after a left/inner join (take two)

hadley:
Part of #722

@mgirlich I think this is way easier to understand now, thanks to all the refactoring I've been working on.

### DIFF
--- a/man/db-sql.Rd
+++ b/man/db-sql.Rd
@@ -68,6 +68,7 @@ sql_query_multi_join(
   table_names,
   by_list,
   select,
+  where = NULL,
   ...,
   distinct = FALSE,
   lvl = 0
--- a/man/sql_build.Rd
+++ b/man/sql_build.Rd
@@ -25,6 +25,7 @@ lazy_multi_join_query(
   table_names,
   vars,
   distinct = FALSE,
+  where = NULL,
   group_vars = op_grps(x),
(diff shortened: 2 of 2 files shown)


R/query-join.R
```diff
@@ -71,11 +73,24 @@ sql_build.lazy_multi_join_query <- function(op, con, ..., sql_options = NULL) {
...
+  # qualified table.column references
+  replacements <- purrr::map2(op$vars$var, op$vars$table, \(var, table_idx) {
+    sql_table_prefix(con, table_names_out[[table_idx]], var)
+  })
+  where <- replace_sym(op$where, op$vars$name, replacements)
+  where_sql <- translate_sql_(
+    where,
+    con = con,
+    context = list(clause = "WHERE")
+  )
```
mgirlich: How about putting this into a function `generate_where_sql()` or something?

mgirlich (approved):

R/query-join.R
hadley: Like this?

```R
translate_sql_where <- function(con, exprs) {
   translate_sql_(exprs, con = con, context = list(clause = "WHERE"))
}
```

And similarly for all the other clauses? If that's what you're thinking, I can do it in another PR.

PR #1766, opened 2026-01-05
Translate `as()` to `cast()`

### LINKED ISSUE #1729: Translate 'as(x, "TYPE")' --> 'CAST(x AS TYPE)'?
We generally support specific vector casts like `as.character()`:

https://github.com/tidyverse/dbplyr/blob/e6e5dc6ecd03124e2771bd22d56bd1fa5b519dd9/R/backend-mysql.R#L88-L92

This is most natural for R vector types; allowing generic `as(x, "TYPE")` to drop directly into a `CAST()` call would naturally extend to _all_ object types that a given SQL flavor supports, _especially_ if it's something backend-specific that might look unnatural as an `as.*` R call.

Taking BigQuery as an example:

https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/data-types

`as(x, "TIME")` would seem most natural to me as opposed to `as.time(x)`.

hadley:
Fixes #1729

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dbplyr (development version)
 
+* `as(x, "type")` is now translated to `CAST(x AS type)`, allowing you to cast
+  to arbitrary database types not covered by the standard `as.*()` functions
+  (#1729).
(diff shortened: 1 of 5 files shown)


MichaelChirico:
Missed this, thanks!

PR #1768, opened 2026-01-05
Implement `sql_set_op_suffix()`

### LINKED ISSUE #1596: Support for union() --> UNION DISTINCT
BigQuery doesn't support unqualified `UNION` clauses:

https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#union

Everything must be `UNION ALL` or `UNION DISTINCT` (the latter being what's usually intended by `UNION` without `ALL`, but clearer to the reader).

Currently, it's not easy to make this tweak in a flavor:

https://github.com/tidyverse/dbplyr/blob/572693098d0098018b6bb2d1e88b650d2e1e1580/R/db-sql.R#L671-L684

I could write my own `sql_query_union()` method, but that would require copy-pasting most of the logic in the default DBIConnection method.

Is the right approach to expose another method, akin to `sql_join_suffix()`, that returns the valid `UNION` operators?

```r
sql_union_operators.DBIConnection <- function(con) c("UNION", "UNION ALL")

sql_union_operators.BigQueryConnection <- function(con) c("UNION DISTINCT", "UNION ALL")
```

hadley:
Fixes #1596

@MichaelChirico what do you think of this? I don't know how I feel about adding a suffix generic, but it does make it easier to generalise across the set operations. It's all claude code so I'm happy to throw it away.

### DIFF
--- a/R/backend-oracle.R
+++ b/R/backend-oracle.R
@@ -220,7 +220,7 @@ sql_values_subquery.Oracle <- function(con, df, types, lvl = 0, ...) {
 sql_set_op_method.Oracle <- function(con, op, ...) {
   # Oracle uses MINUS instead of EXCEPT:
   # https://docs.oracle.com/cd/B19306_01/server.102/b14200/queries004.htm
-  switch(op, "EXCEPT" = "MINUS", "EXCEPT ALL" = "MINUS ALL", op)
+  switch(op, "EXCEPT" = "MINUS", op)
 }
 
 #' @export


MichaelChirico:
It seems fine for the narrower bug fix of enabling `UNION DISTINCT` vs. `UNION ALL`; untouched is whether other extensions can/should be supported, e.g. [`UNION ALL BY NAME`](https://docs.snowflake.com/en/sql-reference/operators-query#general-syntax). I don't have a strong sense of if that's just a nice-to-have.

MichaelChirico:
I'm happy to table `BY NAME` until there's actual user demand. Now it's just in the abstract of us reading documentation & knowing what's available.

MichaelChirico:
LGTM.

nit: the PR title (+description) should be updated

PR #1769, opened 2026-01-06
Ensure that `partial_eval()` returns expressions

hadley:


hadley:
@mgirlich just another FYI. The intent of `partial_eval()` was always to reduce a quosure to an expression, so we no longer need to carry around the environment. Somehow that got partially lost over time, entangling the jobs of `partial_eval()` and `translate_sql()`.

PR #1773, opened 2026-01-12
Replace IIF() with CASE WHEN for Azure Synapse compatibility

rehbbea:
Azure Synapse dedicated pools do not support the IIF() function. This replaces IIF() with standard CASE WHEN syntax which is supported by all SQL Server variants.

- between(): IIF(x BETWEEN l AND r, 1, 0) -> CASE WHEN x BETWEEN l AND r THEN 1 ELSE 0 END
- mssql_as_bit(): CAST(IIF(x, 1, 0) AS BIT) -> CAST(CASE WHEN x THEN 1 ELSE 0 END AS BIT)


hadley:
Thanks!

Issue #1808, opened 2026-03-01
`dbplyr::pull()` behavior with quosures


gergness:
In [a recent `srvyr` issue](https://github.com/gergness/srvyr/issues/204), I'm having trouble with some code that uses `pull()` with quosures. I think the underlying problem is this difference in behavior when a dataset is local vs a `dbplyr` one:

``` r
library(RSQLite)
library(dplyr)
library(dbplyr)

local_df <- data.frame(x = 1:4)
con <- DBI::dbConnect(RSQLite::SQLite(), path = ":memory:")
db_df <- copy_to(con, local_df, "df")

xquo <- rlang::quo(x)
nmquo <- rlang::quo(NULL)

local_df |> pull(!!xquo, !!nmquo)
#> [1] 1 2 3 4

db_df |> pull(!!xquo, !!nmquo)
#> Error in `select()`:
#> ℹ In argument: `all_of(c(var, name))`.
#> Caused by error in `as_indices_impl()`:
#> ! Quosures can only be unquoted within a quasiquotation context.
#> 
#> # Bad: list(!!myquosure)
#> 
#> # Good: dplyr::mutate(data, !!myquosure)
```

<sup>Created on 2026-03-01 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

It's very possible that I'm doing something wrong, it's been a long time since I've had to fully wrap my head around quosures, so let me know if there a better way for me to define `pull.tbl_svy()`?

```
pull.tbl_svy <- function(.data, var = -1, name = NULL, ...) {
  var <- rlang::enquo(var)
  name <- rlang::enquo(name)
  dplyr::pull(.data$variables, !!var, !!name)
}
```

szimmer:
I want to add that this oddly works if there's a value for name other than NULL.

``` r
library(RSQLite)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(dbplyr)
#> 
#> Attaching package: 'dbplyr'
#> The following objects are masked from 'package:dplyr':
#> 
#>     ident, sql

local_df <- data.frame(x = 1:4, id=letters[1:4])
con <- DBI::dbConnect(RSQLite::SQLite(), path = ":memory:")
db_df <- copy_to(con, local_df, "df")

xquo <- rlang::quo(x)
[...]

PR #1809, opened 2026-03-02
Handle NULL quosure for name in pull, add test

szimmer:
This adds code to handle quosures of NULL specified for the name in the pull verb. This addresses #1808


hadley:
Thank you!!

PR #1815, opened 2026-03-23
Translate `anyNA` to SQL

### LINKED ISSUE #1814: `anyNA(x)` is not supported, but `any(is.na(x))` is
I think `anyNA(x)` should be translated exactly as `any(is.na(x))`, but this is not implemented.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(dbplyr)
#> 
#> 
#>     ident, sql
test_dat <- memdb_frame(a = c(1, 2, NA))


summarize(test_dat, anyNA(a))
#> Error in `collect()`:
#> Caused by error:
#> ! no such function: anyNA
summarize(test_dat, any(is.na(a)))
#> Warning: Missing values are always removed in SQL aggregation functions.
#> Use `na.rm = TRUE` to silence this warning
#> This warning is displayed once every 8 hours.
#> # Source:   SQL [?? x 1]
#> # Database: sqlite 3.51.2 [:memory:]
#>   `any(is.na(a))`
#>             <int>
#> 1               1
```

<sup>Created on 2026-03-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

damonbayer:
Closes https://github.com/tidyverse/dbplyr/issues/1814.


hadley:
Thanks @damonbayer! I tweaked the implementation a little to align with other places in dbplyr where we need a "dynamic" translation.

PR #1829, opened 2026-05-15
Implement `filter_out()`

hadley:
Fixes #1803


R/backend-.R
```diff
@@ -288,6 +288,12 @@ base_scalar <- sql_translator(
   na_if = sql_prefix("NULLIF", 2),
   coalesce = sql_prefix("COALESCE"),
 
+  # Internal helper used by `filter_out.tbl_lazy()`.
+  # Not intended for direct use.
+  dbplyr_filter_out_cond = function(x) {
```
hadley: This is needed because `filter_out()` basically generates a `filter()`. This is a bit hacky but it feels like a reasonable approach, and minimises changes elsewhere in the SQL generation pipeline.

It does make me wonder if we should generally provide `%==%` and `%!==%` or similar to expose these semantics everywhere.

R/query-join.R
```diff
@@ -392,7 +392,17 @@ sql_join_tbls <- function(con, by, na_matches) {
       compare <- purrr::map_chr(seq_along(lhs), function(i) {
         op <- by$condition[[i]]
         if (op == "==") {
-          sql_expr_matches(con, lhs[[i]], rhs[[i]])
+          withr::local_options(lifecycle_verbosity = "quiet")
```
hadley: This feels a bit hacky, but I don't see a better way.

hadley:
@DavisVaughan I realised that there's actually little harm in never calling `sql_expr_matches()`: duckplyr will just fallback the default approach until they provide new function translations (https://github.com/duckdb/duckdb-r/issues/2326).

R/backend-mysql.R
```diff
@@ -114,6 +114,9 @@ sql_translation.sql_dialect_mariadb <- function(con) {
   sql_variant(
     sql_translator(
       .parent = base_scalar,
+      # https://dev.mysql.com/doc/refman/5.7/en/comparison-operators.html#operator_equal-to
+      is_distinct_from = \(x, y) sql_glue("NOT (({x}) <=> ({y}))"),
+      is_not_distinct_from = \(x, y) sql_glue("({x}) <=> ({y})"),
```
DavisVaughan, line 119:
```suggestion
      is_distinct_from = function(x, y) sql_glue("NOT (({x}) <=> ({y}))"),
      is_not_distinct_from = function(x, y) sql_glue("({x}) <=> ({y})"),
```

are you using anonymous functions in dbplyr?

DavisVaughan: (if not, there are more cases i did not note in other files)

R/backend-oracle.R
```diff
@@ -143,6 +143,11 @@ sql_translation.sql_dialect_oracle <- function(con) {
   sql_variant(
     sql_translator(
       .parent = base_odbc_scalar,
+      # https://docs.oracle.com/cd/B19306_01/server.102/b14200/functions040.htm
+      is_distinct_from = \(x, y) sql_glue("(decode(({x}), ({y}), 0, 1) = 1)"),
+      is_not_distinct_from = \(x, y) {
+        sql_glue("(decode(({x}), ({y}), 0, 1) = 0)")
+      },
```
DavisVaughan: multiline anonymous, gross!

R/backend-oracle.R
```diff
@@ -143,6 +143,11 @@ sql_translation.sql_dialect_oracle <- function(con) {
   sql_variant(
     sql_translator(
       .parent = base_odbc_scalar,
+      # https://docs.oracle.com/cd/B19306_01/server.102/b14200/functions040.htm
+      is_distinct_from = \(x, y) sql_glue("(decode(({x}), ({y}), 0, 1) = 1)"),
+      is_not_distinct_from = \(x, y) {
+        sql_glue("(decode(({x}), ({y}), 0, 1) = 0)")
```
DavisVaughan: do you think the extra outer `()` is necessary? two places you could maybe remove it
- around all of `(decode()=1)`
- around `({x})` since its the only argument there

R/verb-filter.R
```diff
@@ -1,7 +1,20 @@
...
-#' `WHERE` clause of the SQL query.
+#' @description
+#' These are methods for the dplyr [dplyr::filter()] and [dplyr::filter_out()]
+#' generics. They generate the `WHERE` clause of the SQL query.
+#'
+#' `filter()` is translated directly to `WHERE`, which already matches dplyr's
+#' behaviour of treating `NA` like `FALSE` (SQL's three-valued logic drops
+#' `NULL` rows from `WHERE`).
+#'
+#' `filter_out()` keeps rows where the combined condition is `FALSE` or `NA`.
```
DavisVaughan: I think you should describe `filter_out()` as a _drop_, not a _keep_.

That's the important framing in dplyr.
- `filter()` is "keep when `TRUE`"
- `filter_out()` is "drop when `TRUE`"

R/verb-filter.R
```diff
@@ -1,7 +1,20 @@
...
+#' Multiple conditions are AND-combined and wrapped in
+#' `is_distinct_from(., TRUE)`, which is translated using the backend's
+#' `is_distinct_from()` (e.g. `IS DISTINCT FROM` on PostgreSQL, `IS NOT` on
```
DavisVaughan:
```suggestion
#' `is_distinct_from()` translation (e.g. `IS DISTINCT FROM` on PostgreSQL, `IS NOT` on
```

R/verb-filter.R
```diff
@@ -42,6 +56,31 @@ filter.tbl_lazy <- function(.data, ..., .by = NULL, .preserve = FALSE) {
...
+
+  dots <- enquos(...)
+  # Multiple conditions are AND-combined to match dplyr's semantics, then
+  # wrapped in `is_distinct_from(., TRUE)`. The backend translation of
+  # `is_distinct_from()` provides the dialect-specific SQL.
+  exprs <- lapply(dots, quo_get_expr)
+  combined <- Reduce(function(a, b) expr((!!a) & (!!b)), exprs)
+  quo <- new_quosure(
+    expr(is_distinct_from(!!combined, TRUE)),
+    quo_get_env(dots[[1]])
```
DavisVaughan: I am very sure you can construct counter examples where forcing all expressions to use `quo_get_env(dots[[1]])` does not work, i.e. if the 2nd expression has a different quosure environment.

tests/testthat/_snaps/backend-oracle.md
```diff
@@ -71,7 +71,7 @@
       SELECT "df_LHS"."x" AS "x"
       FROM "df" "df_LHS"
       LEFT JOIN "df" "df_RHS"
-        ON (decode("df_LHS"."x", "df_RHS"."x", 0, 1) = 0)
+        ON ((decode(("df_LHS"."x"), ("df_RHS"."x"), 0, 1) = 0))
```
DavisVaughan: yea seems a bit excessive

R/verb-filter.R
DavisVaughan, line 52: I was kind of expecting that `filter_out()` would not be framed as a `filter()`, but would instead also feed through this common implementation shared by `filter()` and `filter_out()`

Like `add_filter()` would gain an `invert = TRUE/FALSE` parameter, which would be stored as `where_invert` in the `lazy_select_query()` object

And then roughly around this line in `dbplyr:::sql_build.lazy_select_query` is where the inversion comes in

```
where_sql <- translate_sql_(op$where, con = con, context = list(clause = "WHERE"))
```

But I guess that's kinda complicated, and the various cases of `dbplyr:::add_filter()` may make it even more so.

So I guess framing as a `filter()` is fine even if it somehow doesn't feel quite right

DavisVaughan: Maybe the right way is to extract out `filter_impl()` that does share this infrastructure between both functions, but right after

```r
dots <- partial_eval_dots(.data, ..., .named = FALSE)
```

you check `if(invert) {` and then handle the `Reduce + is_distinct_from` bit

then pass on to a standard `add_filter()` call

---

My thoughts about this are that this at least gives `partial_eval_dots()` a chance to evaluate each individual quosure within its dot-specific environment. If I understand `partial_eval_dots()` correctly, then if you do that, then in the `if (invert) {` path you could probably use

```r
quo <- new_quosure(
    expr(is_distinct_from(!!combined, TRUE)),
    # quo_get_env(dots[[1]])
    user_env
)
```

i.e. you assert that `partial_eval_dots()` handled the environment specific evaluation, and now you mostly just have bare expressions that you are combining together, so the environment of your new quosure that wraps all of the individual expressions doesn't matter quite as much and can just be the caller environment

R/verb-filter.R
DavisVaughan: This might fix that https://github.com/tidyverse/dbplyr/pull/1829/changes#r3260688683

R/verb-filter.R
DavisVaughan, line 39: A common `filter_impl()` would also allow any errors here to report `filter_out()` as the caller

DavisVaughan (approved):
Two main comments relating to:
- Docs: Describe filter_out as a _drop_ not a _keep_, consistent with dplyr
- `...` evaluation, and avoiding setting every dot's quosure environment to `quo_env(dots[[1]])` https://github.com/tidyverse/dbplyr/pull/1829/changes#r3260688683

R/backend-mysql.R
hadley: Claude apparently is!

hadley: Looks like I'm using `\(x)` in lots of places.

R/backend-oracle.R
hadley: Definitely can remove in this case, less clear about the others if we're now exposing these functions to the user.

hadley: I think we need to keep the outer, otherwise you might do this:

```R
translate_sql(is_distinct_from(x, y) + 1L, con = dialect_oracle())
#> <SQL> decode("x", "y", 0, 1) = 1 + 1
```

R/verb-filter.R
hadley: I think it might actually be simpler than that since the point of `partial_eval()` is to go from a quosure to an expression.

hadley:
This also made me realised that `filter_out(df)` is supposed to drop all rows.

R/backend-oracle.R
DavisVaughan: That seems reasonable

DavisVaughan: I was mostly irked by the inner one!

R/verb-filter.R
DavisVaughan: I was hoping that was the point of that function, but I saw it was still outputting a quosure in `partial_eval_dots()` via `partial_eval_quo()` so I wasn't sure

hadley: `partial_eval_quo()` is confusingly named; I'm pretty sure it takes a quosure as input, rather than giving a quosure as output.

PR #1831, opened 2026-05-15
Implement `recode_values()` + `replace_values()`

### LINKED ISSUE #1796: FR: support for `recode_values()`
dbplyr has support for `case_match()`, which is now marked as deprecated in dplyr 1.2.0, apparently to be replaced by `recode_values()`.

To still be able to use dbplyr as drop-in-replacement, support for `recode_values()` would be nice.

hadley:
Fixes #1796


tests/testthat/test-translate-sql-conditional.R
```diff
@@ -349,8 +358,190 @@ test_that("`.ptype` not supported", {
...
+    recode_values(x, 1L ~ "a", NULL, 2L ~ "b", NULL),
+    "CASE WHEN (\"x\" IN (1)) THEN 'a' WHEN (\"x\" IN (2)) THEN 'b' END"
+  )
+})
+
+test_that("from/to vector interface works", {
+  con <- dialect_ansi()
+  expect_translation(
+    con,
+    recode_values(x, from = !!c("a", "b"), to = !!c("A", "B")),
```
DavisVaughan, line 430: Try to remove the `!!`

tests/testthat/test-translate-sql-conditional.R
```diff
@@ -349,8 +358,190 @@ test_that("`.ptype` not supported", {
...
+})
+
+test_that("`x` must be a symbol or call", {
+  con <- dialect_ansi()
+  expect_snapshot(error = TRUE, {
+    translate_sql(recode_values("foo", "foo" ~ "FOO"), con = con)
+  })
+})
+
+test_that("list `to` is not supported", {
```
DavisVaughan: A question about this

- When `to` is not a list, it is a vector of size 1 or a vector of size `vec_size(from)`, i.e. map everything to 1 value, or map to 1 value per `from` element. That's the same as `recode_values(x, 'a' ~ 'b')` or `recode_values(x, 'a' ~ 'b', 'c' ~ 'd')`.
- When `to` is a list, the main purpose is to provide a whole column to map to, like "anywhere you see `'a'`, map it to the corresponding location in the `to` column". That's the same as `recode_values(x, 'a' ~ y)`.

Can you actually do `recode_values(x, 'a' ~ y)`? I don't see any tests for this where the RHS is a whole vector to pull matches from. I imagine that could translate to `THEN y`?

Even if you can handle `'a' ~ y`, you probably cannot handle the equivalent `from = 'x', to = list(y)`?. I feel like the lazy evaluation of `y` there makes things awkward. Or maybe you have tooling that works with that (`partial_eval()`?).

tests/testthat/test-translate-sql-conditional.R
```diff
@@ -349,8 +358,190 @@ test_that("`.ptype` not supported", {
...
+  )
+  # `to` recycled to size of `from`
+  expect_translation(
+    con,
+    recode_values(x, from = !!c("a", "b"), to = "X"),
+    "CASE WHEN (\"x\" IN ('a')) THEN 'X' WHEN (\"x\" IN ('b')) THEN 'X' END"
+  )
+})
+
+test_that("from/to list interface works", {
```
DavisVaughan, line 441:
```suggestion
test_that("from list interface works", {
```

`to` lists are not currently supported, see other comment

tests/testthat/test-translate-sql-conditional.R
```diff
@@ -349,8 +358,190 @@ test_that("`.ptype` not supported", {
...
+test_that("NULL values in `...` are dropped", {
+  con <- dialect_ansi()
+  expect_translation(
+    con,
+    recode_values(x, 1L ~ "a", NULL, 2L ~ "b", NULL),
+    "CASE WHEN (\"x\" IN (1)) THEN 'a' WHEN (\"x\" IN (2)) THEN 'b' END"
+  )
+})
+
+test_that("from/to vector interface works", {
```
DavisVaughan, line 426: it might be nice to have an explicit example where the vectors come from an external data frame. like 

```r
table <- tibble(from = , to =)
recode_values(x, from = table$from, to = table$to)
```

in my head i would imagine that this could work? i.e. you always assume `from` and `to` are external sources that you need to evaluate and inline

R/translate-sql-conditional.R
```diff
@@ -1,57 +1,28 @@
...
-
-  sql(paste0(
-    "CASE\n",
-    paste0(clauses, collapse = "\n"),
-    "\nEND"
-  ))
+  sql_case_finalize(clauses)
 }
 
 sql_case_match_clause <- function(f, x, con) {
```
DavisVaughan, line 28: Rename this?

R/translate-sql-conditional.R
```diff
@@ -91,6 +62,178 @@ sql_case_match_clause <- function(f, x, con) {
...
+    if (!is_formula(f, lhs = TRUE)) {
+      cli_abort(
+        "{.arg ...} must contain only two-sided formulas.",
+        call = error_call
+      )
+    }
+    env <- environment(f)
+    query <- sql_case_match_clause(f, x, con)
+    value <- escape(
+      enpar(quo(!!f[[3]]), tidy = FALSE, env = env),
```
DavisVaughan, line 192: i can't comment on the soundness of this 😛

R/translate-sql-conditional.R
```diff
@@ -91,6 +62,178 @@ sql_case_match_clause <- function(f, x, con) {
...
+      call = error_call
+    )
+  }
+
+  n <- length(from)
+  to <- vctrs::vec_recycle(to, n, x_arg = "to", call = error_call)
+
+  clauses <- character(n)
+  for (i in seq_len(n)) {
+    query <- sql_recode_in_clause(from[[i]], x, con)
```
DavisVaughan, line 206: what happens with `recode_values(x, from = list(character()), to = "x")`

i.e. if `sql_recode_in_claude()` gets an empty first input here

DavisVaughan (approved):

tests/testthat/test-translate-sql-conditional.R
hadley: Right: it now supports `recode_values(x, 'a' ~ y)` because we translate the RHS of formulas, but we can't support the equivalent `to = list(y)`, because we don't translate `list()`. But maybe we should?

hadley:
If I pick this up again in the future, I need to start by sitting down and carefully considering the components of a `recode_values()` call that should be translated, and which should be evaluated as is in R. I think the goal will have to be more of an in-spirit translation, since the semantics of dplyr and dbplyr are rather different here.

I'm also slightly concerned that the introduction of a `list()` translation might break existing code since `LIST()` is available in (at least) duckdb. That said `array_agg()` is a more common way to express the same idea.

PR #1835, opened 2026-05-18
Fix `copy_to()` same-source check

hadley:
`copy_to(con, tbl_sql, ...)` was always taking the cross-source branch (collect + dbWriteTable) instead of `compute()`, because `same_src(df, dest)` dispatches to `same_src.tbl_sql()`, which requires `y` to inherit from `tbl_sql`, but `dest` is a `src_sql`. Compare `df$src` to `dest` so both sides are `src_sql` and dispatch reaches `same_src.src_sql()`.

### DIFF
--- a/R/verb-copy-to.R
+++ b/R/verb-copy-to.R
@@ -69,7 +69,7 @@ copy_to.src_sql <- function(
 
   name <- as_table_path(name, dest$con)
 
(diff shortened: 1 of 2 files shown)


hadley:
Fixes accidental regression in #1632

